PR #2, opened 2014-05-26
Related Words

ddaspit:


Build/Src/FwBuildTasks/RegFreeCreator.cs
```diff
@@ -86,19 +86,20 @@ public RegFreeCreator(XmlDocument doc, TaskLoggingHelper Log): this(doc)
...
+		/// 		<item>name (from file name)</item>
+		/// 		<item>version info (from assembly)</item>
+		/// 		<item>type (hard coded as "win32" for now)</item>
+		///  </list>
+		///  This method also adds the root element with all necessary namespaces.
+		///  </summary>
+		///  <param name="pathName">pathname of the file.</param>
+		/// <param name="isSxs"></param>
+		/// ------------------------------------------------------------------------------------
+		public XmlElement CreateExeInfo(string pathName, bool isSxs)
```
JohnThomson: I think some comment is called for here. I don't know what an sxs file is, so others likely won't either. The first hit on google says it is a "SAS Stat Studio Module Source Code file". That's probably wrong. If it isn't some kind of (Linux?) executable then the method name itself has become dubious with this new responsibility.

But, the code above seems to indicate that isSxs will be true if there's only one file we are getting manifest info from and it's the executable. Why would that be reason to change the extension??

Build/Windows.targets
```diff
@@ -99,5 +100,17 @@
...
+	<Target Name="SxSManifests" Condition="'$(OS)'=='Windows_NT'" DependsOnTargets="FwKernel;GrEngine;Views">
+		<RegFree Executable="$(dir-outputBase)\FwKernel.dll" />
+		<RegFree Executable="$(dir-outputBase)\Views.dll" />
+		<RegFree Executable="$(dir-outputBase)\Graphite.dll" NoTypeLib="$(dir-outputBase)\Graphite.dll" />
+	</Target>
+	<PropertyGroup>
+		<ApplicationTransformsPath>..\Src\Transforms\Application</ApplicationTransformsPath>
+		<PresentationTransformsPath>..\Src\Transforms\Presentation</PresentationTransformsPath>
+	</PropertyGroup>
+	<Target Name="BuildWindowsXslAssemblies" Condition="'$(OS)'=='Windows_NT'">
```
JohnThomson, line 108: Is a comment indicated here? I find myself wondering why XslAssemblies are only needed on Windows. Maybe that information is elsewhere?

Src/Common/COMInterfaces/IcuWrappers.cs
```diff
@@ -1266,15 +1264,20 @@ public static int uloc_GetName(string localeID, ref byte[] name, int nameCapacit
...
+
 			const int nSize = 255;
 			IntPtr resPtr = Marshal.AllocCoTaskMem(nSize);
 			try
 			{
+				UErrorCode err;
 				int nResult = uloc_getName(localeID, resPtr, nSize, out err);
-				localeName = Marshal.PtrToStringAnsi(resPtr);
-				return nResult;
+				if (err > UErrorCode.U_ZERO_ERROR)
```
JohnThomson: Is it usual practice to test for errors like this? I believe in fact codes greater than this are errors, while ones less are warnings...but I also thought there was a function to test whether a code is an error? If not I think there should be.

Src/Common/Controls/FwControls/ObtainProjectMethod.cs
```diff
@@ -95,34 +95,25 @@ private static string CreateProjectFromLift(Form parent, IHelpTopicProvider help
 			var anthroListFile = ReflectionHelper.CallStaticMethod(@"FwCoreDlgs.dll", @"SIL.FieldWorks.FwCoreDlgs.FwCheckAnthroListDlg",
 					@"PickAnthroList", parent, null, helpTopicProvider);
 
-			// Do NOT dispose of the thread helper until we dispose of the cache...
-			// CreateProjectTask makes it the thread helper of the FdoCache that it creates and returns,
-			// and we must NOT dispose of the cache's thread helper until after we import the lexicon.
-			// We will actually dispose it twice, most likely, since disposing the cache will dispose it;
-			// but this is harmless.
-			using (var helper = new ThreadHelper())
+			using (var progressDlg = new ProgressDialogWithTask(parent))
```
JohnThomson, line 98: Are you SURE you've dealt with the problem outlined in the deleted comment? It looks to me as though the progress dialog makes its own thread helper, and probably disposes it at the end of this using block, but that is TOO SOON...that thread helper must survive until we are done with ImportObtainedLexicon and dispose the cache.
(Maybe it's OK since we no longer pass a thread helper helper when creating a cache when creating a cache?

Src/Common/Controls/FwControls/FwControlsTests/ProgressDlgTests.cs
```diff
@@ -36,8 +36,7 @@ protected override void Dispose(bool disposing)
 		{
 			if (disposing && !IsDisposed)
 			{
-				if (ThreadHelper != null)
-					ThreadHelper.Dispose();
+				((ThreadHelper) SynchronizeInvoke).Dispose();
```
JohnThomson: This seems strange...how is it better than the old code? Now this method depends on SynchronizeInvoke returning something that needs disposing (and is never null).

Src/Common/Controls/XMLViews/ConfiguredExport.cs
```diff
@@ -640,14 +640,20 @@ private string GetLeadChar(string sEntryNFD, string sWs)
...
-			var col = Icu.ucol_Open(loc, out err);
-			if ((int)err > (int)Icu.UErrorCode.U_ZERO_ERROR)
+
+			IntPtr col;
+			try
+			{
+				string icuLocale = Icu.GetName(sWs);
+				col = Icu.OpenCollator(icuLocale);
+			}
+			catch (Exception)
```
JohnThomson: This looks like a case of using an exception for a situation that isn't all that exceptional...that might well happen in normal operation of the program. I consider that an anti-pattern: among other problems, it is very frustrating when the debugger is set to stop for all exceptions. (We do this quite a lot, but I'm trying to resist making it worse.)

Src/Common/CoreImpl/CoreImplTests/DirectoryFinderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+		/// is set.
+		/// </summary>
+		[TestFixture]
+		[Property("Value", null)]
+		public class GetCommonAppDataNormalTests: GetCommonAppDataBaseTest
+		{
+			/// <summary>Tests the GetFolderPath method for CommonApplicationData when no environment
+			/// variable is set</summary>
+			[Test]
+			[Platform(Include="Linux", Reason="Test is Linux specific")]
```
JohnThomson: Reason does not add anything: Include="Linux" says it is Linux-specific. Reason should say why this is so.

(Would it be clearer to have a single test GetFolderPath_ReturnsExpectedDefault(), with conditional compilation determining the expected result?)

Src/Common/CoreImpl/CoreImplTests/DirectoryFinderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+			{
+				Assert.AreEqual("/bla",
+					DirectoryFinder.GetFolderPath(Environment.SpecialFolder.CommonApplicationData));
+			}
+
+			/// <summary>Tests the GetFolderPath method for CommonApplicationData when the environment
+			/// variable is set</summary>
+			[Test]
+			[Platform(Exclude="Linux", Reason="Test is Windows specific")]
+			public void Windows()
```
JohnThomson: This test appears identical to the one in the default case. Study eventually makes clear that a different setting for the first value will cause SetEnvironmentVariable to be called so Environment.SpecialFolder.CommonApplicationData will yield a different result; but it seems a fragile test. For example, if the environment variable name was changed in some places (but not in the test code above) it would still pass, but it would not be testing what you want.

Src/Common/COMInterfaces/IcuWrappers.cs
ddaspit: This is the normal way to test for errors. The C library has two macros "U_SUCCESS" and "U_FAILURE" that is used to check error codes. The "U_FAILURE" macro basically checks if the error code is greater than zero, which is what I have done here. We don't have any method like the C macros to check the error code.

Src/Common/Controls/FwControls/ObtainProjectMethod.cs
ddaspit, line 98: We changed FdoCache.CreateNewLangProj so that it no longer takes a ThreadHelper, but instead takes an ISynchronizeInvoke. We don't have to create a ThreadHelper anymore. We can now pass in the parent form instead, because it implements ISynchronizeInvoke, which won't be disposed in this method. Does that make sense?

Src/Common/Controls/FwControls/FwControlsTests/ProgressDlgTests.cs
ddaspit: If you look in the constructor of this class, you can see that it creates a ThreadHelper and passes it to the base class constructor. The constructor for ProgressDialogWithTask no longer takes ThreadHelper, instead it takes ISynchronizeInvoke. We know that the SynchronizeInvoke property is actually a ThreadHelper object, which needs to be disposed, so we cast it to ThreadHelper and then dispose it.

Src/Common/Controls/XMLViews/ConfiguredExport.cs
ddaspit: I believe this is an exceptional case. The only way this would throw an exception would be if the writing system ID wasn't a valid ICU locale. I was just trying to mimic the existing code. I could remove the try-catch and just let the exception bubble up the stack.

Src/Common/CoreImpl/CoreImplTests/DirectoryFinderTests.cs
ddaspit: These were existing tests that I moved from FwUtils when I split DirectoryFinder into two classes. You're suggestion makes sense. I will make the change.

Build/Windows.targets
ddaspit, line 108: I added comments to clarify these build targets.

Src/Common/COMInterfaces/IcuWrappers.cs
JohnThomson: I'm not sure how many places in our code know that error codes greater than zero indicate failure, but it seems to me that it would aid readability and just possibly maintainability to make a static method, maybe on IcuWrappers, to encapsulate this.

If we're already doing this widely and the change would make the change set much bigger it's probably not worth it (at least as part of this one).

Src/Common/Controls/FwControls/FwControlsTests/ProgressDlgTests.cs
JohnThomson: And hopefully this field is read-only, so no method of this class or the base class could replace the ThreadHelper with another kind of SynchronizeInvoke?
This still feels fragile, but I suppose anything that puts something else in that field will cause a crash pretty promptly before it escapes into the wild, so it probably doesn't matter. Be nice to make the base class dispose the SynchronizeInvoke iff it is Disposable, but I guess sometimes it is the parent window and we are not ready to dispose it?
Be nice to at least have a comment saying that in this subclass SynchronizeInvoke is always the ThreadHelper created in the constructor.

Src/Common/Controls/FwControls/ObtainProjectMethod.cs
JohnThomson, line 98: Yes, I thought it was probably OK when I saw that you no longer pass a threadhelper to CreateNewLangProject...just wanted to make sure you had thought it through, because I recall this being a nasty bug to deal with.

Src/Common/Controls/XMLViews/ConfiguredExport.cs
JohnThomson: That's probably exceptional enough to let this stand...we don't want to suddenly make the program start crashing if someone has a non-standard code.
Any chance you can pin down what subclass of Exception is produced by a non-standard locale, and just catch that? Catching all Exceptions is also something to avoid if possible since it can mask problems that should be attended to.

Src/Common/CoreImpl/CoreImplTests/DirectoryFinderTests.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
+using NUnit.Framework;
+
+namespace SIL.CoreImpl
+{
+		/// <summary>
+		/// Base class for testing CommonApplicationData. This base class deals with setting
+		/// and resetting the environment variable.
+		/// </summary>
+		public class GetCommonAppDataBaseTest
```
JohnThomson: Public class not in a file of the same name...is this something you just moved, or is it new? I see from your global comments that DirectoryFinder moved, so I'm going to assume that this doesn't need a detailed review.

Src/Common/CoreImpl/CoreImplTests/DirectoryFinderTests.cs
JohnThomson: I realized later that this file was just moved. So I'll try not to be too picky about its content! Up to you whether you think this worth changing. There is some advantage (for managing change history) in NOT making more changes than you can help in the same checkin that moves a file.

Src/Common/CoreImpl/DirectoryFinder.cs
```diff
@@ -0,0 +1,124 @@
...
+				}
+				return s_CommonAppDataFolder;
+			}
+		}
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets a special folder, very similar to Environment.GetFolderPath. The main
+		/// difference is that this method works cross-platform and does some translations.
+		/// For example CommonApplicationData (/usr/share) is not writeable on Linux, so we
```
JohnThomson, line 87: It's not writeable on Windows, either, except for administrators, at least in recent OS versions, so we shouldn't be depending on being able to do that. Do we really need something different from the default on Linux?

Src/Common/FieldWorks/FieldWorks.cs
```diff
@@ -792,15 +803,62 @@ private static FdoCache CreateCache(ProjectId projectId)
...
+				{
+					using (var folderBrowserDlg = new FolderBrowserDialogAdapter())
+					{
+						folderBrowserDlg.Description = Properties.Resources.ksLinkedFilesFolder;
+						folderBrowserDlg.RootFolder = Environment.SpecialFolder.Desktop;
+						folderBrowserDlg.SelectedPath = defaultFolder;
+						if (folderBrowserDlg.ShowDialog() == DialogResult.OK)
+							linkedFilesFolder = folderBrowserDlg.SelectedPath;
+					}
+				}
```
JohnThomson, line 853: Seems that somewhere here we should do something to encourage selecting a folder within the project folder. It's a pain for Send/Receive if the linked files aren't part of the project. I think we have something of the sort in the code for changing the linked folder...maybe we could reuse some of that?

Src/Common/FieldWorks/FieldWorks.cs
```diff
@@ -792,15 +803,62 @@ private static FdoCache CreateCache(ProjectId projectId)
...
+			{
+				if (!Directory.Exists(defaultFolder))
+					defaultFolder = cache.ProjectId.ProjectFolder;
+				MessageBox.Show(String.Format(Properties.Resources.ksInvalidLinkedFilesFolder, linkedFilesFolder), Properties.Resources.ksErrorCaption);
+				while (!Directory.Exists(linkedFilesFolder))
+				{
+					using (var folderBrowserDlg = new FolderBrowserDialogAdapter())
+					{
+						folderBrowserDlg.Description = Properties.Resources.ksLinkedFilesFolder;
+						folderBrowserDlg.RootFolder = Environment.SpecialFolder.Desktop;
```
JohnThomson, line 848: I'm not sure of the significance of RootFolder, but it's not obvious why it should be the desktop.

Src/Common/FieldWorks/FieldWorks.cs
```diff
@@ -792,15 +803,62 @@ private static FdoCache CreateCache(ProjectId projectId)
...
+				NonUndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW(cache.ActionHandlerAccessor, () =>
+					{ cache.LangProject.LinkedFilesRootDir = linkedFilesFolder; });
+			}
+		}
+
+		/// <summary>
+		/// Just make the directory if it's the default.
+		/// See FWNX-1092, LT-14491.
+		/// </summary>
+		internal static void EnsureValidLinkedFilesFolderCore(string linkedFilesFolder, string defaultLinkedFilesFolder)
```
JohnThomson, line 863: It's not obvious why we pull this little piece out into its own method

Src/Common/FieldWorks/FieldWorks.cs
```diff
@@ -1727,7 +1785,16 @@ internal static ProjectId ChooseLangProject(Form dialogOwner, IHelpTopicProvider
 							PropertyTable.SettingsGroup.LocalSettings);
 					}
 				}
-				return dlg.DialogResult != DialogResult.OK ? null : new ProjectId(dlg.Project, dlg.Server);
+
+				if (dlg.DialogResult == DialogResult.OK)
+				{
+					var projId = new ProjectId(dlg.Project, dlg.Server);
+					if (ParatextHelper.GetAssociatedProject(projId) != null)
```
JohnThomson: I've seen these couple of lines changing the projId.Type if there's an associated PT project before. Seems worth a method...if only in case we ever need to do additional or different things when there is an associated PT project, or a different way of testing for it, or there are other reasons to want a shared XML backend...or at least encapsulate the condition in something like SharedBackendNeeded().

Src/Common/FieldWorks/FieldWorks.cs
```diff
@@ -2681,7 +2758,7 @@ private static void CheckForMovingExternalLinkDirectory(FwApp app)
 				if (oldDir == null)
 				{
 					// e.g. "C:\\ProgramData\\SIL\\FieldWorks"
-					oldDir = DirectoryFinder.CommonAppDataFolder("SIL/FieldWorks");
+					oldDir = DirectoryFinder.CommonAppDataFolder("FieldWorks");
```
JohnThomson, line 2766: I didn't pick up when reviewing DirectoryFinder.CommnAppDataFolder that you had 'moved' the root. I hope you caught all callers!

Src/Common/FieldWorks/FieldWorksTests/FieldWorksTests.cs
```diff
@@ -114,5 +127,86 @@ public void GetProjectMatchStatus_SingleProcessMode()
...
+		[Test]
+		public void EnsureValidLinkedFilesFolderCore_IfUsingDefaultDirAndItExists_DoesntCrashOrAnything()
+		{
+			EnsureValidLinkedFilesFolderCore_TestHelper(defaultFolder => {
+				var configuredFolder = defaultFolder;
+
+				// Make default linked files directory already exist
+				FileUtils.EnsureDirectoryExists(defaultFolder);
+
+				Assert.That(FileUtils.DirectoryExists(configuredFolder), Is.True, "Unit test not testing what it's supposed to");
```
JohnThomson: Since you made the two variables the same, and so just ensured that it exists, this assertion amounts to verifying the behavior of EnsureDirectory, more than verifying the initial conditions for the test.

Src/Common/FieldWorks/FieldWorksTests/FieldWorksTests.cs
```diff
@@ -114,5 +127,86 @@ public void GetProjectMatchStatus_SingleProcessMode()
...
+				// Not crash or anything
+				FieldWorks.EnsureValidLinkedFilesFolderCore(configuredFolder, defaultFolder);
```
JohnThomson: Would it be clearer to just pass defaultFolder, defaultFolder?

Might make things clearer to use Assert.DoesNotThrow().

Src/Common/FwUtils/FwDirectoryFinder.cs
```diff
@@ -0,0 +1,698 @@
...
+		/// <param name="file">Name of the file (case-sensistive, with the extension)</param>
+		/// ------------------------------------------------------------------------------------
+		private static string ExeOrDllPath(string file)
+		{
+			if (file == null)
+				throw new ArgumentNullException("file");
+
+			if (Assembly.GetEntryAssembly() == null)
+			{
+				// This seems to happen when tests call this method when run from NUnit
```
JohnThomson, line 193: According to the doc of GetEntryAssembly, it will happen any time the top level exe is not managed code.

Since this hack assumes anyway that all our code is in output/{debug/release}, I wonder whether it would be better to use GetExecutingAssembly.location, like ReflectionHelper?

Src/Common/FwUtils/FwDirectoryFinder.cs
```diff
@@ -0,0 +1,698 @@
...
+		/// <returns></returns>
+		/// ------------------------------------------------------------------------------------
+		public static string GetCodeSubDirectory(string subDirectory)
+		{
+			return GetSubDirectory(CodeDirectory, subDirectory);
+		}
+
+		private static string GetFLExBridgeFolderPath()
+		{
+			// Setting a Local Machine registry value is problematic for Linux/Mono.  (FWNX-1180)
```
JohnThomson, line 260: And for Windows! writing HKLM requires admin.

Src/Common/FwUtils/FwDirectoryFinder.cs
```diff
@@ -0,0 +1,698 @@
...
+		}
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets a subdirectory of FieldWorks either by reading the
+		/// <paramref name="registryValue"/> or by getting <paramref name="defaultDir"/>.
+		/// Will not return <c>null</c>.
+		/// </summary>
+		/// <param name="registryValue">The name of the registry value to read from the FW root
+		/// key in HKLM.</param>
```
JohnThomson, line 304: Actually tries HKCU first

Src/Common/FwUtils/FwDirectoryFinder.cs
```diff
@@ -0,0 +1,698 @@
...
+			if (string.IsNullOrEmpty(rootDir))
+			{
+				throw new ApplicationException(
+					ResourceHelper.GetResourceString("kstidInvalidInstallation"));
+			}
+			// Hundreds of callers of this method are using Path.Combine with the results.
+			// Combine only works with a root directory if it is followed by \ (e.g., c:\)
+			// so we don't want to trim the \ in this situation.
+			string dir = rootDir.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+			return dir.Length > 2 ? dir : dir + Path.DirectorySeparatorChar;
```
JohnThomson, line 373: Could there be a one-or-two character path on Linux where we would not want to add the separator?

Src/Common/FwUtils/FwDirectoryFinder.cs
```diff
@@ -0,0 +1,698 @@
...
+
+		private const string ksProjects = "Projects";
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets or sets the dir where projects are stored. Setting to null will delete the HKCU
+		/// key, so that the HKLM key (system default) will be used for this user.
+		/// </summary>
+		/// <exception cref="SecurityException">If user does not have permission to write to HKLM
+		/// </exception>
```
JohnThomson, line 559: This appears to be obsolete? We no longer try to write to HKLM?

Src/Common/ScriptureUtils/ParatextHelper.cs
```diff
@@ -263,20 +289,36 @@ public IEnumerable<ScrText> GetProjects()
...
+					importSettings.ParatextBTProj = null;
 
-				// If Paratext is not initialized or the new project ID is null, then do not load mappings.
-				if (!m_IsParatextInitialized || project == null)
+				if (!LoadProjectMappings(importSettings.ParatextNotesProj, importSettings.GetMappingListForDomain(ImportDomain.Annotations), ImportDomain.Annotations))
+					importSettings.ParatextNotesProj = null;
+			}
+
+			[SuppressMessage("Gendarme.Rules.Correctness", "EnsureLocalDisposalRule",
+				Justification = "See REVIEW comment")]
```
JohnThomson, line 317: What REVIEW comment??

Src/Common/SimpleRootSite/SimpleRootSite.csproj
```diff
@@ -90,7 +90,7 @@
     <DebugSymbols>true</DebugSymbols>
     <FileAlignment>4096</FileAlignment>
     <NoStdLib>false</NoStdLib>
-    <NoWarn>168,169,219,414,649,1635,1702,1701</NoWarn>
+    <NoWarn>168,169,219,414,649,1635,1702,1701,1685</NoWarn>
```
JohnThomson, line 93: Do you know why we suddenly need to ignore this warning (which I believe has to do with DLLs compiled against different versions of the runtime)?

Src/FDO/Application/ApplicationServices/XmlImportData.cs
```diff
@@ -180,7 +180,7 @@ public XmlImportData(FdoCache cache)
 		/// </summary>
 		/// <returns>true if successful, false if an error occurs</returns>
```
JohnThomson: This comment should be removed

Src/FDO/Application/ApplicationServices/XmlImportData.cs
```diff
@@ -262,7 +260,9 @@ private void LogMessage(string sMsg)
...
```
JohnThomson: Comment obsolete

Src/FDO/Application/ApplicationServices/XmlList.cs
```diff
@@ -83,110 +75,103 @@ public void ImportList(ICmObject owner, string sFieldName, TextReader reader, IP
 			m_wsf = m_cache.WritingSystemFactory;
 			m_progress = progress;
 
-			try
-			{
-				NonUndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW(m_cache.ActionHandlerAccessor, () =>
+			NonUndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW(m_cache.ActionHandlerAccessor, () =>
```
JohnThomson, line 78: Odd indent here

Src/FDO/DomainImpl/OverridesCellar.cs
```diff
@@ -1761,51 +1758,53 @@ internal static ITsString TSSAbbrforWS(FdoCache cache, ICmPossibility pss, int w
...
+
+				CmPossibility rootPossibility = this;
+				while (rootPossibility.Owner is CmPossibility)
+					rootPossibility = (CmPossibility) rootPossibility.Owner;
+				IDsChart chart = Services.GetInstance<IDsChartRepository>().InstancesWithTemplate(rootPossibility).FirstOrDefault();
+				return chart != null && GetIsInUseAsChartColumn();
+			}
+		}
+
+		private bool GetIsInUseAsChartColumn()
```
JohnThomson: This is a prettier name, but it does still also check whether one of its children is thus in use

Src/FDO/DomainImpl/OverridesCellar.cs
```diff
@@ -1761,51 +1758,53 @@ internal static ITsString TSSAbbrforWS(FdoCache cache, ICmPossibility pss, int w
...
-				// This is an actual column; it's a problem if it has instances
-				string msg = string.Format(Strings.ksCantModifyTemplateInUse, textName);
-				MessageBoxUtils.Show(msg, Strings.ksErrorCaption, MessageBoxButtons.OK, MessageBoxIcon.Error);
-				return true;
+				ICmPossibilityList tags = Cache.LangProject.TextMarkupTagsOA;
+				if (OwningList != tags)
+					return false;
+
+				IFdoOwningSequence<ICmPossibility> tagTypes = tags.PossibilitiesOS;
+				return tagTypes.Count == 1 && Hvo == tagTypes[0].Hvo;
```
JohnThomson: Why mess with Hvos? Can't we just test this == tagTypes[0]?

Src/FDO/DomainImpl/OverridesCellar.cs
```diff
@@ -1761,51 +1758,53 @@ internal static ITsString TSSAbbrforWS(FdoCache cache, ICmPossibility pss, int w
...
+			return SubPossibilitiesOS.Cast<CmPossibility>().Any(poss => poss.GetIsInUseAsChartColumn());
 		}
 
-		/// <summary>
-		/// If the recipient is a column in a chart that shouldn't be moved or promoted or deleted, report
-		/// accordingly and return true. Return false if OK to delete or move.
-		/// </summary>
-		/// <returns></returns>
-		public bool CheckAndReportProtectedChartColumn()
+		public bool IsLastTextMarkupTag
```
JohnThomson: Name or implementation seems wrong. Actually returns true if it is the ONLY TextMarkupTag. I guess it means 'last' in the sense of 'only remaining'. That might be OK, but I think something else is worth considering

Src/FDO/DomainImpl/OverridesLangProj.cs
```diff
@@ -144,7 +143,7 @@ public IList<IStText> InterlinearTexts
 
 				// Get regular texts.
 
-				if (FwUtils.IsOkToDisplayScriptureIfPresent && TranslatedScriptureOA != null)
```
JohnThomson, line 147: Why is the removed check no longer necessary?

Src/FDO/DomainImpl/ScrImportSet.cs
```diff
@@ -254,104 +257,7 @@ public TypeOfImport ImportTypeEnum
 		#endregion
 
 		#region Import Project Accessibility Methods
```
JohnThomson: We seem to have gotten rid of all the methods in this region, why not remove the region?

Src/FDO/DomainServices/ClientServerServices.cs
```diff
@@ -162,22 +150,11 @@ public interface IClientServerServices
...
-		/// Display a warning indicating that it may be dangerous to change things in the dialog that
-		/// is about to open when other users are connected. The warning should only be shown if, in fact,
-		/// other users are currently connected. The dialog may contain some information about the other
-		/// users that are connected. Return true to continue, false to cancel opening the dialog.
+		/// Returns the number of other users currently connected
 		/// </summary>
-		/// <returns></returns>
-		bool WarnOnOpeningSingleUserDialog(FdoCache cache);
+		/// <param name="cache">The FDO cache.</param>
+		/// <returns>The number of other users currently connected</returns>
```
JohnThomson: Seems somewhat redundant with the main comment

Src/FDO/DomainServices/Db4OServerFinder.cs
```diff
@@ -63,11 +67,6 @@ public Db4OServerFinder(Action<string> foundServerCallback, Action onCompletedCa
 		/// ------------------------------------------------------------------------------------
 		private void BroadcastToFindHosts()
 		{
-			// Ensure the thread is listening before doing the broadcast
-			// as it could return before we are listening for it.
-			while (m_hostListenerThread.IsAlive && (m_hostListenerSocket == null || !m_hostListenerSocket.IsBound))
-				Thread.Sleep(80);
```
JohnThomson, line 69: I'm not clear why the block above is no longer needed.

Src/FDO/FDOTests/BackupRestore/BackupFileRepositoryTests.cs
```diff
@@ -184,11 +182,11 @@ public void BackupHasOldStyleDatetimeFormat()
...
 			DummyBackupProjectSettings backupSettings = new DummyBackupProjectSettings("monkey",
 				"Floozy", null, FDOBackendProviderType.kXML);
-			string backupFileName1 = Path.Combine(DirectoryFinder.DefaultBackupDirectory,
-				Path.ChangeExtension("Floozy 2010-8-21-0506", FwFileExtensions.ksFwBackupFileExtension));
+			string backupFileName1 = Path.Combine(FwDirectoryFinder.DefaultBackupDirectory,
+				Path.ChangeExtension("Floozy 2010-8-21-0506", FdoFileHelper.ksFwBackupFileExtension));
 			m_fileOs.AddExistingFile(backupFileName1);
 
-			BackupFileRepository repo = new BackupFileRepository();
+			BackupFileRepository repo = new BackupFileRepository(FwDirectoryFinder.DefaultBackupDirectory);
```
JohnThomson, line 189: I rather expected to see more changes to this file, given how much backup/restore code is now throwing exceptions rather than displaying dialogs...maybe the stuff that displayed dialogs was never tested?

Src/FDO/FDOTests/BackupRestore/ProjectRestoreServiceTests.cs
```diff
@@ -292,7 +285,7 @@ internal static void RemoveAnyFilesAndFoldersCreatedByTests(RestoreProjectSettin
 			RemoveAllFilesFromFolderAndSubfolders(settings.OtherExternalFilesPath);
 			RemoveAllFilesFromFolderAndSubfolders(settings.LinkedFilesPath);
 			RemoveAllFilesFromFolderAndSubfolders(settings.WritingSystemStorePath);
-			RemoveAllFilesFromFolderAndSubfolders(Path.Combine(settings.ProjectPath, DirectoryFinder.ksSortSequenceTempDir));
+			RemoveAllFilesFromFolderAndSubfolders(Path.Combine(settings.ProjectPath, FdoFileHelper.ksSortSequenceTempDir));
 
 			//Remove this one last of all because the other folders need to be removed first.
 			RemoveAllFilesFromFolderAndSubfolders(settings.ProjectPath);
```
JohnThomson, line 291: Again, I'm slightly puzzled not to see changes related to reporting problems.

Src/FDO/FDOTests/DummyFdoUI.cs
```diff
@@ -0,0 +1,158 @@
...
+
+using System;
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using SIL.Utils;
+
+namespace SIL.FieldWorks.FDO.FDOTests
+{
+	/// <summary>
+	/// Dummy implementation of FdoUserAction for unit tests
```
JohnThomson: Is FdoUserAction here obsolete?

Src/FDO/FDOTests/FdoScriptureTests.cs
```diff
@@ -1678,7 +1684,8 @@ public void RecordError_NearDuplicate_DifferByCitedText()
 		[Test]
 		public void GetTextTokens_Chapter0()
 		{
-			ScrChecksDataSource dataSource = new ScrChecksDataSource(Cache);
+			ScrChecksDataSource dataSource = new ScrChecksDataSource(Cache, ResourceHelper.GetResourceString("kstidPunctCheckWhitespaceChar"),
```
JohnThomson: This rather complex pair of lines is occurring a lot...maybe worth encapsulating?

Src/FDO/FDOTests/FwSetupFixtureClass.cs
```diff
@@ -24,6 +27,8 @@ public class FwSetupFixtureClass
 		[SetUp]
 		public void SetUp()
 		{
+			ClientServerServices.SetCurrentToDb4OBackend(new DummyFdoUI(), FwDirectoryFinder.FdoDirectories,
```
JohnThomson: Again, I'm starting to feel that I see something very like these two lines in a lot of places. Could we put a helper method somewhere?

Src/FDO/FDOTests/LinkedFilesRelativePathHelperTests.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// <summary>
+		/// Tests the GetLinkedFilesRelativePathFromFullPath method
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[Test]
+		public void GetLinkedFilesRelativePathFromFullPath()
+		{
+			Assert.AreEqual(String.Format("%proj%{0}LinkedFiles", Path.DirectorySeparatorChar),
+				LinkedFilesRelativePathHelper.GetLinkedFilesRelativePathFromFullPath(FwDirectoryFinder.ProjectsDirectory, String.Format("%proj%{0}LinkedFiles", Path.DirectorySeparatorChar),
+					Path.Combine(FwDirectoryFinder.SourceDirectory, "FDO/FDOTests/BackupRestore/Project"),
```
JohnThomson: Will the literal slashes in this line cause a problem for Linux?

Src/FDO/FDOTests/LinkedFilesRelativePathHelperTests.cs
```diff
@@ -0,0 +1,137 @@
...
+			var audioVisualFile = Path.Combine(Path.Combine(linkedFilesRootDir, "AudioVisual"), "StarWars.mvi");
+			var relativeLFPath = LinkedFilesRelativePathHelper.GetRelativeLinkedFilesPath(audioVisualFile, linkedFilesRootDir);
+			Assert.True(String.Equals(String.Format("AudioVisual{0}StarWars.mvi", Path.DirectorySeparatorChar), relativeLFPath));
+
+			//Ensure ORIGINAL path is returned when the path is not relative to the LinkedFiles directory.
+			var pathNotUnderLinkedFiles = Path.Combine(FwDirectoryFinder.DataDirectory, "LordOfTheRings.mvi");
+			relativeLFPath = LinkedFilesRelativePathHelper.GetRelativeLinkedFilesPath(pathNotUnderLinkedFiles, linkedFilesRootDir);
+			Assert.True(String.Equals(pathNotUnderLinkedFiles,relativeLFPath));
+			Assert.That(LinkedFilesRelativePathHelper.GetRelativeLinkedFilesPath(
+				"silfw:\\localhost\\link?app%3dflex%26database%3dc%3a%5cTestLangProj%5cTestLangProj.fwdata%26server%3d%26tool%3dnaturalClassedit%26guid%3d43c9ba97-2883-4f95-aa5d-ef9309e85025%26tag%3d",
```
JohnThomson: Do hyperlinks ever use backslashes?

Src/FDO/FDOTests/PersistingBackendProviderTests.cs
```diff
@@ -828,9 +830,231 @@ internal override void ReportDuplicateGuidsIfTheyExist()
...
+					otherEntry1.CitationForm.set_String(otherCache.DefaultVernWs, "othercitation");
+				});
+				otherCache.ServiceLocator.GetInstance<IUndoStackManager>().Save();
+
+				NonUndoableUnitOfWorkHelper.Do(Cache.ServiceLocator.ActionHandler, () =>
+				{
+					entry1.CitationForm.set_String(Cache.DefaultVernWs, "citation");
+				});
+				Cache.ServiceLocator.GetInstance<IUndoStackManager>().Save();
+				Assert.That(entry1.CitationForm.VernacularDefaultWritingSystem.Text, Is.EqualTo(otherEntry1.CitationForm.VernacularDefaultWritingSystem.Text));
```
JohnThomson: So, they wound up the same...don't we care who won? Shouldn't something warn the user whose changes were lost?

Src/Common/Controls/FwControls/FwControlsTests/ProgressDlgTests.cs
ddaspit: I went ahead and refactored this so that the ThreadHelper is created during test setup and disposed during test teardown.

Src/Common/CoreImpl/CoreImplTests/DirectoryFinderTests.cs
ddaspit: I went ahead and refactored this whole class, because there were a couple of issues.

Src/Common/Controls/XMLViews/ConfiguredExport.cs
ddaspit: I changed it to catch a more specific exception.

Src/Common/COMInterfaces/IcuWrappers.cs
ddaspit: I added methods to check ICU error code for success or failure in IcuWrappers.

Src/Common/CoreImpl/DirectoryFinder.cs
ddaspit, line 87: I don't really know. This was existing code that I moved from FwUtils. I will probably just leave it alone, since I'm not sure why it was done this way.

Src/Common/FieldWorks/FieldWorks.cs
ddaspit, line 848: Yeah, I'm not sure why either. This was existing code that was moved from FDO (I believe).

Src/Common/FieldWorks/FieldWorks.cs
ddaspit, line 853: It looks like the default selected folder in the FolderBrowserDialog is the default project linked files path.

Src/Common/FieldWorks/FieldWorks.cs
ddaspit: I added a method to encapsulate the condition as you suggested.

Src/Common/FieldWorks/FieldWorks.cs
ddaspit, line 2766: I didn't move the root. I believe that the original code was incorrect, so this was basically a bug fix.

Src/Common/FieldWorks/FieldWorksTests/FieldWorksTests.cs
ddaspit: Once again, this was existing code that was moved from FDOTests. I went ahead and refactored it anyway.

Src/Common/FwUtils/FwDirectoryFinder.cs
ddaspit, line 193: I agree that makes sense to me. This is existing code that got moved from elsewhere. I will go ahead and update it.

ddaspit, line 193: I tried changing this code to use GetExecutingAssembly().Location, but it returned a path to a folder in the temp directory, instead of the Output\Debug directory. I will just leave this code as is for now.

Src/Common/FwUtils/FwDirectoryFinder.cs
ddaspit, line 260: All of this is existing code that was moved here. I will probably just leave it alone for now.

Src/Common/ScriptureUtils/ParatextHelper.cs
ddaspit, line 317: The REVIEW comment is a couple of lines farther down. Unfortunately, you can't see it in this diff view.

Src/Common/SimpleRootSite/SimpleRootSite.csproj
ddaspit, line 93: I'm not sure what changed that caused this to be a problem for me. When I try to do a remakefw on my machine, I get a compile error on SimpleRootSite saying that INotifyCollectionChanged is defined multiple times. It is defined in System.dll and in WindowsBase.dll. The WindowsBase.dll is a .NET assembly and SimpleRootSite is using an older version of this assembly (.NET 3) for some reason. In .NET 4, this interface was moved from WindowsBase.dll to System.dll. When I ignore this warning, the remakefw works.

Src/FDO/Application/ApplicationServices/XmlImportData.cs
ddaspit: Removed.

Src/FDO/Application/ApplicationServices/XmlImportData.cs
ddaspit: Removed.

Src/FDO/DomainImpl/OverridesCellar.cs
ddaspit: That is what I meant by "last". I changed the name to IsOnlyTextMarkupTag.

Src/FDO/DomainImpl/OverridesCellar.cs
ddaspit: I renamed this property to reflect that it checks descendants as well.

Src/FDO/DomainImpl/OverridesCellar.cs
ddaspit: Fixed.

Src/FDO/DomainImpl/OverridesLangProj.cs
ddaspit, line 147: IsOkToDisplayScriptureIfPresent is hardcoded to always returns true. The purpose of this property is to make it easier to disable Scripture from appearing anywhere in FieldWorks if we wanted to be able to do that at some point in the future. In any case, this property is FieldWorks-specific, which is why we removed its use from within FDO. FLEx or TE can filter out Scripture from the InterlinearTexts property if it wants to using the IsOkToDisplayScriptureIfPresent property. At this point, I can't find any FieldWorks code that actually uses this property.

Src/FDO/DomainImpl/ScrImportSet.cs
ddaspit: Removed the region.

Src/FDO/DomainServices/ClientServerServices.cs
ddaspit: Removed.

Src/FDO/DomainServices/Db4OServerFinder.cs
ddaspit, line 69: Previously, the listener socket was bound in the thread. This was refactored so that socket is bound prior to starting the thread. This block basically is looping until the socket is bound in the thread. This is no longer necessary as we know that the socket has been bound already.

Src/FDO/FDOTests/BackupRestore/BackupFileRepositoryTests.cs
ddaspit, line 189: Yeah, I think you're right. It looks like the cases that displayed dialogs were never tested.

Src/FDO/FDOTests/DummyFdoUI.cs
ddaspit: Yeah, this was class was renamed at some point. I fixed it.

Src/FDO/FDOTests/FdoScriptureTests.cs
ddaspit: Fixed.

Src/FDO/FDOTests/FwSetupFixtureClass.cs
ddaspit: Encapsulated this call in a method in the FdoTestHelper class.

Src/FDO/FDOTests/LinkedFilesRelativePathHelperTests.cs
ddaspit: I don't think so. I fixed it.

Src/FDO/FDOTests/LinkedFilesRelativePathHelperTests.cs
ddaspit: I don't think they will, but it is still not good practice. This is existing code that got moved. I will fix it anyway.

Src/FDO/FDOTests/PersistingBackendProviderTests.cs
ddaspit: I updated the unit test to check whether the user was prompted.

PR #4, opened 2019-07-18
Fix and Test Patching

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/fieldworks/4)


papeh:
Gerrit hath returned to the land of the living

PR #12, opened 2024-03-21
Ignore tests that are trouble for CI systems

jasonleenaylor:
- They should still run on properly setup developer systems and these tests have not caught regressions in over a decade

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/12)


jasonleenaylor:
Adding this commit on top of the other ci work

PR #14, opened 2024-03-21
LT-21672: Add Tables to Word Export

mark-sil:
This adds the basic table data, without styling. Styling
will be added in a separate change.
Tables cannot be added to a paragraph, they need to be added
directly to a Body. So, the majority of this change was to
not create paragraphs until the end of the process; when we
are in AddEntryData().
What previously was a single paragraph for
the Entry, could now be multiple paragraphs if there are one or
more tables not at the end of the Entry.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/14)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -157,9 +157,8 @@ namespace SIL.FieldWorks.XWorks
 			StringBuilder headerTextBuilder = ConfiguredLcmGenerator.GenerateLetterHeaderIfNeeded(entry, ref lastHeader,
 				headwordWsCollator, settings, clerk);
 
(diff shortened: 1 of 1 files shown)


aror92 (approved):
Looks really nice! Generalizes well to include tables, and the paragraph handling seems flexible too--there may be other situations in future, in addition to table handling, where we'll want to use the forceNewParagraph option in AppendToParagraph.

aror92 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/14)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #17, opened 2024-03-27
LT-21672: Add styles to table cells

mark-sil:
This adds support for when a style is applied to a specific cell or when different styles are applied to different pieces of the text in a cell.

Note: The changes in SetRunStyle() also add support for styles that are applied to all or a portion of the text in other fields (notes, custom fields …).

Change-Id: Id1ca7ccfaeaad7e7ea7e781263151b1714d9d90f

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/17)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -2807,8 +2807,8 @@ namespace SIL.FieldWorks.XWorks
 				var cssStyle = CssGenerator.GenerateCssStyleFromLcmStyleSheet(style,
 					settings.Cache.WritingSystemFactory.GetWsFromStr(writingSystem), settings.PropertyTable);
 				var css = cssStyle.ToString();
-				if (!String.IsNullOrEmpty(css))
-					settings.ContentGenerator.SetRunStyle(writer, config, css);
+
+				settings.ContentGenerator.SetRunStyle(writer, config, writingSystem, css, style);
 			}
 			if (linkDestination != Guid.Empty)
 			{
@@ -3031,7 +3031,7 @@ namespace SIL.FieldWorks.XWorks
 				new ExCSS.Property("color") { Term = new HtmlColor(222, 0, 0) },
 				new ExCSS.Property("font-size") { Term = new PrimitiveTerm(UnitType.Ems, 1.5f) }
 			};
-			settings.ContentGenerator.SetRunStyle(writer, null, css.ToString());
+			settings.ContentGenerator.SetRunStyle(writer, null, writingSystem, css.ToString(), null);
 			if (text.Contains(TxtLineSplit))
 			{
 				var txtContents = text.Split(TxtLineSplit);
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -2804,11 +2804,7 @@ namespace SIL.FieldWorks.XWorks
 			}
 			if (!String.IsNullOrEmpty(style))
 			{
-				var cssStyle = CssGenerator.GenerateCssStyleFromLcmStyleSheet(style,
-					settings.Cache.WritingSystemFactory.GetWsFromStr(writingSystem), settings.PropertyTable);
-				var css = cssStyle.ToString();
-
-				settings.ContentGenerator.SetRunStyle(writer, config, writingSystem, css, style);
+				settings.ContentGenerator.SetRunStyle(writer, config, settings.PropertyTable, writingSystem, style, false);
 			}
 			if (linkDestination != Guid.Empty)
 			{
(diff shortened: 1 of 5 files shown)


jasonleenaylor (changes requested):
> Reviewed 3 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/17)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @aror92 and @mark-sil)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 2811 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/17#-Nu0EJDI4f3mCRnrf_Qd:-Nu0EJDI4f3mCRnrf_Qe:bftfmgp) ([raw file](https://github.com/sillsdev/FieldWorks/blob/9dd6df0ee80168bc831beb98456cff7787630454/Src/xWorks/ConfiguredLcmGenerator.cs#L2811)):*
> ```Smalltalk
> 				var css = cssStyle.ToString();
> 
> 				settings.ContentGenerator.SetRunStyle(writer, config, writingSystem, css, style);
> ```

I think a different refactoring might be better here. What if instead of directly calling the CssGenerator here we use the settings.StyleGenerator and pass that in. css should be renamed style in the interface, and the writing system is still necessary for word, but it would unify the design here.



_Code quote:_
```Smalltalk
				var cssStyle = CssGenerator.GenerateCssStyleFromLcmStyleSheet(style,
					settings.Cache.WritingSystemFactory.GetWsFromStr(writingSystem), settings.PropertyTable);
				var css = cssStyle.ToString();

				settings.ContentGenerator.SetRunStyle(writer, config, writingSystem, css, style);
```

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/17)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @aror92 and @jasonleenaylor)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 2811 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/17#-Nu0EJDI4f3mCRnrf_Qd:-Nu0Q3LM9juD4OBYPVZP:b-py43cl) ([raw file](https://github.com/sillsdev/FieldWorks/blob/9dd6df0ee80168bc831beb98456cff7787630454/Src/xWorks/ConfiguredLcmGenerator.cs#L2811)):*

I think a different refactoring might be better here. What if instead of directly calling the CssGenerator here we use the settings.StyleGenerator and pass that in. css should be renamed style in the interface, and the writing system is still necessary for word, but it would unify the design here.
</blockquote></details>

Hi Jason, I don't understand what you are recommending.  How should settings.StylesGenerator be used?  I'm not sure if this is relevant, but the 'config' doesn't contain the style that we want.

jasonleenaylor:
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 2811 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/17#-Nu0EJDI4f3mCRnrf_Qd:-Nu0aR-nEljNpkv-jpll:b-f2ko4u) ([raw file](https://github.com/sillsdev/FieldWorks/blob/9dd6df0ee80168bc831beb98456cff7787630454/Src/xWorks/ConfiguredLcmGenerator.cs#L2811)):*

Hi Jason, I don't understand what you are recommending.  How should settings.StylesGenerator be used?  I'm not sure if this is relevant, but the 'config' doesn't contain the style that we want.
</blockquote></details>

After our chat in slack the new plan is to make sure that we are handling the export specific code in the right class. So in this case moving the css generating into the xhtml export is the change. Getting the ILcmStylesGenerator to handle this wasn't the best route.

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/17)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @aror92)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 2811 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/17#-Nu0EJDI4f3mCRnrf_Qd:-NuUAeGD5ZKL5krXdmKs:b-988cka) ([raw file](https://github.com/sillsdev/FieldWorks/blob/3911d717e7851c72d9c8604db061ef7ec980f63c/Src/xWorks/ConfiguredLcmGenerator.cs#L2811)):*

After our chat in slack the new plan is to make sure that we are handling the export specific code in the right class. So in this case moving the css generating into the xhtml export is the change. Getting the ILcmStylesGenerator to handle this wasn't the best route.
</blockquote></details>

I’m not sure how to proceed with this.  Should I add an ‘error’ parameter to SetRunStyle() and keying off of that in the xhtml generator to create the css for large red text?

jasonleenaylor (approved):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/17)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aror92)

PR #32, opened 2024-04-26
Add BasedOn style property and fix duplicate styles

aror92:
- Add based on property when generating styles from LCM stylesheet that have based on specified in FLEx.

- Edit styleDictionary to store individual styles instead of groups of styles.

- Revise AddStyles so it adds each style one time only, using the stylenames instead of css classes as the dictionary keys.

NOTE: In the Word export, the style dictionary should save each style individually rather than a collection of styles associated with a given css class.
Otherwise, any given style may be added to the styles xml multiple times, once for each css class using that style.

Remaining task:
- Homograph-Number style is still being duplicated. It is handled in a different function and requires its own fix.

Change-Id: I9530c894fdae8251ab4211f1c7529fa63586b446

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/32)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -36,7 +36,7 @@ namespace SIL.FieldWorks.XWorks
 	{
 		private LcmCache Cache { get; }
 		private static Styles _styleSheet { get; set; } = new Styles();
-		private static Dictionary<string, Styles> _styleDictionary = new Dictionary<string, Styles>();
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/32#-:-NwfE2ExDT-qqihbgrZV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aror92)

PR #42, opened 2024-05-03
Disable borders in Word Export

aror92:
- Comment out handling of borders in Word Export

Borders do not currently display in FLEx.
But once a border is added to a style in FLEx, deselecting the border does not actually remove the border object from the FLEx style. This means if a border has ever been added in FLEx, it will still exist in the configuration node and be added in the word export. Until this is fixed, we should disable borders in the word export.


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/42)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aror92)

PR #54, opened 2024-05-10
Change parser to parse lowercase forms of uppercase words separately

jtmaxwell3:
- Change the parser to always parse a lowercase form of an uppercase wordform even if the uppercase wordform has a parse
- Change the parser to store the lowercase analysis of an uppercase wordform on the lowercase wordform
- Change RecordList to update Word Analyses window if a WfiWordformTags.kflidAnalyses change
- Change IText.PropChanged to update uppercase wordforms if their lowercase forms change

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/54)


jasonleenaylor:
*[`Src/LexText/Interlinear/InterlinDocRootSiteBase.cs` line 1072 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/54#-NxmlLMXBm1E8S1TZUVy:-NxmlLMY8Qwem7Yc8RmG:biezj4p) ([raw file](https://github.com/sillsdev/FieldWorks/blob/7aab0f2baa386c1564b9acde5ff8f75358ec8827/Src/LexText/Interlinear/InterlinDocRootSiteBase.cs#L1072)):*
> ```Smalltalk
> 							}
> 						}
> 					}
> ```

I was looking at this code and considering if a more compact solution was possible.
I don't particularly like variables with a number on the end, they don't make for easy code reading.
Linq doesn't always make for easier reading either, but it is a bit less code.
Getting the lower case versions into the wordformsToUpdate could be done similar to the following (I haven't checked if this compiles).
```m_wordformsToUpdate.UnionWith(uniqueWordForms.Where(
  wf => {
    var otherForm = wf.Form.VernacularDefaultWritingSystem;
    return otherForm != null && form != otherForm && !string.IsNullOrEmpty(wf.Text) && cf.ToLower(otherForm.Text) == form.Text;
}
```  
But that still leaves the issue of knowing if we should add a call to the idlequeue.
I guess we could check if we have changed the wordformsToUpdate count by checking the size at the beginning and end of the case statement...
What do you think?



_Code quote:_
```Smalltalk
					foreach (IWfiWordform wordform2 in uniqueWordforms)
					{
						var form2 = wordform2.Form.VernacularDefaultWritingSystem;
						if (form2 != form && form2 != null && !string.IsNullOrEmpty(form2.Text))
						{
							string sLower = cf.ToLower(form2.Text);
							if (sLower == form.Text)
							{
								m_wordformsToUpdate.Add(wordform2);
								m_mediator.IdleQueue.Add(IdleQueuePriority.High, PostponedUpdateWordforms);
							}
						}
					}
```

jtmaxwell3:
I don't know anything about Linq.  There could be more than one uppercase wordform (e.g. titlecase, uppercase, camelcase).  I would be happy to rename wordform2 to ucWordform and form2 to ucForm to make the code clearer.

jasonleenaylor (commented):
> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/54)* status: 2 of 4 files reviewed, all discussions resolved

___
*[`Src/LexText/Interlinear/InterlinDocRootSiteBase.cs` line 1072 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/54#-NxmlLMXBm1E8S1TZUVy:-Nxn5aiL9rY1a11iFuxm:b-y7es6p) ([raw file](https://github.com/sillsdev/FieldWorks/blob/7aab0f2baa386c1564b9acde5ff8f75358ec8827/Src/LexText/Interlinear/InterlinDocRootSiteBase.cs#L1072)):*

I was looking at this code and considering if a more compact solution was possible.
I don't particularly like variables with a number on the end, they don't make for easy code reading.
Linq doesn't always make for easier reading either, but it is a bit less code.
Getting the lower case versions into the wordformsToUpdate could be done similar to the following (I haven't checked if this compiles).
```m_wordformsToUpdate.UnionWith(uniqueWordForms.Where(
  wf => {
    var otherForm = wf.Form.VernacularDefaultWritingSystem;
    return otherForm != null && form != otherForm && !string.IsNullOrEmpty(wf.Text) && cf.ToLower(otherForm.Text) == form.Text;
}
```  
But that still leaves the issue of knowing if we should add a call to the idlequeue.
I guess we could check if we have changed the wordformsToUpdate count by checking the size at the beginning and end of the case statement...
What do you think?
</blockquote></details>

I think your suggested variable name updates are adequate.

jasonleenaylor (approved):
:lgtm:

> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/54#-:-NxxhY4g1B-nEzFOcVGD:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #61, opened 2024-05-22
Fix LT-21797: Missing highlights in Analyze tab

jtmaxwell3:
Implementing LT-21777 (Mark guessed analyses in InterlinearExport) broke the highlighting in the interlinear display because the extra AddProp threw the display code off.  I changed the code to mark the guessed analysis using set_StringProp instead, which required me to implement set_StringProp in CollectorEnv.

### DIFF
--- a/Src/Common/RootSite/CollectorEnv.cs
+++ b/Src/Common/RootSite/CollectorEnv.cs
@@ -92,6 +92,8 @@ namespace SIL.FieldWorks.Common.RootSites
 			public int m_tag;
 			/// <summary>Index of the current item</summary>
 			public int m_ihvo;
+			/// <summary>String properties of the current item</summary>
+			public Dictionary<int, string> m_stringProps;
 			/// <summary>Handles counting of previous occurrences of properties</summary>
 			public PrevPropCounter m_cpropPrev = new PrevPropCounter();
 
@@ -111,6 +113,7 @@ namespace SIL.FieldWorks.Common.RootSites
 				m_hvo = hvo;
 				m_tag = tag;
 				m_ihvo = ihvo;
+				m_stringProps = new Dictionary<int, string>();
 			}
 
 			/// --------------------------------------------------------------------------------
@@ -333,6 +336,10 @@ namespace SIL.FieldWorks.Common.RootSites
 		protected IFwMetaDataCache m_mdc = null;
 		/// <summary>This is used to find virtual property handlers in setting notifiers.  See LT-8245</summary>
 		protected IVwCacheDa m_cda = null;
+		/// <summary>
+		/// This is used to store string props for the next object added.
+		/// </summary>
+		protected Dictionary<int, string> m_stringProps = new Dictionary<int, string>();
 		#endregion
 
 		#region Constructor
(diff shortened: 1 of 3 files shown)


mark-sil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/61)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

___
*[`Src/Common/RootSite/CollectorEnv.cs` line 96 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/61#-Ny_QpzJ6l05ID326Oby:-Ny_QpzJ6l05ID326Obz:b-8x0wif) ([raw file](https://github.com/sillsdev/FieldWorks/blob/9d01f5f595f831c637ade409c2ac90827bd6391f/Src/Common/RootSite/CollectorEnv.cs#L96)):*
> ```Smalltalk
> 			public int m_ihvo;
> 			/// <summary>String properties of the current item</summary>
> 			public Dictionary<int, string> m_stringProps;
> ```

Are you intending to make additional changes that will take advantage of the Dictionaries that are declared here and at line 342? If not, then it is not clear to me why we are using Dictionaries.  Can you accomplish the same thing with a tagAnalysisStatus bool or enum? 



_Code quote:_
```Smalltalk
Dictionary<int, string> m_stringProps;
```

jtmaxwell3:
CollectorEnv is used by multiple clients.  None of them currently make use of set_StringProp, but I figured that as long as I was going to implement that function, I should implement it for the general case and not just for InterlinearExplorer's current need.  In the general case, set_StringProp could be called multiple times before the next AddObj, so it seemed appropriate to use a Dictionary for the properties.

mark-sil (commented):
I didn't think other clients were calling into Fieldworks code.  Do you mean multiple derived classes?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/61#-:-NyaCUxr-pUNOeZCzfpp:bmzabwb)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

jtmaxwell3:
I'm not sure what you mean by multiple derived classes.  I meant that CollectorEnv is used by ConfiguredExport, ItemsCollectorEnv, FindCollectorEnv, and DiscourseExporter in addition to InterlinearExporter.    The original comment on set_StringProp said "Nothing to do here. None of our collectors cares about string properties (yet)", so I wanted to future-proof my change.

mark-sil (approved):
Yes, those look like derived classes.  I tend toward simplifying until needed but your reasoning makes sense.  LGTM.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/61#-:-NyaMxrh2jxAKQsIpYb7:b-pp1df6)* status: all files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/61)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #84, opened 2024-06-19
Word Export Style Display Names

mark-sil:
This submission adds support for displaying the style Display Name in the Word Styles dialog (instead of displaying the style Id).

This addresses the main cases; more work is needed to handle the ‘normal’ styles, “Dictionary-Normal”, abbreviations, paragraph Styles properties, and the possibility of the same Display Name being used by different style Ids.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/84)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -163,7 +163,7 @@ namespace SIL.FieldWorks.XWorks
 				headwordWsCollator, settings, clerk);
 
 			// Create LetterHeader doc fragment and link it with the letter heading style.
-			return DocFragment.GenerateLetterHeaderDocFragment(headerTextBuilder.ToString(), WordStylesGenerator.LetterHeadingStyleName);
+			return DocFragment.GenerateLetterHeaderDocFragment(headerTextBuilder.ToString(), WordStylesGenerator.LetterHeadingDisplayName);
 		}
 
 		/*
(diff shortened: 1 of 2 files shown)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/84)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`Src/xWorks/WordStylesGenerator.cs` line 350 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/84#-O-l0rvS3ZzJil4tzJxY:-O-l4qvW7rJQfLzG9jlu:b-nmasg4) ([raw file](https://github.com/sillsdev/FieldWorks/blob/b1e079f9660091dc6fc2f46b7c8f291111cb0657/Src/xWorks/WordStylesGenerator.cs#L350)):*

So where do Character styles get handled?
</blockquote></details>

Most of the character styles are added to the _styleDictionary when they are first referenced.  An example is in AddRun().  When we add support for styles with different properties but the same DisplayName, I think it is going to be easier to move all style creation to when it's first referenced.  If we don't then I think we are going to have a difficult time trying to determine when to use a style like 'Definition1' instead of 'Definition'.

jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/84#-:-O-lPPX91brOtcKSrglX:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #91, opened 2024-06-25
LT-21787: Fix hyperlink to another project

JakeOliver28:
Fix hyperlink to another project if the target
object exists in the source project.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/91)

### DIFF
--- a/Src/xWorks/LinkListener.cs
+++ b/Src/xWorks/LinkListener.cs
@@ -438,13 +438,13 @@ namespace SIL.FieldWorks.XWorks
 		{
 			try
 			{
+				var cache = m_propertyTable.GetValue<LcmCache>("cache");
 				//Debug.Assert(!(m_lnkActive is FwAppArgs), "Beware: This will not handle link requests for other databases/applications." +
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
*[`Src/xWorks/LinkListener.cs` line 505 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/91#-O0R5k84BEJuL6iWLvs4:-O0R5k84BEJuL6iWLvs5:b4vl1o9) ([raw file](https://github.com/sillsdev/FieldWorks/blob/34cde00c366c85f665a25eee5a5a872b2809509f/Src/xWorks/LinkListener.cs#L505)):*
> ```Smalltalk
> 				// Return false if the link is to a different database
> 				var databases = m_lnkActive.PropertyTableEntries.Where(p => p.name == "database");
> 				if (databases.Any() && databases.First().name != cache.LangProject.ShortName)
> ```

This could probably be shortened with a .FirstOrDefault(), 
I see that SameDatabase doesn't work for a FwLinkArgs, this is probably as good as we can do without a bigger refactor.

PR #109, opened 2024-07-24
Word Export: Generate unique display names

mark-sil:
- Generate unique display names when the FLEX style name is different, but the style display name is the same.
- Generate the Sense number style in the same method that adds the sense number.
- Move the code that adds the sense number before and after text to the same method that adds the sense number.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/109)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/109)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1821 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/109#-O2eDXOAADeEucxDSokZ:-O2eFw5231C3g1KGA-pd:bumqff0) ([raw file](https://github.com/sillsdev/FieldWorks/blob/44a272da26a4bbd51ecca87919328151981eaa77/Src/xWorks/LcmWordGenerator.cs#L1821)):*

These locks are now superfluous. All the necessary locking is done inside of the WordStyleCollection, which is nice actually.
</blockquote></details>

The locks are needed if TryGetStyle() is called.  The TryGetStyle() call needs to be in the same lock as the associated AddStyle() call.  I originally wanted to not have a TryGetStyle() and instead do everything in AddStyle(), but AddStyle() started to get too complicated.

jasonleenaylor (commented):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/109)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

mark-sil:
added with pr #110

PR #118, opened 2024-08-01
LT-21708: Word Export - Fix Caption Character Styles

aror92:
- Update AddImageCaption method to work with an IFragment instead of a string so that caption styles are preserved.

- Merge all caption runs into a single paragraph in AddImageCaption.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/118)


aror92:
I don't love the second step of merging into a single paragraph in AddImageCaption since it requires an extra clone of all the caption runs in the Word Export.

Another way to fix this would be to create a new Append function that can be used in ConfiguredLCMGenerator (lines 1940ff) for building the caption (e.g. AppendCaption or AppendAndMergeParagraphs). But this would add some mess to the XhtmlGenerator and JsonGenerator which would need to implement the new function and just call their normal Append inside it.

mark-sil (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/118)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aror92)

PR #124, opened 2024-08-02
Make some parser tester UX improvements

jasonleenaylor:
* Move buttons from columns to button row at the bottm
* Change button text to clearer options

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/124)


jtmaxwell3 (approved):
This will need to be internationalized, but that can wait.

PR #133, opened 2024-08-08
Fix 2-column display by correcting image anchor

aror92:
The horizontal anchor type for the textframe should be set to text rather than margin.

A text type anchor prompts the textframe to move with the text it is anchored to, while a page or margin type anchor would keep the textframe fixed in a particular location on the page.

If the horizontal anchor type is set to page or margin, images can appear in the wrong column when the document layout is set to two columns.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/133)


JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/133#-:-O3nQWGT77-pi11nPuwC:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #137, opened 2024-08-12
Start using IPCFramework Nuget package

jasonleenaylor:
- Remove one more TeamCity dependency by switching to nuget

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/137)


JakeOliver28 (approved):

megahirt:
Yeah!!  Great job on moving this TC dep to NuGet!!!

Can we safely delete the TC build for this?  https://build.palaso.org/buildConfiguration/bt278#all-projects

Looks like TC is connected to https://bitbucket.org/smcconnel/ipcframework but the repo has already been moved to https://github.com/sillsdev/ipcframework
But I don't see a GHA defined here, and it's not clear where the NuGet package was built.

So it feels like are still some TODOs left, but I don't have the full story either.

PR #139, opened 2024-08-14
LT-21840: Fix missing before/between/after text and add relevant test

JakeOliver28:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/139)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -2121,6 +2121,7 @@ namespace SIL.FieldWorks.XWorks
 								// targets
 								settings.ContentGenerator.AddCollection(xw, config, IsBlockProperty(child),
 									CssGenerator.GetClassAttributeForConfig(child), content);
+								settings.StylesGenerator.AddStyles(child);
 							}
 							break;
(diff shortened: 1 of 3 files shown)


jasonleenaylor:
*[`Src/xWorks/xWorksTests/ConfiguredXHTMLGeneratorTests.cs` line 3848 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/139#-O4HfP3FDAdmhe8eYW1j:-O4HfP3G7XizUvKZa50Z:b65ats6) ([raw file](https://github.com/sillsdev/FieldWorks/blob/22a31442a362303216812911951d3dcfbbdef213/Src/xWorks/xWorksTests/ConfiguredXHTMLGeneratorTests.cs#L3848)):*
> ```Smalltalk
> 			var settings = new ConfiguredLcmGenerator.GeneratorSettings(Cache, m_propertyTable, false, false, null);
> 			//SUT
> 			settings.StylesGenerator.AddStyles(targetsNode);
> ```

If we run this line instead `ConfiguredLcmGenerator.GenerateContentForEntry(mainEntry, mainEntryNode, null, settings)` we prove the integration. I think the direct AddStyles call would work even without our changes.

PR #140, opened 2024-08-16
Word Export: specify target Word version

aror92:
* Create and add a document settings part
* In settings, specify version number corresponding to Word 2010

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -144,6 +144,39 @@ namespace SIL.FieldWorks.XWorks
 					stylePart.Styles = ((Styles)styleSheet.CloneNode(true));
 				}
 
+				// Add document settings
+				DocumentSettingsPart settingsPart = fragment.mainDocPart.DocumentSettingsPart;
+				if (settingsPart == null)
+				{
+					// Initialize word doc's settings part
+					settingsPart = AddDocSettingsPartToPackage(fragment.DocFrag);
+
+					settingsPart.Settings = new WP.Settings(
+						new Compatibility(
+							new CompatibilitySetting()
+							{
+								Name = CompatSettingNameValues.CompatibilityMode,
+								// val determines the version of word we are targeting.
+								// 14 corresponds to Office 2010; 16 would correspond to Office 2019
+								Val = new StringValue("14"),
+								Uri = new StringValue("http://schemas.microsoft.com/office/word")
+							},
+							new CompatibilitySetting()
+							{
+								// specify that table style should not be overridden
+								Name = CompatSettingNameValues.OverrideTableStyleFontSizeAndJustification,
+								Val = new StringValue("0"),
+								Uri = new StringValue("http://schemas.microsoft.com/office/word")
+							}
+							// If in the future, if we find that certain style items are different in different version of word,
+							// it may help to specify more compatibility settings.
+							// A full list of all possible compatibility settings may be found here:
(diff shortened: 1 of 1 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/140)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1838 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/140#-O4R0MFRBhTgXoxR64oC:-O4R0MFRBhTgXoxR64oD:bnkl4o9) ([raw file](https://github.com/sillsdev/FieldWorks/blob/78b41cfc3fce8ec46ff8e029c7fad136e571a7c7/Src/xWorks/LcmWordGenerator.cs#L1838)):*
> ```Smalltalk
> 			part = doc.MainDocumentPart.AddNewPart<DocumentSettingsPart>();
> 			return part;
> 		}
> ```

Is this method necessary? Is there a functional reason for calling AddDocSettingPartToPackage() as opposed to doc.MainDocumentPart.AddNewPart<DocumentSettingsPart>(), or is this more for readability/simplicity?

aror92:
*[`Src/xWorks/LcmWordGenerator.cs` line 1838 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/140#-O4R0MFRBhTgXoxR64oC:-O4jizJu6bGOsZtvsKtY:bsxhkvt) ([raw file](https://github.com/sillsdev/FieldWorks/blob/78b41cfc3fce8ec46ff8e029c7fad136e571a7c7/Src/xWorks/LcmWordGenerator.cs#L1838)):*

Is this method necessary? Is there a functional reason for calling AddDocSettingPartToPackage() as opposed to doc.MainDocumentPart.AddNewPart<DocumentSettingsPart>(), or is this more for readability/simplicity?
</blockquote></details>

That's a good point. This was to match the style of AddImagePartToPackage() and AddStylesPartToPackage(), so that we'd have similar methods for each document part we add. But in the case of document settings, it's just the one line and we likely won't need to call it more than once, so it might make more sense not to make a separate method for this. What do you think is best? I can see reasons to go either way.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/140#-:-O4k7Z4eEFk_N4ELy5mo:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1838 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/140#-O4R0MFRBhTgXoxR64oC:-O4k7TDg7-4RBJLlP4LF:bfb42ti) ([raw file](https://github.com/sillsdev/FieldWorks/blob/78b41cfc3fce8ec46ff8e029c7fad136e571a7c7/Src/xWorks/LcmWordGenerator.cs#L1838)):*

That's a good point. This was to match the style of AddImagePartToPackage() and AddStylesPartToPackage(), so that we'd have similar methods for each document part we add. But in the case of document settings, it's just the one line and we likely won't need to call it more than once, so it might make more sense not to make a separate method for this. What do you think is best? I can see reasons to go either way.
</blockquote></details>

That makes sense. Having similar methods for each document part we add sounds like a good enough reason to me.

PR #147, opened 2024-08-29
Word Export LT-21830: Fix indentation issues

mark-sil:
Many of the code changes were a result of Word not allowing nested paragraphs.  As a result, all paragraphs need to be moved to first level elements in the Body.
- Add paragraphs instead of starting a new line.
- Nested content like sub-senses still result in a new paragraph.
- Removed condition from GenerateWordStyleFromLcmStyleSheet() that was preventing nodes under senses from indenting correctly (examples).
- The basedOn value can now come from either the styles basedOn value or from the parent node.

Change-Id: Ibf9faba0ad79131100a7783d258a26429a9513d5

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/147)


jasonleenaylor:
*[`Src/xWorks/LcmWordGenerator.cs` line 2227 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/147#-O5ZYl07Av9SEKn3yibs:-O5ZYl07Av9SEKn3yibt:btf1r54) ([raw file](https://github.com/sillsdev/FieldWorks/blob/5b01bb1266d13794cf524999c1304847e79eaffa/Src/xWorks/LcmWordGenerator.cs#L2227)):*
> ```Smalltalk
> 			var workingParagraph = outputParagraph;
> 			var elements = ((DocFragment)contentToAdd).DocBody.Elements();
> 			foreach (OpenXmlElement elem in elements)
> ```

What is the behavior if we encounter two un-nestable elements back to back? If it isn't harmful we don't need to do anything really.

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/147)* status: 0 of 3 files reviewed, all discussions resolved

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1778 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/147#-O5ZV60CDnLmLzS0AIbQ:-O5ZV60CDnLmLzS0AIbR:b-5nnzfq) ([raw file](https://github.com/sillsdev/FieldWorks/blob/5b01bb1266d13794cf524999c1304847e79eaffa/Src/xWorks/LcmWordGenerator.cs#L1778)):*
> ```Smalltalk
> 					{
> 						style.BasedOn.Val = basedOnStyle.StyleId;
> 						if(continuationStyle && style.BasedOn.Val != WordStylesGenerator.NormalParagraphStyleName)
> ```

If you intend that we will fix this method to work with continuationStyle then I approve of this change, if you don't think we need it for that we should remove any use of continuationStyle after the Assert and return.

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/147)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1778 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/147#-O5ZV60CDnLmLzS0AIbQ:-O5rZugw8OFrUDVvD2ce:b-en9mjq) ([raw file](https://github.com/sillsdev/FieldWorks/blob/5b01bb1266d13794cf524999c1304847e79eaffa/Src/xWorks/LcmWordGenerator.cs#L1778)):*

If you intend that we will fix this method to work with continuationStyle then I approve of this change, if you don't think we need it for that we should remove any use of continuationStyle after the Assert and return.
</blockquote></details>

Right now, I don’t know of a situation that requires it, but I’m kind of guessing it will be needed, so I left it in.
___
*[`Src/xWorks/LcmWordGenerator.cs` line 2227 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/147#-O5ZYl07Av9SEKn3yibs:-O5rWct77cT4r1o9XqCv:baeotpg) ([raw file](https://github.com/sillsdev/FieldWorks/blob/5b01bb1266d13794cf524999c1304847e79eaffa/Src/xWorks/LcmWordGenerator.cs#L2227)):*

What is the behavior if we encounter two un-nestable elements back to back? If it isn't harmful we don't need to do anything really.
</blockquote></details>

Multiple un-nestable elements back to back is not a problem.  Each will be in it's own paragraph, then when/if we encounter nestable elements a continuation paragraph is started.

jasonleenaylor:
*[`Src/xWorks/LcmWordGenerator.cs` line 1778 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/147#-O5ZV60CDnLmLzS0AIbQ:-O5t8P-BDl-Q2o5McPee:b-8fx2u7) ([raw file](https://github.com/sillsdev/FieldWorks/blob/7b82d6e15f743237ebd87e53543910d646a31cc7/Src/xWorks/LcmWordGenerator.cs#L1778)):*

Right now, I don’t know of a situation that requires it, but I’m kind of guessing it will be needed, so I left it in.
</blockquote></details>

:+1:

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/147)* status: 0 of 3 files reviewed, all discussions resolved

PR #153, opened 2024-09-03
Word Export: Add missing Between content

mark-sil:
Removed the check for DictionaryNodeOptions.  In addition to IParaOption, sometimes we get in here for a
DictionaryNodeListOptions and sometimes it is null. Removing the check lead to another problem when adding the type. The type was always being passed false for ‘first’ when it should always be passed true.  We don’t want between content for the type Node.
This fixes many of the issues identified in LT-21808.

Change-Id: I347eaa0876f7f09b650932a3563c0c06b82d3891

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/153)


JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/153#-:-O5tg3vrDvl9axIMJXGF:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #155, opened 2024-09-04
Word Export: Add missing Before/After/Between content

mark-sil:
- Add between content for CrossRreference types.
- Add before and after content for LexicalRelations->Targets.

Note: I think passing child to AddCollection() is the correct node for the collection being added. The other implementations of AddCollection() do not use this parameter so they are not affected.

This fixes one of the issues identified in LT-21808.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/155)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -2005,9 +2005,12 @@ namespace SIL.FieldWorks.XWorks
 			var organizedRefs = SortAndFilterLexRefsAndTargets(collection, cmOwner, config);
 
 			// Now that we have things in the right order, try outputting one type at a time
+			bool first = true;
 			foreach (var referenceList in organizedRefs)
 			{
 				var xBldr = GenerateCrossReferenceChildren(config, pubDecorator, referenceList, cmOwner, settings);
+				settings.ContentGenerator.BetweenCrossReferenceType(xBldr, config, first);
+				first = false;
 				bldr.Append(xBldr);
 			}
 		}
@@ -2119,7 +2122,7 @@ namespace SIL.FieldWorks.XWorks
 							if (!content.IsNullOrEmpty())
 							{
 								// targets
-								settings.ContentGenerator.AddCollection(xw, config, IsBlockProperty(child),
+								settings.ContentGenerator.AddCollection(xw, child, IsBlockProperty(child),
 									CssGenerator.GetClassAttributeForConfig(child), content);
 								settings.StylesGenerator.AddStyles(child);
 							}
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -2005,12 +2005,12 @@ namespace SIL.FieldWorks.XWorks
 			var organizedRefs = SortAndFilterLexRefsAndTargets(collection, cmOwner, config);
 
 			// Now that we have things in the right order, try outputting one type at a time
-			bool first = true;
+			bool firstIteration = true;
 			foreach (var referenceList in organizedRefs)
 			{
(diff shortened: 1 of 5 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/155)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @mark-sil)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 2008 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/155#-O5xv4xz77Ua-Mmh3CZi:-O5xv4y-6qO6fFqhA3_w:b-485747) ([raw file](https://github.com/sillsdev/FieldWorks/blob/394b590082b669a0059a36fd7858210bad2f03d5/Src/xWorks/ConfiguredLcmGenerator.cs#L2008)):*
> ```Smalltalk
> 
> 			// Now that we have things in the right order, try outputting one type at a time
> 			bool first = true;
> ```

Could we call this firstIteration or isFirstIteration? Took me a second to figure out the purpose/meaning of this variable. And same thing for the parameter first in all the BetweenCrossReferenceType methods.
___
*[`Src/xWorks/LcmJsonGenerator.cs` line 414 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/155#-O5xvcCmDmFeIADA-bR4:-O5xvcCmDmFeIADA-bR5:bcahvlj) ([raw file](https://github.com/sillsdev/FieldWorks/blob/394b590082b669a0059a36fd7858210bad2f03d5/Src/xWorks/LcmJsonGenerator.cs#L414)):*
> ```Smalltalk
> 		}
> 
> 		public void BetweenCrossReferenceType(IFragment content, ConfigurableDictionaryNode node, bool first)
> ```

Is this method necessary? We just implement it because of the method being added to the interface, correct? Do we need to add the method to the interface, or could we only implement it in the class that uses it?

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/155)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @JakeOliver28)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 2008 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/155#-O5xv4xz77Ua-Mmh3CZi:-O5y6EX7ESifT1itC-59:b-prwill) ([raw file](https://github.com/sillsdev/FieldWorks/blob/394b590082b669a0059a36fd7858210bad2f03d5/Src/xWorks/ConfiguredLcmGenerator.cs#L2008)):*

Could we call this firstIteration or isFirstIteration? Took me a second to figure out the purpose/meaning of this variable. And same thing for the parameter first in all the BetweenCrossReferenceType methods.
</blockquote></details>

I change the name based on the context.  Sometimes firstIteration and sometimes firstItem.
___
*[`Src/xWorks/LcmJsonGenerator.cs` line 414 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/155#-O5xvcCmDmFeIADA-bR4:-O5y3vf04YXlybvw3sQw:buehcoo) ([raw file](https://github.com/sillsdev/FieldWorks/blob/394b590082b669a0059a36fd7858210bad2f03d5/Src/xWorks/LcmJsonGenerator.cs#L414)):*

Is this method necessary? We just implement it because of the method being added to the interface, correct? Do we need to add the method to the interface, or could we only implement it in the class that uses it?
</blockquote></details>

The caller doesn't know that it is calling a LcmWordGenerator, All it knows is that it is calling some type of ILcmContentGenerator, so to continue with the existing pattern we need to use the interface.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/155#-:-O5yEg5B8W3AizxJzqui:bnfp4nl)* status: 0 of 5 files reviewed, 2 unresolved discussions

PR #157, opened 2024-09-05
Word Export: Prevent extra Before/After content for senses

mark-sil:
When senses are in separate paragraphs we were still adding the before and after content that was around the collection of senses.  It is now only added when not in separate paragraphs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/157)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -1633,8 +1633,12 @@ namespace SIL.FieldWorks.XWorks
 
 		public IFragment WriteProcessedSenses(ConfigurableDictionaryNode config, bool isBlock, IFragment senseContent, string className, IFragment sharedGramInfo)
 		{
-			// Add Before text for the senses.
-			if (!string.IsNullOrEmpty(config.Before))
+			bool eachInAParagraph = config != null &&
+									config.DictionaryNodeOptions is DictionaryNodeSenseOptions &&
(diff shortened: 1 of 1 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/157)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mark-sil)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1638 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/157#-O61l1nYEqQO9Dr_HUYZ:-O61l1nYEqQO9Dr_HUY_:b-mahkeo) ([raw file](https://github.com/sillsdev/FieldWorks/blob/ac90ccdb05dd327fe6c5581c85a94b8854921d03/Src/xWorks/LcmWordGenerator.cs#L1638)):*
> ```Smalltalk
> 			bool eachInAParagraph = config != null &&
> 									config.DictionaryNodeOptions is DictionaryNodeSenseOptions &&
> 									((DictionaryNodeSenseOptions)(config.DictionaryNodeOptions)).DisplayEachSenseInAParagraph;
> ```

Would this be clearer/more concise?



_Code snippet:_
```Smalltalk
var senseOptions = config?.DictionaryNodeOptions as DictionaryNodeSenseOptions;
bool eachInAParagraph = senseOptions?.DisplayEachSenseInAParagraph ?? false;
```

mark-sil:
closing

PR #165, opened 2024-09-23
Fix LT-21888: boundary marker name can be deleted when editing rules

jtmaxwell3:
I have a fix for this, but it might not be the best solution.

If you select a boundary marker (or any phoneme) in a phonological rule and type a character, its name gets deleted but the boundary marker does not.  I couldn't track down where in the code this happens.  I prevented this from happening by suppressing all characters except for Backspace and Delete.  This means that the user can only enter phonemes by using the menu provided.  This feels like overkill, but Matthew is OK with it if we eventually add code to allow phonemes to be entered by typing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/165)

### DIFF
--- a/Src/LexText/LexTextControls/PatternView.cs
+++ b/Src/LexText/LexTextControls/PatternView.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Windows.Forms;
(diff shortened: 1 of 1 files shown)


jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/165)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

___
*[`Src/LexText/LexTextControls/PatternView.cs` line 141 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/165#-O7_4AzK8GV5lqyX5CuS:-O7_4AzK8GV5lqyX5CuT:b-jwdmmc) ([raw file](https://github.com/sillsdev/FieldWorks/blob/34e36211a8e96e562536b497a4441736c874cb86/Src/LexText/LexTextControls/PatternView.cs#L141)):*
> ```Smalltalk
> 				if (RemoveItemsRequested != null)
> 					RemoveItemsRequested(this, new RemoveItemsRequestedEventArgs(false));
> 				e.Handled = true;
> ```

For correctness move e.Handled = true above the if. By returning and ignoring we are 'handling' the key press, and the event should  reflect this.

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/165)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #174, opened 2024-10-04
Fix LT-21839: Avoid crash and allow paste success with no ws

jasonleenaylor:
Analyzing the data that caused the crash showed that in some edge
   cases no writing system can be identified in MakeSubString.
   This could cause an empty builder to have no writing system.
   I couldn't identify any actual side effects in this edge case so
   I left the assert, but allowed the paste to succeed

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/174)

### DIFF
--- a/Src/views/VwSelection.cpp
+++ b/Src/views/VwSelection.cpp
@@ -5315,6 +5315,7 @@ void VwTextSelection::MakeSubString(ITsString * ptss, int ichMin, int ichLim, IT
 	int cch;
 	CheckHr(ptss->GetBldr(&qtsb));
 	CheckHr(ptss->get_Length(&cch));
+
 	if (ichLim < cch)
 		CheckHr(qtsb->Replace(ichLim, cch, NULL, NULL));
 	if (ichMin)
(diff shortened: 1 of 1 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/174)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor)

___
*[`Src/views/VwSelection.cpp` line 5358 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/174#-O8bwfSy-e0RdBfkzSLN:-O8bwfSy-e0RdBfkzSLO:bgkjjs7) ([raw file](https://github.com/sillsdev/FieldWorks/blob/5f289b40522fb1955c8d4e18a26983b2aaca303f/Src/views/VwSelection.cpp#L5358)):*
> ```cpp
> 			}
> 
> 			if(wsNew <= 0)
> ```

Could we just put the Assert and return inside of the first "if (wsNes <=0)" instead of having a duplicate if statement?

jasonleenaylor:
*[`Src/views/VwSelection.cpp` line 5358 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/174#-O8bwfSy-e0RdBfkzSLN:-O8cHgwvDYhmYHRLT_IU:b-v2lib) ([raw file](https://github.com/sillsdev/FieldWorks/blob/5f289b40522fb1955c8d4e18a26983b2aaca303f/Src/views/VwSelection.cpp#L5358)):*

Hmm, I retract, we can still put this inside the first if.
</blockquote></details>

After drinking more coffee I retract my retraction, if we want it to 'return' under the correct conditions a second check is necessary even if we put it inside the first if. So this way reduces the nesting.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/174#-:-O8cMyL23aR8fb3aOTuI:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion

PR #184, opened 2024-10-16
Word Export: prevent cross-thread exception

mark-sil:
When disposing the ClerkActivator do not update the status bar in the call to ActivateUI().
XHTML export is also calling this from a different thread but we were not seeing this error because the text string was the same as the current value so no change was made.
Notes:
In OnExport(), after the command completes, we call ActivateUI() a second time, and this time we do update the status bar.
The exception was only happening when there were reversals and it seemed to only/mostly happen in debug builds.

Change-Id: Ife8957181e5a07736b1f56e2f488aa92c8f226c5

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/184)


jasonleenaylor:
*[`Src/xWorks/DictionaryExportService.cs` line 194 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/184#-O9LyXpB3Xh9kZH3nzPm:-O9LyXpC2py9Ga4ru9iT:b-rxe7z8) ([raw file](https://github.com/sillsdev/FieldWorks/blob/f73d4843b28972feeabfa9bdb47e74c802f9ba1c/Src/xWorks/DictionaryExportService.cs#L194)):*
> ```Smalltalk
> 				System.Diagnostics.Debug.WriteLineIf(!disposing, "****** Missing Dispose() call for " + GetType() + " ******");
> 				if (disposing && m_currentClerk != null && !m_currentClerk.IsDisposed)
> 					m_currentClerk.ActivateUI(true, false);
> ```

So if this is an exception that is happening because it is trying to modify winforms UI code from a different thread there is a more robust solution closer to where the control is modified. Every control has an InvokeRequired property that tells you if you are trying to execute something on the wrong thread. You can then use an Invoke method to get it to execute the code you want.

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/184)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`Src/xWorks/DictionaryExportService.cs` line 194 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/184#-O9LyXpB3Xh9kZH3nzPm:-OG_f5A89KKsUDNGqt_r:b-ombivp) ([raw file](https://github.com/sillsdev/FieldWorks/blob/18ca0696ddc4bdfc41071141d149f75648209d84/Src/xWorks/DictionaryExportService.cs#L194)):*

So if this is an exception that is happening because it is trying to modify winforms UI code from a different thread there is a more robust solution closer to where the control is modified. Every control has an InvokeRequired property that tells you if you are trying to execute something on the wrong thread. You can then use an Invoke method to get it to execute the code you want.
</blockquote></details>

I implemented the solution you suggested.  Much better, thanks!

mark-sil (commented):
I implemented the solution you suggested.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/184#-:-OG_fjQA4X7wVtYBr4IA:b-6c1yyo)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

jasonleenaylor (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/184#-:-OGk1yPe0ih56Km979pM:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #188, opened 2024-10-25
Fix LT-21088: Yellow Box while parsing words after promoting category

jtmaxwell3:
I fixed LT-21088 by locally defining slots that are in the wrong scope.  This keeps XAmple from crashing, but it may or may not parse templates that have slots in the wrong scope correctly.  Andy thinks that the right solution is for users to define the scopes of the slots correctly.  I'd like to report the problem to the user somehow but I don't see an easy way to do that from ParserCore.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/188)

### DIFF
--- a/Build/mkall.targets
+++ b/Build/mkall.targets
@@ -287,7 +287,7 @@
 		<ParatextNugetVersion>9.4.0.1-beta</ParatextNugetVersion>
 		<LcmNugetVersion>11.0.0-beta0104</LcmNugetVersion>
 		<IcuNugetVersion>70.1.123</IcuNugetVersion>
-		<HermitCrabNugetVersion>3.3.0</HermitCrabNugetVersion>
+		<HermitCrabNugetVersion>3.4.1</HermitCrabNugetVersion>
 		<IPCFrameworkVersion>1.1.1-beta0001</IPCFrameworkVersion>
 		<!-- bt393 is the master branch build of ExCss for Windows development. Update when appropriate. -->
 		<ExCssBuildType Condition="'$(OS)'=='Windows_NT'">bt393</ExCssBuildType>
--- a/Build/nuget-common/packages.config
+++ b/Build/nuget-common/packages.config
@@ -66,8 +66,8 @@
   <package id="SIL.libpalaso.l10ns" version="6.0.0" targetFramework="net461" />
   <package id="SIL.Lift" version="13.0.0-beta0076"  targetFramework="net461" />
   <package id="SIL.Media" version="13.0.0-beta0076"  targetFramework="net461" />
-  <package id="SIL.Machine" version="3.3.0"  targetFramework="netstandard2.0" />
-  <package id="SIL.Machine.Morphology.HermitCrab" version="3.3.0"  targetFramework="netstandard2.0" />
+  <package id="SIL.Machine" version="3.4.1"  targetFramework="netstandard2.0" />
+  <package id="SIL.Machine.Morphology.HermitCrab" version="3.4.1"  targetFramework="netstandard2.0" />
   <package id="SIL.ParatextShared" version="7.4.0.1"  targetFramework="net40" /> <!-- REVIEW (Hasso) 2023.05: do we still integrate with PT 7? -->
   <package id="SIL.Scripture" version="13.0.0-beta0076"  targetFramework="net461" />
   <package id="SIL.TestUtilities" version="13.0.0-beta0076"  targetFramework="net461" />
--- a/Src/Common/Controls/DetailControls/DataTree.cs
+++ b/Src/Common/Controls/DetailControls/DataTree.cs
@@ -1192,7 +1192,7 @@ namespace SIL.FieldWorks.Common.Framework.DetailControls
 				DeepSuspendLayout();
 				// NB: The ArrayList created here can hold disparate objects, such as XmlNodes and ints.
 				if (m_root != null)
-					CreateSlicesFor(m_root, null, null, null, 0, 0, new ArrayList(20), new ObjSeqHashMap(), null);
+					CreateSlicesFor(m_root, null, null, null, 0, 0, new ArrayList(20), null);
 			}
 			finally
 			{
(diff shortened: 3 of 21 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/188)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

___
*[`Src/LexText/ParserCore/M3ToXAmpleTransformer.cs` line 133 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/188#-OAIqynT3u95YSuZAtWN:-OAIqynU9QpvCshDc0wN:b1s0gqa) ([raw file](https://github.com/sillsdev/FieldWorks/blob/a097b67d7b49c9284a92e9414be48829ee350706/Src/LexText/ParserCore/M3ToXAmpleTransformer.cs#L133)):*
> ```Smalltalk
> 		}
> 
> 		private void GetUndefinedSlots(XElement element, ISet<string> undefinedSlots)
> ```

Could we do this without recursion?

Here's an alternative way of doing it that I believe may work. DescendantsAndSelf() gets a flat structure that can be simply iterated over. Unsure if the order is fine without recursion based on your last comment in this method.



_Code snippet:_
```Smalltalk
private void GetUndefinedSlots(XElement element, ISet<string> undefinedSlots)
{
	// Iterate over all descendant elements in a flat structure.
	foreach (XElement element in rootElement.DescendantsAndSelf())
	{
		// Record slots encountered.
		if (element.Name == "PrefixSlots" || element.Name == "SuffixSlots")
		{
			undefinedSlots.Add((string)element.Attribute("dst"));
		}

		// Check and remove slots that are defined at this level.
		if (element.Name == "AffixSlots")
		{
			foreach (XElement slot in element.Elements())
			{
				undefinedSlots.Remove((string)slot.Attribute("Id"));
			}
		}
	}
}
```

jtmaxwell3:
I think that the flat version may get the wrong result because AffixSlots has to be processed after PrefixSlots and SuffixSlots.  I would prefer to keep the recursive version because it is more obvious whether this constraint is preserved.

JakeOliver28 (commented):
Could we do two passes then? The recursive code feels less readable/easily understandable to me.

```
private void GetUndefinedSlots(XElement rootElement, ISet<string> undefinedSlots)
{
	var allElements = rootElement.DescendantsAndSelf();	

	// First pass: Add slots encountered in PrefixSlots or SuffixSlots
	foreach (XElement element in allElements)
	{
		if (element.Name == "PrefixSlots" || element.Name == "SuffixSlots")
		{
			undefinedSlots.Add((string)element.Attribute("dst"));
		}
	}

	// Second pass: Remove any slots that are defined in AffixSlots elements
	foreach (XElement element in allElements)
	{
		if (element.Name == "AffixSlots")
		{
			foreach (XElement slot in element.Elements())
			{
				undefinedSlots.Remove((string)slot.Attribute("Id"));
			}
		}
	}
}
```

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/188#-:-OAIzf-j8lUYS4mtr1nY:bhofycz)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/188)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jtmaxwell3)

___
*[`Src/LexText/ParserCore/M3ToXAmpleTransformer.cs` line 118 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/188#-OATDF3V5DkJasDfnpl7:-OATDF3V5DkJasDfnpl8:b-m2n0wy) ([raw file](https://github.com/sillsdev/FieldWorks/blob/070009e810c565ea78826ef3429a7fbf1e6f8276/Src/LexText/ParserCore/M3ToXAmpleTransformer.cs#L118)):*
> ```Smalltalk
> 				return;
> 			// Add undefined slots to AffixSlots.
> 			foreach (XElement elem in templateElem.Elements())
> ```

We can get rid of the outer foreach loop by grabbing the first element with "AffixSlots" as its name.

We could also get rid of the inner foreach loop with LINQ, although that change is arguably beneficial.



_Code snippet:_
```Smalltalk
	XElement affixSlotsElem = templateElem.Element("AffixSlots");
    if (affixSlotsElem == null)
        return;

    affixSlotsElem.Add(CreateAffixSlotElements(undefinedSlots));
    
    
   
// Separate helper method for readability
private IEnumerable<XElement> CreateAffixSlotElements(IEnumerable<string> slotIds)
{
    return slotIds.Select(slotId => new XElement("MoInflAffixSlot", new XAttribute("Id", slotId)));
}
```

jtmaxwell3:
Two flat passes gets the wrong result.  We are trying to determine whether there is a variable that is used but has no definition in scope.  So we have to use recursion to compute scope.

I don't understand your second suggestion, so I can't verify if it is correct.  I find LINQ code hard to read.

I don't see what is wrong with my original code.

JakeOliver28 (commented):
Sounds good!

Just wanting to avoid nested foreach loops & an if statement if possible. For my second suggestion, we can leave out the LINQ and still remove the outer foreach loop.

```
XElement affixSlotsElem = templateElem.Element("AffixSlots");
if (affixSlotsElem == null)
   return;

foreach (string slotId in undefinedSlots)
{
	XElement slot = new XElement("MoInflAffixSlot");
	slot.SetAttributeValue("Id", slotId);
	affixSlotsElem.Add(slot);
}
```

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/188#-:-OATMyE--hSpaa6qNgn5:bfx2813)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jtmaxwell3)

jtmaxwell3:
I removed the extra for loop and made the documentation clearer.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/188#-:-OATaJ7qEMZYXRP7hRXm:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jtmaxwell3)

PR #192, opened 2024-10-29
Word Export LT-21673: Add support for bidi text

mark-sil:
- Use the default vernacular writing system for before/after/between content.  This causes the order of many runs to be correct and causes characters like parentheses to be display correctly.
- For situations where two runs that are both lt2 we now add a run between them that is rtl.  This causes the order to be switched so they appear the same as in FLEX.
- Sense numbers now generate before/after content the same as all other before/after content.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/192)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -42,6 +42,7 @@ namespace SIL.FieldWorks.XWorks
 		private ReadOnlyPropertyTable _propertyTable;
 		internal const int maxImageHeightInches = 1;
 		internal const int maxImageWidthInches = 1;
+		public static bool IsBidi { get; private set; }
 
 		public LcmWordGenerator(LcmCache cache)
 		{
@@ -63,8 +64,9 @@ namespace SIL.FieldWorks.XWorks
 				var readOnlyPropertyTable = new ReadOnlyPropertyTable(propertyTable);
 
 				generator.Init(readOnlyPropertyTable);
+				IsBidi = ConfiguredLcmGenerator.IsEntryStyleRtl(readOnlyPropertyTable, configuration);
 				var settings = new ConfiguredLcmGenerator.GeneratorSettings(cache, readOnlyPropertyTable, false, true, System.IO.Path.GetDirectoryName(filePath),
-							ConfiguredLcmGenerator.IsEntryStyleRtl(readOnlyPropertyTable, configuration), System.IO.Path.GetFileName(cssPath) == "configured.css")
+							IsBidi, System.IO.Path.GetFileName(cssPath) == "configured.css")
 							{ ContentGenerator = generator, StylesGenerator = generator};
 				settings.StylesGenerator.AddGlobalStyles(configuration, readOnlyPropertyTable);
 				string lastHeader = null;
@@ -203,7 +205,9 @@ namespace SIL.FieldWorks.XWorks
 			}
 		}
 
-		internal static IFragment GenerateLetterHeaderIfNeeded(ICmObject entry, ref string lastHeader, Collator headwordWsCollator, ConfiguredLcmGenerator.GeneratorSettings settings, ReadOnlyPropertyTable propertyTable, LcmStyleSheet mediatorStyleSheet, RecordClerk clerk = null)
+		internal static IFragment GenerateLetterHeaderIfNeeded(ICmObject entry, ref string lastHeader, Collator headwordWsCollator,
+			ConfiguredLcmGenerator.GeneratorSettings settings, ReadOnlyPropertyTable propertyTable, LcmStyleSheet mediatorStyleSheet,
+			RecordClerk clerk = null)
 		{
 			StringBuilder headerTextBuilder = ConfiguredLcmGenerator.GenerateLetterHeaderIfNeeded(entry, ref lastHeader,
 				headwordWsCollator, settings, clerk);
@@ -446,7 +450,7 @@ namespace SIL.FieldWorks.XWorks
 			/// </summary>
 			/// <param name="run">The run to append.</param>
 			/// <param name="forceNewParagraph">Even if a paragraph exists, force the creation of a new paragraph.</param>
-			public void AppendToParagraph(IFragment fragToCopy, OpenXmlElement run, bool forceNewParagraph)
+			public void AppendToParagraph(IFragment fragToCopy, Run run, bool forceNewParagraph)
 			{
 				WP.Paragraph lastPar = null;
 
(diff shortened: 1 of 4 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/192)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mark-sil)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 501 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-OAOaXDZ3ZSoKteKog7_:-OAOaXDZ3ZSoKteKog7a:b-3ah3tv) ([raw file](https://github.com/sillsdev/FieldWorks/blob/57a29f5bfcd77eaa2d9b9f9e7042804a8edc324c/Src/xWorks/LcmWordGenerator.cs#L501)):*
> ```Smalltalk
> 					// then add a run between them that is rtl and contains a rtl character.
> 					// This is needed to get the runs to switch their order in Word.
> 					if (IsBidi && run.RunProperties?.RightToLeftText == null)
> ```

Would this work? It limits the if statements and makes boolean expressions more readable, but I'm not sure if creating childRuns and previousRun everytime would be acceptable/efficient. We could eliminate the comments above with the newly named booleans.



_Code snippet:_
```Smalltalk
var childRuns = lastPar?.Elements<Run>();
var previousRun = childRuns?.Last();

bool newRunNotRTL = run.RunProperties?.RightToLeftText == null;
bool paragraphContainsRuns = childRuns?.Any() ?? false;
bool previousRunNotRTL = previousRun?.RunProperties?.RightToLeftText == null;

if (IsBidi && newRunNotRTL && paragraphContainsRuns && previousRunNotRTL)
{
	// Add a unicode RTL mark between two runs that are not rtl.
	var rtlRun = new WP.Run();
	rtlRun.AppendChild(new WP.Text("\u200f"));
	rtlRun.RunProperties = new RunProperties();
	rtlRun.RunProperties.RightToLeftText = new RightToLeftText();
	lastPar.AppendChild(rtlRun);
}
```

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/192)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @JakeOliver28)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 501 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-OAOaXDZ3ZSoKteKog7_:-OAOkd_O0tXQl3Z8Vf9l:b3hs8vx) ([raw file](https://github.com/sillsdev/FieldWorks/blob/57a29f5bfcd77eaa2d9b9f9e7042804a8edc324c/Src/xWorks/LcmWordGenerator.cs#L501)):*

Would this work? It limits the if statements and makes boolean expressions more readable, but I'm not sure if creating childRuns and previousRun everytime would be acceptable/efficient. We could eliminate the comments above with the newly named booleans.
</blockquote></details>

I don't think we should get childRuns and previousRun until we need them. I started to integrate the bool's into the existing 'if' structure but it didn't seem to help readability.

JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/192)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @mark-sil)

___
*[`Src/xWorks/WordStyleCollection.cs` line 52 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-OAOyR-W7Co3AaZVbr2e:-OAOyR-W7Co3AaZVbr2f:bdjzi79) ([raw file](https://github.com/sillsdev/FieldWorks/blob/57a29f5bfcd77eaa2d9b9f9e7042804a8edc324c/Src/xWorks/WordStyleCollection.cs#L52)):*
> ```Smalltalk
> 			{
> 				// Get an enumerator to the flattened list of all StyleElements.
> 				var styleElementsEnumerator = styleDictionary.Values.SelectMany(x => x);
> ```

We could use LINQ instead of a foreach loop. Would this be more readable?



_Code snippet:_
```Smalltalk
return styleDictionary.Values
                              .SelectMany(x => x)
                              .FirstOrDefault(styleElement => styleElement.Style.StyleId == uniqueDisplayName);
```

jasonleenaylor:
*[`Src/xWorks/LcmWordGenerator.cs` line 45 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-OAP98Y71e0vFKNPOtu7:-OAP98Y71e0vFKNPOtu8:bmfccex) ([raw file](https://github.com/sillsdev/FieldWorks/blob/57a29f5bfcd77eaa2d9b9f9e7042804a8edc324c/Src/xWorks/LcmWordGenerator.cs#L45)):*
> ```Smalltalk
> 		internal const int maxImageHeightInches = 1;
> 		internal const int maxImageWidthInches = 1;
> 		public static bool IsBidi { get; private set; }
> ```

I see this static member, but then also it looks like it is passed into the GeneratorSettings. Do we need the static?

jasonleenaylor:
*[`Src/xWorks/LcmWordGenerator.cs` line 2688 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-OAP9MXh-rPpbewbpAeU:-OAP9MXh-rPpbewbpAeV:b-v6kelf) ([raw file](https://github.com/sillsdev/FieldWorks/blob/57a29f5bfcd77eaa2d9b9f9e7042804a8edc324c/Src/xWorks/LcmWordGenerator.cs#L2688)):*
> ```Smalltalk
> 		{
> 			var runProp = new RunProperties(new RunStyle() { Val = uniqueDisplayName });
> 			if (IsBidi)
> ```

Ah, here is where you currently need that static I was wondering about..

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/192)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @JakeOliver28 and @jasonleenaylor)

___
*[`Src/xWorks/WordStyleCollection.cs` line 52 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-OAOyR-W7Co3AaZVbr2e:-OASNHAT21EP7k4Au8CH:b-896fix) ([raw file](https://github.com/sillsdev/FieldWorks/blob/57a29f5bfcd77eaa2d9b9f9e7042804a8edc324c/Src/xWorks/WordStyleCollection.cs#L52)):*

We could use LINQ instead of a foreach loop. Would this be more readable?
</blockquote></details>

Done.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/192#-:-OAT3u8G6Tu7e8p5JMX9:bnfp4nl)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @jasonleenaylor)

PR #193, opened 2024-10-31
Word Export LT-21673: Fix font problems for bidi text

mark-sil:
For some writing systems Word was not using the correct font or correct font size.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/193)

### DIFF
--- a/Src/xWorks/WordStylesGenerator.cs
+++ b/Src/xWorks/WordStylesGenerator.cs
@@ -507,9 +507,13 @@ namespace SIL.FieldWorks.XWorks
 
 			if (fontName != null)
 			{
-				// Note: if desired, multiple fonts can be used for different text types in a single run
-				// by separately specifying font names to use for ASCII, High ANSI, Complex Script, and East Asian content.
-				var font = new RunFonts(){Ascii = fontName};
+				var font = new RunFonts()
+				{
+					Ascii = fontName,
+					HighAnsi = fontName,
+					ComplexScript = fontName,
+					EastAsia = fontName
+				};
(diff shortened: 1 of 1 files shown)


aror92 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/193)* status: all files reviewed, 1 unresolved discussion (waiting on @mark-sil)

___
*[`Src/xWorks/WordStylesGenerator.cs` line 510 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/193#-OAXr2iCB_EGJgQdZtEo:-OAYAJ0UDdrqFgJFUUuW:b6dynaq) ([raw file](https://github.com/sillsdev/FieldWorks/blob/065d2fb1a449505c784ffeaa60a8e0a493549c09/Src/xWorks/WordStylesGenerator.cs#L510)):*

@aror92 It is not always necessary to set all four font names and to set the font size for complex scripts, but I think it is probably safest. I'm not sure of a reliable way to figure out when they are needed, and it's a relatively small amount of data so it probably doesn't matter.
</blockquote></details>

I also can't think of a good way to check which font type a run should use. Of course some fonts are not allowed to be used for certain types (e.g. Scheherazade can't be Ascii), but I think this will still work fine in Word. And the way we create the runs & fonts, we shouldn't run into situations where a run with a strictly ComplexScript fontName would have characters that would actually use the specified Ascii font anyway. I do wonder if we may see some issues related to this in LibreOffice or when importing to InDesign, but we can investigate that later.

PR #194, opened 2024-10-31
Word Export LT-21673: bidi problem with switching run order

mark-sil:
To be consistent with how FLEX displays consecutive left-to-right runs, we do not want to switch their order. Remove the code that was doing the switching and causing the display to be different between Word and FLEX.

Change-Id: I396a54c88e9952112e396f0157f6420723282dff

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/194)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -491,30 +491,6 @@ namespace SIL.FieldWorks.XWorks
 				else
 				{
 					lastPar = GetLastParagraph();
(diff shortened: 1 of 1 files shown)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/194)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aror92)

PR #195, opened 2024-10-31
Fix LT-21971: Crash adding a Feature Set

jtmaxwell3:
The crash occurred in RemoveFeatureStructureFromMSA because it tried to create a new non-undoable unit of work inside a PropChanges broadcast.  The property change was caused when DefaultCreateNewUIObject  created a new object inside an undoable unit of work.  DataTree.PropChanged called DataTree.RefreshList, which removed the existing slices which removed an existing feature structure.

I noticed that there was nothing to be done in the non-undoble unit of work because m_obj's ClassID didn't match.  So, I changed the code to avoid creating the non-undoable unit of work in this case.  This allowed the feature structure to be created without crashing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/195)

### DIFF
--- a/Src/LexText/Lexicon/MsaInflectionFeatureListDlgLauncherSlice.cs
+++ b/Src/LexText/Lexicon/MsaInflectionFeatureListDlgLauncherSlice.cs
@@ -105,6 +105,12 @@ namespace SIL.FieldWorks.XWorks.LexEd
 		{
 			if (m_obj != null)
 			{
+				if (m_obj.ClassID != MoStemMsaTags.kClassId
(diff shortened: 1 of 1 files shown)


jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/195#-:-OAYhhwrEe-_Yfvn6nQ0:bnfp4nl)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @jtmaxwell3)

PR #199, opened 2024-11-07
Fix LT-21737: Make highlight button work

jasonleenaylor:
* Generate xhtml with a nodeId attribute that uses the ConfigurableDictionaryNode hash and use that to set highlight
* Modify the 'Configure Dictionary' right click menu to use the nodeId attribute to select the correct tree node

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/199)

### DIFF
--- a/Src/xWorks/ConfigurableDictionaryNode.cs
+++ b/Src/xWorks/ConfigurableDictionaryNode.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2014-2017 SIL International
+// Copyright (c) 2014-2017 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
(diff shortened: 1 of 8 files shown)


aror92 (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/199)* status: all files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor)

___
*[`Src/xWorks/DictionaryConfigurationController.cs` line 1589 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/199#-OBXb3rP1gMsVqGBHBBQ:-OBXb3rP1gMsVqGBHBBR:b-m0limm) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2d05f6fa6ea243994cd065cb95c7c58cb46a8db9/Src/xWorks/DictionaryConfigurationController.cs#L1589)):*
> ```Smalltalk
> 					}
> 				}
> 				if (topNode == null)
> ```

Why is it no longer necessary to check if topNode is null & return in that case?

PR #200, opened 2024-11-11
Fix LT-21933: Picture thumbnail is not showing configuration dialog

jtmaxwell3:
This bug was introduced on April 12th.  I reverted the change that caused the bug and then added a useUri parameter to GenerateSettings to get the desired behavior for Word after consulting with Ariel.

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -861,7 +861,7 @@ namespace SIL.FieldWorks.XWorks
 			{
 				filePath = MakeSafeFilePath(file.AbsoluteInternalPath);
 			}
-			return filePath;
+			return (settings.UseRelativePaths || !settings.UseUri) ? filePath : new Uri(filePath).ToString();
 		}
 
 		private static string GenerateSrcAttributeForMediaFromFilePath(string filename, string subFolder, GeneratorSettings settings)
@@ -3175,6 +3175,8 @@ namespace SIL.FieldWorks.XWorks
 			public LcmCache Cache { get; }
 			public ReadOnlyPropertyTable PropertyTable { get; }
 			public bool UseRelativePaths { get; }
+
+			public bool UseUri { get; }
 			public bool CopyFiles { get; }
 			public string ExportPath { get; }
 			public bool RightToLeft { get; }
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/200#-:-OBSHTquDryn8-Qdi5VJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

aror92 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/200)* status: all files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 3196 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/200#-OBVwMiV6yKf_bTDSLsH:-OBVwMiV6yKf_bTDSLsI:b-eyrpay) ([raw file](https://github.com/sillsdev/FieldWorks/blob/03d1073ef92663467f92090736350e65f095a2e4/Src/xWorks/ConfiguredLcmGenerator.cs#L3196)):*
> ```Smalltalk
> 			}
> 
> 			public GeneratorSettings(LcmCache cache, ReadOnlyPropertyTable propertyTable, bool relativePaths, bool useUri, bool copyFiles, string exportPath, bool rightToLeft = false, bool isWebExport = false, bool isTemplate = false)
> ```

GeneratorSettings is also used by the LcmJsonGenerator and LcmXhtmlGenerator. This change to the arguments needs to be handled in those files as well.

I see some potential for confusion since the decision whether or not to useUri still partly depends on the value of relativePaths. (i.e. if relativePaths is true AND useUri is true, the Uri won't actually be used.) It may be better to call this something like "forceToUseFilepath" or "forceNotToUseUri", set the value to True for the word export, and change the behavior to "return (settings.UseRelativePaths || settings.forceNotToUseUri) ? filePath : new Uri(filePath).ToString();"

aror92 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/200#-:-OBW0odKELUZCcNpVd5u:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 3196 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/200#-OBVwMiV6yKf_bTDSLsH:-OBW0G8W1YBUz_L2XjG1:bp8oh7k) ([raw file](https://github.com/sillsdev/FieldWorks/blob/03d1073ef92663467f92090736350e65f095a2e4/Src/xWorks/ConfiguredLcmGenerator.cs#L3196)):*

GeneratorSettings is also used by the LcmJsonGenerator and LcmXhtmlGenerator. This change to the arguments needs to be handled in those files as well.

I see some potential for confusion since the decision whether or not to useUri still partly depends on the value of relativePaths. (i.e. if relativePaths is true AND useUri is true, the Uri won't actually be used.) It may be better to call this something like "forceToUseFilepath" or "forceNotToUseUri", set the value to True for the word export, and change the behavior to "return (settings.UseRelativePaths || settings.forceNotToUseUri) ? filePath : new Uri(filePath).ToString();"
</blockquote></details>

Ah, nevermind, I missed that you made an alternate constructor.

PR #222, opened 2024-12-05
LT-21882: Webonary: Fix style names

mark-sil:
References to styles in the xhtml were not matching up with the correct style names in the css.

- In the Json generator we were not using the _styleDictionary to generate the css.  Instead we were re-generating the css from the configuration. Now use the _styleDictionary to generate the css.
- In the Json generator there were two _styleDictionaries being used. They were filled at different times and could have different values because of unique name generation. Now use one _styleDictionary.
- In many situations we were still generating styles with unique names when an existing style was identical.  Changed the code to re-use existing styles that had unique names. (This change was not necessary to fix the problems described in this defect but resulted in a substantial size reduction in the css file size. The Jii-entries project that is attached to the defect, and the Sena 3 project both result in a css file that is less than 10% of the old size.)

Change-Id: I457674ca44b37ed966cb1d8995011a0a19df4cca

### DIFF
--- a/Src/xWorks/CssGenerator.cs
+++ b/Src/xWorks/CssGenerator.cs
@@ -99,8 +99,7 @@ namespace SIL.FieldWorks.XWorks
 					return className;
 				}
 				// Otherwise get a unique but useful class name and re-generate the style with the new name
-				className = GetBestUniqueNameForNode(_styleDictionary, node);
-				_styleDictionary[className] = GenerateCssFromConfigurationNode(node, className, _propertyTable).NonEmpty();
+				className = GetBestUniqueNameForNode(node);
 				return className;
 			}
 		}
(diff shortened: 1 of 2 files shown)


JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/222#-:-ODMzjT966rmRvyM1zgM:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

jasonleenaylor:
*[`Src/xWorks/CssGenerator.cs` line 127 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/222#-ODN1QlJ0UBqC1D5kuqW:-ODN1QlK1yAvVdxo370E:bc1lpdz) ([raw file](https://github.com/sillsdev/FieldWorks/blob/cfdb6c0183bb35786a2a945d512cffbe638150fa/Src/xWorks/CssGenerator.cs#L127)):*
> ```Smalltalk
> 			string classNameBase = className;
> 			int counter = 0;
> 			while (_styleDictionary.ContainsKey(className))
> ```

I think we need to lock _styleDictionary here for thread safety.

PR #224, opened 2024-12-05
Fix LT-21999: Improve saving of parser test reports

jtmaxwell3:
This implements https://jira.sil.org/browse/LT-21999 with Jason's permission.
It adds a "Save Report" button to the window for individual parser reports.
It adds the comment as a label right after Save Report button.
It changes saving so that you can change a comment by saving the report again (deleting the old report).
It also fixes a bug that prevented comment fields from being updated properly when comments are changed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/224)

### DIFF
--- a/Src/LexText/ParserUI/ParserListener.cs
+++ b/Src/LexText/ParserUI/ParserListener.cs
@@ -619,8 +619,8 @@ namespace SIL.FieldWorks.LexText.Controls
 					ReadParserReports();
 					// Write an empty parser report.
 					var parserReport = CreateParserReport();
-					AddParserReport(parserReport);
-					ShowParserReport(parserReport, m_mediator, m_cache);
+					ParserReportViewModel viewModel = AddParserReport(parserReport);
+					ShowParserReport(viewModel, m_mediator, m_cache);
 				}
 			}
 			m_parserConnection.UpdateWordforms(wordforms, priority, checkParser);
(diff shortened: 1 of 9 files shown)


jasonleenaylor:
*[`Src/LexText/ParserUI/ParserReportDialog.xaml` line 19 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/224#-ODNoPSN1Fq3MFGdCxv4:-ODNoPSO4hrS3TlXA2qP:b7w03s2) ([raw file](https://github.com/sillsdev/FieldWorks/blob/8c8fef4b284801b2d702c1f3a4edf4f18f9748ae/Src/LexText/ParserUI/ParserReportDialog.xaml#L19)):*
> ```XML
> 		<Grid.RowDefinitions>
> 			<RowDefinition Height="*" />
> 			<RowDefinition Height="Auto" />
> ```

These are reversed from what I would expect, and it might explain why the Save button looks like it is cut off at the bottom.
Does it look better if they are swapped?

jasonleenaylor (approved):
> Reviewed 9 of 9 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/224)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #233, opened 2024-12-11
Word Export: LT-22006: RTL - Fix indentation and clipping

mark-sil:
Word does not expect the indentation direction (left/right) to be flipped for right to left output.  Removed the extra code and output the same indentation direction for both LTR and RTL.

This problem also existed with the single column output, but became more obvious when we switched to double column output.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/233)


JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/233#-:-ODqoiP1E8RMLQxAqRtI:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #234, opened 2024-12-11
Word Export: RTL – Set first column to be on the right

mark-sil:
For right to left documents this sets every section so that the column on the right is filled first.
With this change the user no longer needs to do the following steps in the Word document:
  In the Layout tab, click the icon at the lower right of the
  Page Setup section, and in the dialog that comes up, set
  Section direction: to Right-to-left. This is critical for RTL
  documents so that the first column on the page is on the right
  side.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/234)


JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/234#-:-ODr4UHS8d4Ro3-77LaX:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #239, opened 2024-12-13
Fix LT-22011: Adding slots lead to crash

jtmaxwell3:
This was another issue where a display object got disposed prematurely.  I came up with a more general solution to the problem by changing DataTree.PropChanged to use BeginInvoke to postpone the change until the invoker is done.  I also removed the fix to https://jira.sil.org/browse/LT-21980 since it is covered by the general solution.  The general solution should work, but I wonder if it is too general.  Jason, could this cause a problem?  Also, should we make this even more general by postponing all of DataTree.PropChanged?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/239)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/239)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #263, opened 2025-02-13
LT-22042: Fix incorrect styles being used in Flex

mark-sil:
- Keep the node path back to the root in the css file.  This is needed to ensure that styles are applied from the most specific (leaf node) to the lease specific (root node).
- All node names now have a unique identifier.  This is to prevent the wrong style from being used when nodes have the same display name (ie. Headword) but are located at different places in the tree. Without the unique identifier a node without a style could end up using a style from a different node with the same display name. With unique names, a node without a style will look to its parent node for style information, and then it’s grandparent node, … up to the root.
- The change in RemoveBeforeAndAfterForNoteInParaRules(): Changed the method so that we keep non-before/after rules. This fixes a problem with notes that should be in their own paragraph, not displaying correctly in Flex.

Additional cleanup still needed for the .css files (these issues Existed before this PR):
- We are currently creating multiple css entries with the same fields but different values.  In this case we always use the values from the last entry in the file.  We need to figure out why we are doing this and either support the intended behavior in a different way or remove the duplicates.
Example: ‘.subentries .subentry{}’ is written multiple times with the same 12 fields. The only field that ever changes value is ‘margin-left’.

- Sometimes we are removing the last letter from entry names in the css file for ‘between’ content. Need to investigate if this currently works to provide ‘between’ content.
Example: .headword> .headwor+ .headwor:before{ content:', ';}

Change-Id: I4bdf810bfd1be3a387f2d15dc656fa83615a5a5f

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/263)

### DIFF
--- a/Src/xWorks/CssGenerator.cs
+++ b/Src/xWorks/CssGenerator.cs
@@ -58,6 +58,7 @@ namespace SIL.FieldWorks.XWorks
 		private LcmCache _cache;
 		private ReadOnlyPropertyTable _propertyTable;
 		private Dictionary<string, List<StyleRule>> _styleDictionary = new Dictionary<string, List<StyleRule>>();
+		private Dictionary<string, string> _uniqueNodeNames = new Dictionary<string, string>();
 		private StyleSheet _styleSheet = new StyleSheet();
 
 		public void Init(ReadOnlyPropertyTable propertyTable)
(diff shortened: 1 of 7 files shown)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/263)* status: 0 of 7 files reviewed, 1 unresolved discussion

___
*[`Src/xWorks/CssGenerator.cs` line 100 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/263#-OJ4HE84AdyMGTiTh-uu:-OJ4HE84AdyMGTiTh-uv:bx7x5g5) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2c56cf1a074d4b9a8e92964a860663933ceaf1b3/Src/xWorks/CssGenerator.cs#L100)):*
> ```Smalltalk
> 				string uniqueNodeName = null;
> 				string uniqueNodePath = null;
> 				for (int ii=0; ii < nodes.Count; ii++)
> ```

Is there a reason this variable is named ii as opposed to i or index?

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/263)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @JakeOliver28)

___
*[`Src/xWorks/CssGenerator.cs` line 100 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/263#-OJ4HE84AdyMGTiTh-uu:-OJ4Sa4BDArS0ZndAYz4:b-cmiagf) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2c56cf1a074d4b9a8e92964a860663933ceaf1b3/Src/xWorks/CssGenerator.cs#L100)):*

Is there a reason this variable is named ii as opposed to i or index?
</blockquote></details>

   Just a habit that makes it easy to search for.

aror92 (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/263)* status: all files reviewed, 2 unresolved discussions (waiting on @JakeOliver28)

___
*[`Src/xWorks/CssGenerator.cs` line 114 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/263#-OJJ5G5h2HSbMuwikCNG:-OJJ5G5h2HSbMuwikCNH:b-nrruu) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2c56cf1a074d4b9a8e92964a860663933ceaf1b3/Src/xWorks/CssGenerator.cs#L114)):*
> ```Smalltalk
> 						// because of the way the path to ".mainentrysubentries" is generated we can
> 						// get the wrong value for ".sense-?. So add ".sensecontent .sense" to the path.
> 						uniqueNodePath = uniqueNodePath + ".sensecontent .sense ";
> ```

It could be helpful to add a comment here that the space at the end of uniqueNodePath is intentional. It wasn't clear to me that it was intentional until I reviewed AddUniquePathToStyleRules (especially since there isn't a space at the end of ".sensecontent .sense" in the comment).

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/263#-:-OJK6GUC-sFbLssGmdqy:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @mark-sil)

PR #265, opened 2025-02-19
LT-22042: Fix incorrect styles being used for subentries

mark-sil:
The style information used for the nodes under
MainEntry->Subentries, is also used for
MainEntry->Senses->Subentries.
Fix the problem so that we now use the style information for both node paths.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/265)

### DIFF
--- a/Src/xWorks/CssGenerator.cs
+++ b/Src/xWorks/CssGenerator.cs
@@ -97,37 +97,65 @@ namespace SIL.FieldWorks.XWorks
 				// Generate the unique name and css for each node (starting from the root node).
 				string uniqueNodeName = null;
 				string uniqueNodePath = null;
+				string sensesSubentryNodePath = null;
+				bool buildSenseSubentryRules = false;
 				for (int ii=0; ii < nodes.Count; ii++)
 				{
 					var workingNode = nodes[ii];
-					string nodePath = GetNodePath(workingNode);
 					var workingClassName = $".{GetClassAttributeForConfig(workingNode)}";
 
 					// If this node is ".subentries" and the next node is ".mainentrysubentries",
-					// then set the uniqueNodeName to ".mainentrysubentry" and skip the next node.
+					// then we need to build two sets of rules, one for ".entry-1 .subentries-?"
+					// and one for ".entry-1 .senses-? .subentries-?". We only need to build the two
+					// set for the children. The node ".entry-1 .senses-? .subentries-?" does NOT
+					// need a second set of rules.  It's rules will get created through the normal
+					// execution of this method (and could use a different style).
 					if (workingClassName == ".subentries" && (ii + 1 < nodes.Count) &&
 						$".{GetClassAttributeForConfig(nodes[ii + 1])}" == ".mainentrysubentries")
 					{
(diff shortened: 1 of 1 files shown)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/265)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`Src/xWorks/CssGenerator.cs` line 122 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/265#-OJsfXbs71oo2KV8rpMd:-OJxALtx2Gvt0wB782_n:b-7kvlpj) ([raw file](https://github.com/sillsdev/FieldWorks/blob/9f85e9646e28ffe432ae1d81df287ed8f1c0afdc/Src/xWorks/CssGenerator.cs#L122)):*

I think I'm missing something about the logic of the entry > subentries VS. entry > senses > subentries cases. Is "buildSenseSubentryRules" meant to be keeping track of whether we're dealing with the entry > senses > subentries case? Why would this check ensure that we are looking at the entry > senses > subentries case--wouldn't we enter this if statement in the entry > subentries case as well?
</blockquote></details>

There are no child nodes for ‘.entry .senses .subentries’. The existing code is designed to use the ‘.entry .subentries’ child nodes to determine how to display both the children of ‘.entry .subentries’ and the children of ‘.entry .senses .subentries’. After buildSenseSubentryRules is set to true, we are processing the child nodes of ‘.entry .subentries’, so we need to start building two sets of rules. One set to determine the display for ‘.entry .subentries’ children and another set to determine the display for ‘.entry .senses .subentries’ children.

aror92 (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/265)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

___
*[`Src/xWorks/CssGenerator.cs` line 122 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/265#-OJsfXbs71oo2KV8rpMd:-OJxJ9IY9U1pthDwFHRO:b-f743p8) ([raw file](https://github.com/sillsdev/FieldWorks/blob/9f85e9646e28ffe432ae1d81df287ed8f1c0afdc/Src/xWorks/CssGenerator.cs#L122)):*

There are no child nodes for ‘.entry .senses .subentries’. The existing code is designed to use the ‘.entry .subentries’ child nodes to determine how to display both the children of ‘.entry .subentries’ and the children of ‘.entry .senses .subentries’. After buildSenseSubentryRules is set to true, we are processing the child nodes of ‘.entry .subentries’, so we need to start building two sets of rules. One set to determine the display for ‘.entry .subentries’ children and another set to determine the display for ‘.entry .senses .subentries’ children.
</blockquote></details>

Ah, that makes sense. Thank you!

aror92 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/265)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #268, opened 2025-02-21
Fix LT-21953: Sort Reversal Subentries crash when subentry missing form

jtmaxwell3:
I replaced null with "" to avoid the crash.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/268)

### DIFF
--- a/Src/ManagedLgIcuCollator/LgIcuCollator.cs
+++ b/Src/ManagedLgIcuCollator/LgIcuCollator.cs
@@ -94,6 +94,10 @@ namespace SIL.FieldWorks.Language
 		public int Compare(string bstrValue1, string bstrValue2, LgCollatingOptions colopt)
 		{
 			EnsureCollator();
+			if (bstrValue1 == null)
+				bstrValue1 = "";
+			if (bstrValue2 == null)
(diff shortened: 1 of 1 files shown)


jasonleenaylor (commented):
Well, that will certainly avoid the crash. This code clearly expected never to receive null values, alternatively we could decide that nulls sort to the top and handle all that compare logic here without involving m_collator at all.
if(bstrValue1 == null || bstrValue2 == null)
   return bstrValue1 == null ? (bstrValue2 == null ? 0 : -1) : 1;

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/268#-:-OJdP7qW7Y5qxS5FlKEl:b-8wpors)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

jasonleenaylor:
> I don't know what the LgCollatingOptions parameter colopt is supposed to do, so I would rather keep the current code because it calls CompareVariant with colopt. Bypassing CompareVariant might cause subtle issues at some point in the future.

Hmm, that is theoretically possible. :LGTM:

PR #269, opened 2025-02-21
Fix LT-22047 without breaking LT-21810

jtmaxwell3:
The original code broke because FindOrCreateWordform was not in the scope of a unit of work helper.  Not knowing better, I fixed in LT-21810 by avoided calling FindOrCreateWordform.  But this causes problems if the corpus has an uppercase version of a lowercase word that doesn't appear in the corpus.  So I added a unit of work helper around FindOrCreateWordform that solves both LT-22047 and LT-21810.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/269)

### DIFF
--- a/Src/LexText/ParserCore/ParserWorker.cs
+++ b/Src/LexText/ParserCore/ParserWorker.cs
@@ -158,26 +158,24 @@ namespace SIL.FieldWorks.WordWorks.Parser
 			if (sLower != form.Text)
 			{
 				var text = TsStringUtils.MakeString(sLower, form.get_WritingSystem(0));
-				IWfiWordform lcWordform;
-				// We cannot use WfiWordformServices.FindOrCreateWordform because of props change (LT-21810).
-				// Only parse the lowercase version if it exists.
-				if (m_cache.ServiceLocator.GetInstance<IWfiWordformRepository>().TryGetObject(text, out lcWordform))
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
*[`Src/LexText/ParserCore/ParserWorker.cs` line 173 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/269#-OJdxyqKDmCWewKBsQl9:-OJdxyqKDmCWewKBsQlA:b9zrryn) ([raw file](https://github.com/sillsdev/FieldWorks/blob/d37f81c61fa0595d8e804a1c71a9eaeb7252df50/Src/LexText/ParserCore/ParserWorker.cs#L173)):*
> ```Smalltalk
> 				{
> 					IWfiWordform lcWordform = null;
> 					NonUndoableUnitOfWorkHelper.Do(
> ```

There is a safer version DoUsingNewOrExisting (or something like that) which I would recommend.



_Code quote:_
```Smalltalk
Do(
```

jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/269#-:-OK2BJFI04aT8THl0HcR:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #270, opened 2025-02-21
LT-21757: Allow moving templates and slots up or down in Category hierarchy

jtmaxwell3:
I added code to move affix templates and affix slots in the right-click menu of Template Name and Slot Name. Moving a template up will also move local slots up if it doesn't cause a conflict. Moving a template anywhere else will leave the slot in place, but will highlight the slot in the template if it is out of scope.

This is for release/9.3.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/270)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/270)* status: 0 of 9 files reviewed, 1 unresolved discussion

___
*[`Src/Common/Controls/XMLViews/XmlVc.cs` line 3892 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/270#-OJsxTrR1VbiYMzYKfP8:-OJsxTrR1VbiYMzYKfP9:b-9xlqnv) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b9bf95be3c9973ab5be116b9fb9332b204725ec/Src/Common/Controls/XMLViews/XmlVc.cs#L3892)):*
> ```Smalltalk
> 			while (partOfSpeech != null)
> 			{
> 				foreach (IMoInflAffixSlot slot in partOfSpeech.AllAffixSlots)
> ```

We can eliminate this foreach loop with a FirstOrDefault. I think this would be simpler, quicker, and more readable.



_Code snippet:_
```Smalltalk
while (partOfSpeech != null)
{
    var matchingSlot = partOfSpeech.AllAffixSlots
        .FirstOrDefault(slot => slot.Name.BestAnalysisVernacularAlternative.Text == name);

    if (matchingSlot != null)
        return matchingSlot;

    partOfSpeech = partOfSpeech.Owner as IPartOfSpeech;
}
```

JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/270)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @jtmaxwell3)

___
*[`Src/xWorks/DTMenuHandler.cs` line 624 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/270#-OJt-n86D1ifX3NAMo9I:-OJt-n86D1ifX3NAMo9J:bren0qi) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b9bf95be3c9973ab5be116b9fb9332b204725ec/Src/xWorks/DTMenuHandler.cs#L624)):*
> ```Smalltalk
> 			while (partOfSpeech != null)
> 			{
> 				foreach (IMoInflAffixSlot slot in partOfSpeech.AllAffixSlots)
> ```

And here. Should we consolidate these GetPOSSlot methods into one since they're identical? Guessing that might not be easily doable.
___
*[`Src/LexText/ParserCore/HCLoader.cs` line 1503 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/270#-OJt-irA8Wxk1ami350h:-OJt-irA8Wxk1ami350i:b-sdk4jt) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b9bf95be3c9973ab5be116b9fb9332b204725ec/Src/LexText/ParserCore/HCLoader.cs#L1503)):*
> ```Smalltalk
> 			while (partOfSpeech != null)
> 			{
> 				foreach (IMoInflAffixSlot slot in partOfSpeech.AllAffixSlots)
> ```

Same thing here.

jtmaxwell3:
I find Linq-style code less readable than for loops so I only use it if the for loops are hard to read.  In these cases, it is easy to see what the for loop is doing, so I don't think that this is worth changing.

XmlVc, DTMenuHandler, and HCLoader don't have an import in common within FieldWorks and aren't accessible to each other, so it isn't obvious where to put GetPOSSlot so that it can be shared.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/270#-:-OJyT9Q80tt_KE0rWSK_:bnfp4nl)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @jtmaxwell3)

PR #282, opened 2025-03-07
Fix LT-21826: Usage figures for Publications show inverted values

jtmaxwell3:
Publications are included unless that are stored in DoNotPublishIn.  So, the ReferringObjects are the excluded lexical entries, not the included lexical entries.  I inverted the count in LabelNode.CountUsages.  But there has to be a better way to determine that we are dealing with publications than what I did.

This could be in release/9.2, since it is in the backlog.

### DIFF
--- a/Src/Common/Controls/XMLViews/LabelNode.cs
+++ b/Src/Common/Controls/XMLViews/LabelNode.cs
@@ -105,6 +105,14 @@ namespace SIL.FieldWorks.Common.Controls
 			// I think only label.Object is likely to be null, but let's prevent crashes thoroughly.
 			if (label != null && label.Object != null && label.Object.ReferringObjects != null)
 			{
+				if (label.Object.Owner.ToString() == "Publications")
+				{
+					// IPublication is the value of DoNotPublishIn,
+					// so we need to invert the count.
+					ILexEntryRepository repository = label.Object.Cache.ServiceLocator.GetInstance<ILexEntryRepository>();
+					count = repository.Count - label.Object.ReferringObjects.Count;
+					return count;
+				}
 				count = label.Object.ReferringObjects.Count;
 				foreach (ICmObject x in label.Object.ReferringObjects)
 				{


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/282)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`Src/Common/Controls/XMLViews/LabelNode.cs` line 106 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/282#-OKmUb6D7w4YKPHtxmFb:-OKmUb6D7w4YKPHtxmFc:b-t19fse) ([raw file](https://github.com/sillsdev/FieldWorks/blob/0cb043e225ba6625b661e0fa5a0ed7adefa8a0e8/Src/Common/Controls/XMLViews/LabelNode.cs#L106)):*
> ```Smalltalk
> 			int count = 0;
> 			// I think only label.Object is likely to be null, but let's prevent crashes thoroughly.
> 			if (label != null && label.Object != null && label.Object.ReferringObjects != null)
> ```

Consolidating the null check and removing the loop can simplify things



_Code snippet:_
```Smalltalk
var obj = label?.Object;

if (obj?.referringObjects == null)
        return 0;
    
int count; 
        
if (obj.Owner.ToString() == "Publications")
{
	var totalEntries = obj.Cache.ServiceLocator.GetInstance<ILexEntryRepository>().Count;
	count = totalEntries - obj.ReferringObjects.Count;
}
else
{
	count = obj.ReferringObjects.Count - obj.ReferringObjects.OfType<ICmOverlay>().Count();     
}
	
return count;
```

jasonleenaylor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/282)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JakeOliver28 and @jtmaxwell3)

___
*[`Src/Common/Controls/XMLViews/LabelNode.cs` line 108 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/282#-OKmdb_aACzn3SRQjQOY:-OKmdb_aACzn3SRQjQOZ:b-8ubbqr) ([raw file](https://github.com/sillsdev/FieldWorks/blob/0cb043e225ba6625b661e0fa5a0ed7adefa8a0e8/Src/Common/Controls/XMLViews/LabelNode.cs#L108)):*
> ```Smalltalk
> 			if (label != null && label.Object != null && label.Object.ReferringObjects != null)
> 			{
> 				if (label.Object.Owner.ToString() == "Publications")
> ```

I don't particularly like having this level of specificity in what is a very generic class.
I'm wrestling with what the fix looks like. It does involve more code and a much larger change, but I also know that the classes where we have let this kind of cheat sneak in have become nearly unmaintainable.

Here is the alternative:
Add a new ObjectLabel type PublicationLabel
Modify CreateObjectLabel in ObjectLabel to test if obj is IPublication and create the correct type.
Make a public virtual field on ObjectLabel that returns Count.
Move the existing Count code to that.
Override Count in the IPublication to do the inverted count.

I'm going to sleep on this.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/282#-:-OL-ZhNnCZRytn39QYc5:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor and @jtmaxwell3)

jtmaxwell3:
Jason, your proposal can be simplified to:
Add a new LabelNode type PublicationLabelNode
Modify CreateLabelNode in ReallySimpleListChooser to test if obj is a CmPossibility with owner "Publications" and create the correct type.
Add PublicationLabelNode.CountUsages.

jasonleenaylor (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/282#-:-OL5D3k15w1Z0RKtIasl:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

PR #283, opened 2025-03-11
LT-22039: xhtml has new nodeId attribute

jtmaxwell3:
A nodeId attribute was recently added to the XHTML representation to aid the right-click Dictionary Configuration option in finding the correct data after recent changes that reduced the number of styles in a project.  The nodeId attribute is also included when exporting a project to XHTML, which is undesirable according to Ken Zook.  I fixed this by added an IsXhtmlExport property to GeneratorSettings and passing GeneratorSettings to all of the methods in LcmXhtmlGenerator that wrote a nodeId attribute.  I also added a WriteNodeId method which contains the logic for deciding whether a nodeId attribute should be written out.

I believe that the IsXhtmlExport property is only set when exporting XHTML, but I don't know how to test whether the Dictionary Configuration option is still working correctly.  I also don't know how to verify that exporting to Word and Webonary is still working correctly.

This should be considered for release/9.2 since it is a regression caused by changes in FW 9.2.5.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/283)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -344,7 +344,7 @@ namespace SIL.FieldWorks.XWorks
 				{
 					var clerk = settings.PropertyTable.GetValue<RecordClerk>("ActiveClerk", null);
 					var entryClassName = settings.StylesGenerator.AddStyles(configuration).Trim('.');
-					settings.ContentGenerator.StartEntry(xw, configuration,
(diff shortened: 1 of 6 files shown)


mark-sil (approved):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/283)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #286, opened 2025-03-12
Add css support for document wide image settings

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/286)

### DIFF
--- a/DistFiles/Language Explorer/Configuration/DictionaryConfiguration.xsd
+++ b/DistFiles/Language Explorer/Configuration/DictionaryConfiguration.xsd
@@ -42,6 +42,7 @@
 <xs:complexType name="PicturesType">
 	<xs:attribute type="alignmentType" name="alignment"/>
 	<xs:attribute type="xs:float" name="width"/>
+	<xs:attribute type="xs:float" name="height"/>
 </xs:complexType>
 <xs:complexType name="SenseOptionsType">
 	<xs:attribute type="xs:string" name="numberStyle"/>
--- a/Src/xWorks/ConfigurableDictionaryNode.cs
+++ b/Src/xWorks/ConfigurableDictionaryNode.cs
@@ -30,6 +30,13 @@ namespace SIL.FieldWorks.XWorks
 		public StringTable StringTable { get; set; }
 
 		/// <summary>
+		/// This is the model that the configuration node is associated with, used for some
+		/// more global settings.
+		/// </summary>
+		[XmlIgnore]
+		public DictionaryConfigurationModel Model { get; set; }
+
+		/// <summary>
 		/// The non-editable portion of the label to display for this node
 		/// </summary>
 		[XmlAttribute(AttributeName = "name")]
--- a/Src/xWorks/CssGenerator.cs
+++ b/Src/xWorks/CssGenerator.cs
@@ -846,13 +846,6 @@ namespace SIL.FieldWorks.XWorks
 					Term = new PrimitiveTerm(UnitType.Inch, pictureOptions.MinimumHeight)
 				});
 			}
-			if(pictureOptions.MaximumHeight > 0)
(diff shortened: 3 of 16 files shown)


aror92 (changes requested):
> Reviewed 2 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/286)* status: 2 of 16 files reviewed, 1 unresolved discussion

___
*[`Src/xWorks/CssGenerator.cs` line 808 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/286#-OLKC9au8Zehq0kg1d2R:-OLKC9av609njURtTEnn:bjmqwi7) ([raw file](https://github.com/sillsdev/FieldWorks/blob/761a5643a8cc97caed3acbe790e7233d399f8c19/Src/xWorks/CssGenerator.cs#L808)):*
> ```Smalltalk
> 			{
> 				case "left":
> 					marginPropName = "margin-right";
> ```

Is it intentional that the margin is right when the alignment is left (and below that margin is left in the default case of right alignment)? If so, I think a comment would be helpful.

JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/286)* status: 2 of 16 files reviewed, 2 unresolved discussions (waiting on @aror92 and @jasonleenaylor)

___
*[`Src/xWorks/CssGenerator.cs` line 770 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/286#-OLKDXkJ71At1uSSn3qK:-OLKDXkJ71At1uSSn3qL:bdclw0x) ([raw file](https://github.com/sillsdev/FieldWorks/blob/761a5643a8cc97caed3acbe790e7233d399f8c19/Src/xWorks/CssGenerator.cs#L770)):*
> ```Smalltalk
> 			captionRule.Value = baseSelection + " .captionContent";
> 			// set the display as table-caption to allow the caption to be centered under the image
> 			captionRule.Declarations.Properties.Add(new Property("display")
> ```

Would it be beneficial to create a helper method for adding properties like this? We could consolidate logic and create an overloaded method to where we only need to put something like this:



_Code snippet:_
```Smalltalk
AddProperty(captionRule, "display", "table-caption");
// or
AddProperty(captionRule, "text-indent", UnitType.Point, 0);
```

jasonleenaylor:
*[`Src/xWorks/CssGenerator.cs` line 770 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/286#-OLKDXkJ71At1uSSn3qK:-OLKOacDB8XDBx6f4h9C:b1qitv0) ([raw file](https://github.com/sillsdev/FieldWorks/blob/761a5643a8cc97caed3acbe790e7233d399f8c19/Src/xWorks/CssGenerator.cs#L770)):*

Would it be beneficial to create a helper method for adding properties like this? We could consolidate logic and create an overloaded method to where we only need to put something like this:
</blockquote></details>

We could. I think I would want to save that for a future refactor though it would make this change larger and harder to understand in history.

jasonleenaylor:
*[`Src/xWorks/CssGenerator.cs` line 808 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/286#-OLKC9au8Zehq0kg1d2R:-OLKP4yV75QtOLn6bAA_:bxdxilr) ([raw file](https://github.com/sillsdev/FieldWorks/blob/761a5643a8cc97caed3acbe790e7233d399f8c19/Src/xWorks/CssGenerator.cs#L808)):*

Is it intentional that the margin is right when the alignment is left (and below that margin is left in the default case of right alignment)? If so, I think a comment would be helpful.
</blockquote></details>

Yes, I'll add comments.

PR #294, opened 2025-03-20
Fix LT-22083: Crash occurs while importing interlinear into empty pane

jtmaxwell3:
This crashed because RootStText.Cache was null.  I assume that this is some sort of dummy RootStText.  I fixed this by avoiding processing it.  I'm not sure how RootStText.Cache became null.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/294)

### DIFF
--- a/Src/LexText/Interlinear/InterlinDocRootSiteBase.cs
+++ b/Src/LexText/Interlinear/InterlinDocRootSiteBase.cs
@@ -1030,7 +1030,7 @@ namespace SIL.FieldWorks.IText
 		{
 			//If the RootStText is null we are either in a place that doesn't care about parser related updates
 			// or we are not yet completely displaying the text, so we should be fine, I hope? (LT-12493)
-			if (SuspendResettingAnalysisCache || RootStText == null)
+			if (SuspendResettingAnalysisCache || RootStText == null || RootStText.Cache == null)
 				return;
 
 			switch (tag)


JakeOliver28 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/294)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`Src/LexText/Interlinear/InterlinDocRootSiteBase.cs` line 1033 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/294#-OLoeYpeAjvCE8585E34:-OLoeYpeAjvCE8585E35:b-mmv5kg) ([raw file](https://github.com/sillsdev/FieldWorks/blob/8312c2d83d084cd66af82d868d57a4e04631a711/Src/LexText/Interlinear/InterlinDocRootSiteBase.cs#L1033)):*
> ```Smalltalk
> 			//If the RootStText is null we are either in a place that doesn't care about parser related updates
> 			// or we are not yet completely displaying the text, so we should be fine, I hope? (LT-12493)
> 			if (SuspendResettingAnalysisCache || RootStText == null || RootStText.Cache == null)
> ```

Could we do this?



_Code snippet:_
```Smalltalk
if (SuspendResettingAnalysisCache || RootStText?.Cache == null)
```

jtmaxwell3:
I think that RootStText?.Cache == null makes the logic less clear, so I would prefer to leave the code as it is.

JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/294#-:-OLozxtE67kmNh0w2t_K:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jtmaxwell3)

PR #301, opened 2025-03-27
Fix LT-22089: Creating and undoing a text leads to confusion

jtmaxwell3:
The problem with https://jira.sil.org/browse/LT-22089 is that RecordList gets a PropChanged notification before InterestingTextList does, so RecordList.GetObjectSet returns the old value for InterestingTextList.   At Jason's suggestion, I fixed it by setting m_requestedLoadWhileSuppressed = true so that RecordList would get updated again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/301)


JakeOliver28 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/301#-:-OMOCiekA7dZcYEgPmZM:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #302, opened 2025-03-28
Word Export: LT-22051: Fix problems with incorrect styles

mark-sil:
Change the way that we generate character styles so that it includes styles from all the parent nodes. Also changed the way paragraph styles are generates so they no longer directly include run properties.  Now they are linked to a character style that will be used to build the character style properties.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 1077 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMgWKyD94t8fWTUc7Iz:-OMlKz3fApnP5fzt_KZE:b-k0bg0u) ([raw file](https://github.com/sillsdev/FieldWorks/blob/83a0aee2a7819879ff113df338586ecf5c8a2f8a/Src/xWorks/LcmWordGenerator.cs#L1077)):*

Is there potential for cases when we have more than one run but need a writing system for Before/After other than the default?
</blockquote></details>

Yes, there is potential for that, but I haven’t seen it yet.  If there is more than one run at this point then they could have different writing systems, and to match what Flex displays, we would want to use the default ws. If we find a case where there are multiple runs that all use the same ws, and Flex uses that ws instead of default, then we would need to figure out what makes that case unique and add special case code to mimic what Flex displays.

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @aror92)

___
*[`Src/xWorks/WordStyleCollection.cs` line 94 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMlgjdO-NSc0HWYOy87:-OMmEBxi1C-lWYPO9zUZ:b-d8en19) ([raw file](https://github.com/sillsdev/FieldWorks/blob/83a0aee2a7819879ff113df338586ecf5c8a2f8a/Src/xWorks/WordStyleCollection.cs#L94)):*

Does characterStyles need to be locked as well?
</blockquote></details>

Good catch. I need to use the same lock object consistently, in all places. I’ll push another change.
___
*[`Src/xWorks/WordStyleCollection.cs` line 406 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMlkHzU5_UPdhAAkEJG:-OMmE3GX13_uK7DCUg2F:b-i2frwu) ([raw file](https://github.com/sillsdev/FieldWorks/blob/83a0aee2a7819879ff113df338586ecf5c8a2f8a/Src/xWorks/WordStyleCollection.cs#L406)):*

Just to make sure I'm understanding correctly:

Suppose two styles have the same displayNameBase and different nodePaths, but all the style elements are identical between the two, then the redirect ensures we use a single style for both, is that right?

And the redirection will only be used for styles with the same displayNameBase? (This makes sense--for identical styles with the same displayNameBase, it seems likely the user would want/expect the two styles to behave as a single style in Word export, e.g. if the style is edited from inside Word.)
</blockquote></details>

Yes, your statement is correct.

aror92 (commented):
> Reviewed 1 of 5 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: 3 of 5 files reviewed, 3 unresolved discussions

___
*[`Src/xWorks/WordStyleCollection.cs` line 130 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMmywyxClVTkzwlirAY:-OMmywyxClVTkzwlirAZ:bmdte09) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b35c763c8fee423178203fa382cbcec86b5c35d/Src/xWorks/WordStyleCollection.cs#L130)):*
> ```Smalltalk
> 		/// </summary>
> 		/// <returns>The uniqueDisplayName for the character style that is created.</returns>
> 		private string AddStyle(string styleName, string displayNameBase, string nodePath,
> ```

Given that AddStyle starts by creating a character style, and then maybe creates a paragraph style as well, I'm wondering: what happens if a style in FLEx specifies some paragraph style elements but does not specify any character style elements, is it possible to make a paragraph style without also making a character style? Would this just result in an empty character style which would be discarded, or in reusing a default normal character style?
___
*[`Src/xWorks/WordStyleCollection.cs` line 157 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMmxh99BnO_s37ljJRl:-OMmxh99BnO_s37ljJRm:bwsv445) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b35c763c8fee423178203fa382cbcec86b5c35d/Src/xWorks/WordStyleCollection.cs#L157)):*
> ```Smalltalk
> 
> 				bool processParagraphStyle = ((nodePath != WordStylesGenerator.RootCharacterNodePath) &&
> 											  (wsId == WordStylesGenerator.DefaultStyle) &&
> ```

Why is a paragraph style processed only if the wsId is the default? Is this just to make sure a paragraph style is only made one time? (since in line 241 AddStyle is called again with the default ws if the current ws is not default)
___
*[`Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs` line 530 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMn-LqxEGH7uNrb-WtY:-OMn-LqxEGH7uNrb-WtZ:b-tgeduv) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b35c763c8fee423178203fa382cbcec86b5c35d/Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs#L530)):*
> ```Smalltalk
> 			Assert.True(outXml.Contains(beforeAbbreviation));
> 
> 			// Between text 'TW3' is before the spanish abbreviation, which is before 'glossFR'.
> ```

is it french abbreviation now?

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on @aror92)

___
*[`Src/xWorks/WordStyleCollection.cs` line 130 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMmywyxClVTkzwlirAY:-OMn5ycF21p-d_1wQqHH:bz60y53) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b35c763c8fee423178203fa382cbcec86b5c35d/Src/xWorks/WordStyleCollection.cs#L130)):*

Given that AddStyle starts by creating a character style, and then maybe creates a paragraph style as well, I'm wondering: what happens if a style in FLEx specifies some paragraph style elements but does not specify any character style elements, is it possible to make a paragraph style without also making a character style? Would this just result in an empty character style which would be discarded, or in reusing a default normal character style?
</blockquote></details>

The character style would contain RunProperties that are empty, and since it is linked to a paragraph style it would not get discarded.  The paragraph style always expects the linked character style to exist, so this is an optimization we don't currently do.  
___
*[`Src/xWorks/WordStyleCollection.cs` line 157 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMmxh99BnO_s37ljJRl:-OMn7PJYA8vDzmiAQvIO:b-lpylqi) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b35c763c8fee423178203fa382cbcec86b5c35d/Src/xWorks/WordStyleCollection.cs#L157)):*

Why is a paragraph style processed only if the wsId is the default? Is this just to make sure a paragraph style is only made one time? (since in line 241 AddStyle is called again with the default ws if the current ws is not default)
</blockquote></details>

Yes, correct. The paragraph style will get created when the default character style is created. Paragraph styles are always default ws.
___
*[`Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs` line 530 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMn-LqxEGH7uNrb-WtY:-OMn8c1Y234C1jito30O:b-vky07p) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2b35c763c8fee423178203fa382cbcec86b5c35d/Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs#L530)):*

is it french abbreviation now?
</blockquote></details>

Yes, I pushed a new file with this change.

aror92 (changes requested):
> Reviewed 1 of 5 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: all files reviewed, 1 unresolved discussion (waiting on @mark-sil)

___
*[`Src/xWorks/WordStyleCollection.cs` line 415 at r3](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMqt63mB7bIq3udA6GO:-OMqt63mB7bIq3udA6GP:b-alk9f3) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2d5193f0d4015f6832c4ae4e9da1993052192e12/Src/xWorks/WordStyleCollection.cs#L415)):*
> ```Smalltalk
> 				{
> 					// Iterate through the proceeding elements to check if they have the same properties.
> 					for (int proceedingElem = 0; proceedingElem < currentElem; proceedingElem++)
> ```

This isn't a big deal, but for code readability in the future, it may be worth updating the variable & the comments: if I'm understanding this correctly, I think the intent here is preceedingElem (proceeding is for items that come after and preceeding for items that come before)

aror92 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: all files reviewed, 1 unresolved discussion (waiting on @mark-sil)

___
*[`Src/xWorks/WordStyleCollection.cs` line 415 at r3](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMqt63mB7bIq3udA6GO:-OMr77sTCuYStMi1wpOD:b-dz6dqu) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2d5193f0d4015f6832c4ae4e9da1993052192e12/Src/xWorks/WordStyleCollection.cs#L415)):*

I want to change the comment to make it clearer, but I don't know where you are seeing the inconsistency. Would it help if I added a comment that said, "A proceedingElem is any element that comes before currentElem."?
</blockquote></details>

I was trying to suggest replacing "proceeding" with "preceding" throughout the code & comments

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`Src/xWorks/WordStyleCollection.cs` line 415 at r3](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-OMqt63mB7bIq3udA6GO:-OMrAZWZ25bHQn9GlAZ_:b-kg3sdc) ([raw file](https://github.com/sillsdev/FieldWorks/blob/2d5193f0d4015f6832c4ae4e9da1993052192e12/Src/xWorks/WordStyleCollection.cs#L415)):*

I was trying to suggest replacing "proceeding" with "preceding" throughout the code & comments
</blockquote></details>

Thanks for clarifying.  I pushed a new file to clean it up.

aror92 (approved):
:lgtm:

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302#-:-OMrHr2W8w8NeX8tz5GZ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

aror92 (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/302)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #310, opened 2025-04-03
Word Export: LT-22051: Removed unused paragraph styles

mark-sil:
- Removed unused paragraph styles.
- If there are multiple paragraph styles with the same DisplayNameBase, and they have identical properties, then redirect the references to use the same style and remove the duplicate.

Change-Id: Ie00017b6127b02ba2350076f2f2df41897178e7e

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/310)

### DIFF
--- a/Src/xWorks/LcmWordGenerator.cs
+++ b/Src/xWorks/LcmWordGenerator.cs
@@ -116,11 +116,13 @@ namespace SIL.FieldWorks.XWorks
 				var propStyleSheet = FontHeightAdjuster.StyleSheetFromPropertyTable(propertyTable);
 
 				s_styleCollection.RedirectCharacterElements();
+				s_styleCollection.RedirectParagraphElements();
 
 				foreach (var entry in entryContents)
 				{
 					if (!entry.Item2.IsNullOrEmpty())
 					{
+						// Change run references to redirected styles.
 						foreach (WP.Run run in ((DocFragment)entry.Item2).DocBody.Descendants<WP.Run>())
 						{
 							// Reset the run styles to any redirected styles and mark styles as being used.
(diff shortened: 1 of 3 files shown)


aror92 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/310)* status: all files reviewed, 2 unresolved discussions (waiting on @mark-sil)

___
*[`Src/xWorks/WordStyleCollection.cs` line 542 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/310#-ONKmqwREJ7ZR1C_Y0p4:-ONKmqwREJ7ZR1C_Y0p5:b42551z) ([raw file](https://github.com/sillsdev/FieldWorks/blob/66e36610c5985b5e68e55d7e36024a9ed5dd3ac8/Src/xWorks/WordStyleCollection.cs#L542)):*
> ```Smalltalk
> 
> 				// Set the paragraph and character styles to used.
> 				charElem.Used = true;
> ```

Why aren't these needed? I'm guessing they're being set somewhere else?
___
*[`Src/xWorks/LcmWordGenerator.cs` line 160 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/310#-ONKmGRlBMlDLfwCJQFf:-ONKkshNCYk9CCem3ARG:b-5n5s6j) ([raw file](https://github.com/sillsdev/FieldWorks/blob/cdde6dda89a80db5a76dd5ca020343fc9dcd804a/Src/xWorks/LcmWordGenerator.cs#L160)):*
> ```Smalltalk
> 									if (paraElem.LinkedCharacterElement != null)
> 									{
> 										paraElem.LinkedCharacterElement.Redirect.Used = true;
> ```

Is there a need to change the unique display name for the redirected character element? (Like line 134 above?)

mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/310)* status: all files reviewed, 2 unresolved discussions (waiting on @aror92)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 160 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/310#-ONKmGRlBMlDLfwCJQFf:-ONL8dlD8o4tGddH7zzK:b-6gnfe7) ([raw file](https://github.com/sillsdev/FieldWorks/blob/cdde6dda89a80db5a76dd5ca020343fc9dcd804a/Src/xWorks/LcmWordGenerator.cs#L160)):*

Is there a need to change the unique display name for the redirected character element? (Like line 134 above?)
</blockquote></details>

Good question.  Line 134 actually does the redirection for the linked character element. RedirectParagraphElements() sets the redirect on the LinkedCharacterElement so that it gets processed while we are going through the runs (just like other characterElements).
___
*[`Src/xWorks/WordStyleCollection.cs` line 542 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/310#-ONKmqwREJ7ZR1C_Y0p4:-ONL6rfoBiXggf0R6VE9:bp4tlbj) ([raw file](https://github.com/sillsdev/FieldWorks/blob/66e36610c5985b5e68e55d7e36024a9ed5dd3ac8/Src/xWorks/WordStyleCollection.cs#L542)):*

Why aren't these needed? I'm guessing they're being set somewhere else?
</blockquote></details>

Correct.  This code was forcing all of them to be considered 'used'.  Now they are only set to 'used' when they are identified as being used. That code is in LcmWordGenerator.cs around lines 157 and 165.

aror92 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/310)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #311, opened 2025-04-03
Fix LT-22095: Affix Template not duplicating correctly

jtmaxwell3:
https://jira.sil.org/browse/LT-22095 was caused by a bug fix I made in December:

https://github.com/sillsdev/FieldWorks/commit/a41e13c9fec5cc59ea53b621439e20f1b11193f4

I fixed LT-22095 by not postponing PropChanged when duplicating an Affix Template.
This does not appear to reintroduce the bug that was fixed in December.

### DIFF
--- a/Src/FdoUi/FdoUiCore.cs
+++ b/Src/FdoUi/FdoUiCore.cs
@@ -400,7 +400,7 @@ namespace SIL.FieldWorks.FdoUi
 			switch (classId)
 			{
 				default:
-					return DefaultCreateNewUiObject(classId, hvoOwner, flid, insertionPosition, cache);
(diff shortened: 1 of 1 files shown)


jtmaxwell3:
Should I change m_postponePropChanged so that it is off by default and only turn it on when we encounter a problem?  This would make it less likely that new bugs like this one get introduced, but I'm not sure I can find all the places it should be turned on.

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/311)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #314, opened 2025-04-09
Fix LT-12587: Bulk Edit-List Choice should allow Not Sure

jtmaxwell3:
I changed the code so that selecting Not Sure causes inflection features to be cleared.

Should this pull request be postponed until the test team is done testing the latest version of FieldWorks?

### DIFF
--- a/Src/FdoUi/InflectionFeatureEditor.cs
+++ b/Src/FdoUi/InflectionFeatureEditor.cs
@@ -39,6 +39,7 @@ namespace SIL.FieldWorks.FdoUi
 		InflectionFeaturePopupTreeManager m_InflectionFeatureTreeManager;
 		int m_selectedHvo = 0;
 		string m_selectedLabel;
+		bool m_notSure = false;
 		private int m_displayWs = 0;
 		public event EventHandler ControlActivated;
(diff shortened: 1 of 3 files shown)


jtmaxwell3:
Should <Not Sure> be renamed <None> or <Empty>?

jasonleenaylor:
I think there should be a corresponding change in FakeDoIt so that the preview shows correctly, also I'm wondering if there is a way to do this without making the menu item text public. I generally like to keep strings contained in their project. So if we can move the 'Not Sure' to a resource that is already publicly available to both locations or use other information that would be good.

jtmaxwell3:
I changed the code to identify the NotSure node using hvoNode.Tag instead of hvoNode.Text.

The preview already shows correctly because both DoIt and FakeDoIt use IsItemEligible to determine what to show.  Only DoIt has to have special code for setting NotSure.

jasonleenaylor (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/314)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #320, opened 2025-04-16
Fix LT-22036: Change node identification algorithm

jasonleenaylor:
* Stop using hashcode and use a customized node path id generating algorithm
* Also update an obsolete variable name

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/320)


hahn-kev (approved):
Looks good to me, there should be a separator between nodes in the node path string.

PR #326, opened 2025-04-22
LT-22099: Fix XHTML and Word Style problems

mark-sil:
We were not displaying the correct styles for Subentries and Subsenses (nodes that are shared). The problem was that we were using the style from the parent node, when we needed to use the style from the node that was making the reference.  This change keeps a node list, which allows us to collect all the style information from where the node is used, not necessarily from the parent node.

Change-Id: Id8c2adbc7c641831be802e61bfec834d61b616c9

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/326)


aror92 (approved):
:lgtm:

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/326#-:-OOc8WenDBYbUIDqux9r:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #329, opened 2025-04-24
Word Export: LT-22106: Leave ws tag on styles

mark-sil:
If a style with a ws tag was identical to the same style without the ws tag (the default) then in an effort to reduce the number of styles in the Word document we were changing all references to the style with the ws tag to the style name without the ws tag. That optimization was undesirable so we now leave the style with the ws tag.

Change-Id: I7dbdc54c2581bc5cbcc46eaab7be65b7901836f5

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/329)


aror92 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/329)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #330, opened 2025-04-25
Word Export: Fix caption width

aror92:
Fix textbox width to match the user-specified max width.

Otherwise, if textbox width matches the actual image width, then when images are narrower than the max width
the textbox containing picture and caption will be narrow, and there will be less space for the caption than in other textboxes.

Example of current behavior with textbox width = image width:
image

Example of new behavior with textbox width = max width:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/330)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/330)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @aror92)

PR #343, opened 2025-05-06
LT-22048 Make picture copyright use analysis WS style

aror92:
WS to use for picture copyright is already specified as an argument in AddProperty and used in the Word Export. Add handling in the xhtml generator to add a lang element to the property.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/343)


jasonleenaylor:
@aror92 
I can think of several reasons not to do it, but none of them are particularly good. You might want to raise it to Ken's attention to see if he anticipates any problems. (Also I assume you saw the tests which need adjusting after this change)

aror92:
Ken gave the go-ahead to use first analysis WS as default for properties that are not WS aware.

PR #347, opened 2025-05-13
Fix LT-21461: Problems with Choose Allomorph dialog box

jtmaxwell3:
Double-clicking on an abstract morpheme no longer caused a crash, but the Choose Allomorph dialog box wasn't displaying any morphemes after the first search.  This was because the French DefaultCollation was being changed after the first search by RecordSorter, and then the DefaultCollation no longer matched the data in the searcher.  The DefaultCollation was being changed from a standard collation to the system collation to a void a bug when a standard collation doesn't have any rules.  I copied the code to change the DefaultCollation to the initialization code in FieldWorks.EnsureDefaultCollationsPresent to avoid this problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/347)

### DIFF
--- a/Src/Common/Controls/XMLViews/MatchingObjectsBrowser.cs
+++ b/Src/Common/Controls/XMLViews/MatchingObjectsBrowser.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2014-2017 SIL International
+// Copyright (c) 2014-2017 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
@@ -138,6 +138,16 @@ namespace SIL.FieldWorks.Common.Controls
 				m_startingObject = value;
 			}
 		}
+
+		/// <summary>
+		/// Delegate for FilterResult.
+		/// </summary>
+		public delegate bool FilterResultDelegate(int hvo);
+
+		/// <summary>
+		/// Should we filter a result?
+		/// </summary>
+		public FilterResultDelegate FilterResult;
 
 		/// <summary>
 		/// Used by a Find dialog's SearchEngine to determine whether to search on a particular field or not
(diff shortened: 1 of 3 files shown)


mark-sil (changes requested):
In the pull request comment you said:
"I copied the code to change the DefaultCollation to the initialization code in FieldWorks.EnsureDefaultCollationsPresent to avoid this problem."
Should the code be removed from the old location?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/347#-:-OQDt16rEQrvZg2JgeR8:b-zd9zp2)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`Src/Common/FieldWorks/FieldWorks.cs` line 881 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/347#-OQDqyHI8HOx-BH2TUAg:-OQDqyHI8HOx-BH2TUAh:bvbezsx) ([raw file](https://github.com/sillsdev/FieldWorks/blob/962e2d50f146abd84b097eef2b49e96ed5c7636c/Src/Common/FieldWorks/FieldWorks.cs#L881)):*
> ```Smalltalk
> 				}
> 				// Check HasValidRules here rather than in RecordSorter to avoid LT-21461 problem.
> 				if (ws != null && ws.DefaultCollation != null && !HasValidRules(ws.DefaultCollation))
> ```

If we get in the preceding if statement, then is it true that we would never need to call HasValidRules() and execute the code in this if statement? If that is correct, then should this be an 'else if'?
___
*[`Src/Common/FieldWorks/FieldWorks.cs` line 899 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/347#-OQDq4Y6AkJ5M3EUIclT:-OQDq4Y6AkJ5M3EUIclU:bc49019) ([raw file](https://github.com/sillsdev/FieldWorks/blob/962e2d50f146abd84b097eef2b49e96ed5c7636c/Src/Common/FieldWorks/FieldWorks.cs#L899)):*
> ```Smalltalk
> 		/// This method supports the band-aid fallback to SystemCollationDefinition.
> 		/// </summary>
> 		/// <returns>true if the CollationDefinition is a RulesCollationDefinition with valid non empty rules</returns>
> ```

It looks like this method also returns true if cd is not a RulesCollationDefinition. If that is the intended behavior, then this comment should be changed.

jtmaxwell3:
I didn't remove the code from the old location because I was worried that there might be other ways that the situation occurred.  For instance, if the user created a new writing system with no collation rules after the project was loaded.

mark-sil (approved):
> Reviewed 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/347)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #353, opened 2025-05-22
LT-21996: Don’t create a link when there is no target

mark-sil:
Regardless of whether we are generating an entry for viewing in Flex or for exporting, use the same rules to determine if links to other entries should be created. If the target entry is not displayed then don’t generate a link (Guid).

Change-Id: I3cc000e3d5cace5f9041c92a0b6ad5255e3ac3ae

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/353)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -346,7 +346,6 @@ namespace SIL.FieldWorks.XWorks
 			}
 
 			// Second check if we are publishing minor entries.
-			//if (displayed && entryType != EntryType.mainEntry && !lexEntry.PublishAsMinorEntry)
 			if (displayed && !mainEntry && !lexEntry.PublishAsMinorEntry)
 			{
 				displayed = false;
@@ -361,6 +360,8 @@ namespace SIL.FieldWorks.XWorks
 				{
 					displayed = false;
 				}
+				// Note: A better name for ShowMainEntryIn() would probably be
+				// ShowAsHeadwordIn(), since it applies to both main and minor entries.
 				if (!lexEntry.ShowMainEntryIn.Contains(currentPubPoss))
 				{
 					displayed = false;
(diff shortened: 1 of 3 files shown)


aror92 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/353)* status: all files reviewed, 2 unresolved discussions (waiting on @mark-sil)

___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 349 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/353#-ORHArpL8vfJou4euUVc:-ORHArpL8vfJou4euUVd:ble8yvh) ([raw file](https://github.com/sillsdev/FieldWorks/blob/8704c9656782e6aa3ce4fdac6f2d3e42bb116d37/Src/xWorks/ConfiguredLcmGenerator.cs#L349)):*
> ```Smalltalk
> 
> 			// Second check if we are publishing minor entries.
> 			//if (displayed && entryType != EntryType.mainEntry && !lexEntry.PublishAsMinorEntry)
> ```

probably want to delete this; it looks like a commented line that was missed in clean-up
___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 366 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/353#-ORHBfXU-AM7_VnfjAIX:-ORHBfXU-AM7_VnfjAIY:b-fbnyql) ([raw file](https://github.com/sillsdev/FieldWorks/blob/8704c9656782e6aa3ce4fdac6f2d3e42bb116d37/Src/xWorks/ConfiguredLcmGenerator.cs#L366)):*
> ```Smalltalk
> 				if (!lexEntry.ShowMainEntryIn.Contains(currentPubPoss))
> 				{
> 					displayed = false;
> ```

Want to make sure I understand the logic here: if we're dealing with a main entry, the logic is clear. In the case that we're dealing with a minor entry, why does the condition above mean displayed needs to be false? Is it indicating that the main entry associated with the minor entry should not be shown and therefore neither should this minor entry?

aror92 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/353#-:-ORMDehjC6gpNk3ar58t:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #398, opened 2025-07-16
LT-22122: Word Export performance improvement #2

mark-sil:
There was data in a DocFragment that was only needed by the fragment that was going to write to the docx file. We now moved that data to a MasterDocFragment and only have a Body in the regular DocFragment. Regular DocFragments are frequently created and discarded.  Not having them open a memory stream saves a significant amount of time.  Combined with the performance improvements in commit e1943fb6, export times are now 30% to 50% of previous times.

Change-Id: I9ec6fa92709d3562c9331ec30c4c2920a71468b4

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/398)

### DIFF
--- a/Src/GenerateHCConfig/App.config
+++ b/Src/GenerateHCConfig/App.config
@@ -12,7 +12,7 @@
       </dependentAssembly>
 	  <dependentAssembly>
 		<assemblyIdentity name="SIL.Lexicon" publicKeyToken="cab3c8c5232dfcf2" culture="neutral" />
-			<bindingRedirect oldVersion="0.0.0.0-13.0.0.0" newVersion="13.0.0.0" />
+			<bindingRedirect oldVersion="0.0.0.0-15.0.0.0" newVersion="15.0.0.0" />
 	  </dependentAssembly>
     </assemblyBinding>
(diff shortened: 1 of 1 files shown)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/398)* status: all files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`Src/xWorks/LcmWordGenerator.cs` line 944 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/398#-OVJtUEq1umg6OboK8j6:-OVNDwBzDU-lpA5ATvym:b-va4cgo) ([raw file](https://github.com/sillsdev/FieldWorks/blob/ddf828323a1feefae8fb68a5ccd3762635dc5066/Src/xWorks/LcmWordGenerator.cs#L944)):*

Do we want a Flush method like this for the MasterDocFragment class?
</blockquote></details>

If there is a need for it then we can add it at that time.  This Flush method is never being called on what is now the MasterDocFragment, so I didn’t see a reason to add it.

PR #405, opened 2025-07-18
LT-22203: punct only line; multiple punct; rare no morph

AndyBlack:
This fixes three issues in XLingPaper interlinear export.  The fixes are all in the xml2XLingPapCommonConcatMorphemes.xsl file.  The diff may be a bit hard to read since it looks like a lot of changes, but it's a case of having to create some "choose/when" items to fix them so a lot of code which stays the same is now in a different location.
The three issues are:
1. If a line begins with two punct items, only the first is output.  The fix is:

<xsl:value-of select="normalize-space(.)"/>

2. If a line has no vernacular words in it at all (as in the new BruceCoxEmpty test case), then the punct items are not output at all and there is an empty <lineGroup/> element in the XLingPaper output.  The fix is in the first (new) choose/when case with this comment:

3. In rare cases (as illustrated in the new Jibiyal3Text test case), the FLExText would not have a word with a failed morphemes indication. (as in text 59 in the text)  The fix is the code in this choose/when case:
<xsl:when test="name()='morph'">

As mentioned, there are two new test cases to check for these issues.  The existing nszEnglishWords test case needed to be updated because it, too, had cases fixed by number 1.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/405)


jasonleenaylor (approved):
Sorry, xslt takes me a while to review. It makes my brain hurt. :lgtm:

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/405#-:-OW1MmIoDW0mza6k7yeX:b-gpd2p3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AndyBlack)

AndyBlack:
Yes, it can be challenging.  I wish these changes did not make it even harder with so much actually being the same, just in a different location.
Thanks for doing it, though.

PR #417, opened 2025-07-29
LT-21960: Fix export issues when there is filtering

mark-sil:
When there is filtering applied to the entries, then only the currently viewed entries would export correctly.  (If the main dictionary was being viewed, then the filtered reversal entries would not export correctly. If a reversal was currently being viewed, then a filtered main dictionary would not export correctly.)
The same problem existed for both Word and Xhtml exports. To resolve the problem with filtered entries we now do the collection of entries on the main thread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/417)

### DIFF
--- a/Src/xWorks/DictionaryExportService.cs
+++ b/Src/xWorks/DictionaryExportService.cs
@@ -21,8 +21,8 @@ namespace SIL.FieldWorks.XWorks
 		private readonly Mediator m_mediator;
 		private readonly LcmCache m_cache;
 
-		private const string DictionaryType = "Dictionary";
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/417#-:-OWRat4WAEOkhBKd6b6h:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #431, opened 2025-08-06
LT-22124: Fix Sense homographs in Lexical Relations

mark-sil:
When trying to get the HeadWordRef for a LexSense, get the value from the associated LexEntry. The LexEntry has a custom field that adjusts the HeadWordRef based on the decorator, the LexSense does not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/431)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -682,10 +682,10 @@ namespace SIL.FieldWorks.XWorks
 		}
 
 		/// <summary>
-		/// Gets the value of the requested custom field associated with the fieldOwner object
+		/// Gets the value of the requested field associated with the fieldOwner object
 		/// </summary>
-		/// <returns>true if the custom field was valid and false otherwise</returns>
-		/// <remarks>propertyValue can be null if the custom field is valid but no value is stored for the owning object</remarks>
+		/// <returns>true if the field was valid and false otherwise</returns>
+		/// <remarks>propertyValue can be null if the field is valid but no value is stored for the owning object</remarks>
 		private static bool GetPropValueForModelField(object fieldOwner, ConfigurableDictionaryNode config,
 			LcmCache cache, ISilDataAccess decorator, string customFieldName, ref object propertyValue, string cfOwnerClassName = null)
 		{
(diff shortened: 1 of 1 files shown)


mark-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/431)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/431#-OX-4oTY2pB_tjZCIlkX:-OX-4oTY2pB_tjZCIlkY:bah9yna):*
Is the second sentence in this comment correct? If it is correct, then is it ok to get the HeadWordRef from the associated LexEntry when we have a LexSense? If it is not ok, then should I add a custom field for the LexSense that is created and maintained similar to what is done for the LexEntry?
___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 705 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/431#-OX-53UYDPI5QtSwwNqN:-OX-53UYDPI5QtSwwNqO:bltqzea) ([raw file](https://github.com/sillsdev/FieldWorks/blob/801b0729c1df24f376cb380a5cfd237c56b29f5b/Src/xWorks/ConfiguredLcmGenerator.cs#L705)):*
> ```Smalltalk
> 				{
> 					// Get the HeadWordRef from the LexEntry.
> 					if (customFieldName == "HeadWordRef")
> ```

Instead of specifically checking for "HeadWordRef", is there something more general that I should be checking?

jasonleenaylor (commented):
Sorry that it took me all day.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/431#-:-OX0Hjh_CV1Igy3nJNQ7:b-idehws)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @mark-sil)

___
*[`-- commits` line 8 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/431#-OX-4oTY2pB_tjZCIlkX:-OX0Gnwi7hZZjTzlBIIx:b-dhuiep):*

Is the second sentence in this comment correct? If it is correct, then is it ok to get the HeadWordRef from the associated LexEntry when we have a LexSense? If it is not ok, then should I add a custom field for the LexSense that is created and maintained similar to what is done for the LexEntry?
</blockquote></details>

The comment isn't quite correct, because this isn't actually a custom field. The customFieldName in GetPropValueForModelField should be renamed to fieldName (this used to be used only for custom fields).

I think we probably should be handling the HeadWordRef for senses somewhere else.
___
*[`Src/xWorks/ConfiguredLcmGenerator.cs` line 705 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/431#-OX-53UYDPI5QtSwwNqN:-OX0GDs404-bUL0hzAJP:b1qcicc) ([raw file](https://github.com/sillsdev/FieldWorks/blob/801b0729c1df24f376cb380a5cfd237c56b29f5b/Src/xWorks/ConfiguredLcmGenerator.cs#L705)):*

Instead of specifically checking for "HeadWordRef", is there something more general that I should be checking?
</blockquote></details>

After much head scratching I don't think this is the correct solution. It seems that there should be a Sense method which uses the decorator to get the information...and it feels like it should already exist but it just isn't being used by this export yet.

PR #435, opened 2025-08-12
Fix LT-22187: Allow reordering fields

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22187.  When you move a field, it moves it after the part for the next slice.  For this to work, you have to get the right part for the slice (the one that can be moved).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/435)


jtmaxwell3:
Removing the Lexeme Form label was necessary.  It was redundant with the label in AsLexemeForm, and it made the code think that there was a header above AsLexemeForm that could be moved, something like:

Lexeme Form
    Lexeme Form
        Morph Type

Then the code determined that AsLexemeForm could not be move because it had no siblings.
Deleting the higher Lexeme Form allowed AsLexemeForm to be move.

If we don't want Lexeme Form to be moved, then we can restore the higher Lexeme Form label.

jasonleenaylor (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/435#-:-OXZ3c_P5j0DU57J7F_9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jtmaxwell3)

PR #437, opened 2025-08-13
LT-22124: Fix Sense homographs in Lexical Relations

mark-sil:
When getting a property value for a field on a ISenseOrEntry object, call the new method added to LCM that determines the specific object (Sense or Entry), and the specific field on that object, that should be queried to get the value.

In this specific defect we are trying to get the HeadWordRef and the ISenseOrEntry is a LexSense. The specific field remains HeadWordRef, but the specific object we need to get the value from is the associated LexEntry. This is because the LexEntry adjusts the HeadWordRef based on the decorator, the LexSense does not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/437)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -682,26 +682,23 @@ namespace SIL.FieldWorks.XWorks
 		}
 
 		/// <summary>
-		/// Gets the value of the requested custom field associated with the fieldOwner object
+		/// Gets the value of the requested field associated with the fieldOwner object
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/437#-:-OXZ4RrAD79ZbXGT1Eo3:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

PR #442, opened 2025-08-15
Fix LT-22249: Don't modify existing analysis on flextext import

jasonleenaylor:


jtmaxwell3 (commented):
When I looked at this code today, the bug was that FindExistingGloss found an analysis that matched the wordform and the gloss but not the morphemes, and that the code that followed FindExistingGloss changed the morphemes.  Don't you have the same problem if FindExistingAnalysisStack fails?

jtmaxwell3 (approved):
I read through all of the code and I didn't see any problems.  I'm trusting the test cases to catch anything I might have missed in the code.

PR #449, opened 2025-08-21
LT-21042: XAmple derivational infix has wrong orderclass sign

AndyBlack:
The XAmple transform that produces the lexicon input file used the wrong order class sign for an infix.  It should have been with a minus.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/449)


jtmaxwell3 (approved):
Will this cause existing XAmple grammars to break?

AndyBlack:
No, it should not cause any grammars to break,  Rather, it should make some parse better than before.

PR #453, opened 2025-09-02
WIP: Fix the extra span being generated (There may be a better fix)

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/453)

### DIFF
--- a/Src/xWorks/CssGenerator.cs
+++ b/Src/xWorks/CssGenerator.cs
@@ -400,7 +400,9 @@ namespace SIL.FieldWorks.XWorks
 						//Generate the rules for the default font info
 						rule.Declarations.Properties.AddRange(GenerateCssStyleFromLcmStyleSheet(configNode.Style, DefaultStyle, configNode,
 							propertyTable));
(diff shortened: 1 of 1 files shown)


Src/xWorks/CssGenerator.cs
```diff
@@ -400,7 +400,9 @@ private static List<StyleRule> GenerateCssFromConfigurationNode(ConfigurableDict
 						//Generate the rules for the default font info
 						rule.Declarations.Properties.AddRange(GenerateCssStyleFromLcmStyleSheet(configNode.Style, DefaultStyle, configNode,
 							propertyTable));
-						selectors.AddRange(GenerateCssForWritingSystems(baseSelection + " span", configNode.Style, propertyTable));
+						// add span for the writing system if we aren't already working on one
+						baseSelection = baseSelection.EndsWith(" span") ? baseSelection : baseSelection + " span";
```
hahn-kev, line 404: assuming this is a css selector could you do something like this?
```css
baseSelection, baseSelection span {
   styles...
}
```
the comma here means OR. So if it already ends with `span` then you'll generate:
`base span, base span span` as the selector.

probably not what you want, but the only suggestion I can think of that might spark some other kind of idea.

PR #464, opened 2025-09-17
LT-21960: Fix export for multiple reversals and filtering

mark-sil:
When exporting to Word, Webonary, or XHTML, we now get the entry data from direct data access, instead of relying on the GUI to display the data we want to get.
The GUI updates were mostly being done through property change notifications or mediator messages.  The problems we were seeing was when these notifications and messages were combined with the many existing special cases where we don’t want to update the GUI, or we want to delay the GUI update. Getting the combination correct to fix one situation was resulting in other situations that were incorrect.

Problems resolved:
- Exporting multiple reversals to Word or Webonary.
- Exporting multiple reversals with different filtering to Word or Webonary.
- Exporting to Word, Webonary, or XHTML immediately after starting Flex. In this state, depending on what is being viewed, either the main dictionary clerk or the reversal clerk has been created, but not both.
- The GUI no longer flash with visual updates while doing an export.

Notes:
Since now we are getting the entry text without calling XmlBrowseViewBaseVc.DisplayCell(), we no longer need the changes that were added to fix LT-21198. This PR also removes the change in Commit 0503457b that added OverrideWs.

Remaining Issues:
- Exporting multiple reversals with different sorting to Word or Webonary.
- When there are multiple reversals with different sorting then the display in Flex can also sometimes be incorrect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/464)

### DIFF
--- a/.github/workflows/base-installer-cd.yml
+++ b/.github/workflows/base-installer-cd.yml
@@ -199,8 +199,8 @@ jobs:
         working-directory: Installers
         shell: cmd
         run: |
-            insignia -ab online-engine.exe "${{ steps.stage_installers.outputs.offline_exe }}" -o "${{ steps.stage_installers.outputs.offline_exe }}"
-            insignia -ab offline-engine.exe "${{ steps.stage_installers.outputs.online_exe }}" -o "${{ steps.stage_installers.outputs.online_exe }}"
+            insignia -ab offline-engine.exe "${{ steps.stage_installers.outputs.offline_exe }}" -o "${{ steps.stage_installers.outputs.offline_exe }}"
+            insignia -ab online-engine.exe "${{ steps.stage_installers.outputs.online_exe }}" -o "${{ steps.stage_installers.outputs.online_exe }}"
 
       - name: Sign Bundles
         if: github.event_name != 'pull_request'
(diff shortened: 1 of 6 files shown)


jasonleenaylor (commented):
> @jasonleenaylor reviewed 39 of 41 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/464)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mark-sil)

___
*[`Src/xWorks/DictionaryConfigurationModel.cs` line 340 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/464#-O_Y6ALk8SLNxEssr-w_:-O_Y6ALk8SLNxEssr-wa:b-adpj8s) ([raw file](https://github.com/sillsdev/FieldWorks/blob/87735b5adfa75818da1bb0548cd02eb4509515c9/Src/xWorks/DictionaryConfigurationModel.cs#L340)):*
> ```Smalltalk
> 		/// Gets the reversal configuration model for the specified writing system.
> 		/// </summary>
> 		public static DictionaryConfigurationModel GetReversalConfigurationModel(string ws, LcmCache cache, PropertyTable propTable)
> ```

I'm sure this method existed somewhere already...or did you move it.
___
*[`Src/xWorks/RecordClerk.cs` line 476 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/464#-O_Y7xFz8rmOyIiFB6Hi:-O_Y7xFz8rmOyIiFB6Hj:b-ifye16) ([raw file](https://github.com/sillsdev/FieldWorks/blob/87735b5adfa75818da1bb0548cd02eb4509515c9/Src/xWorks/RecordClerk.cs#L476)):*
> ```Smalltalk
> 		{
> 			CheckDisposed();
> 			return m_list.PropertyTableId("sorter", dictConfig);
> ```

Does this handle 'null' safely?
___
*[`Src/xWorks/RecordClerk.cs` line 3092 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/464#-O_YIaFG27E-TDzgKGi1:-O_YIaFH6b8v3C5N5rz3:bqzw84g) ([raw file](https://github.com/sillsdev/FieldWorks/blob/87735b5adfa75818da1bb0548cd02eb4509515c9/Src/xWorks/RecordClerk.cs#L3092)):*
> ```Smalltalk
> 		/// </summary>
> 		/// <param name="reversalGuid">The Guid that identifies the reversal.</param>
> 		public int[] GetReversalFilteredAndSortedEntries(Guid reversalGuid, DictionaryPublicationDecorator decorator, DictionaryConfigurationModel revConfig)
> ```

I don't love having these methods here. RecordClerk should in theory be ignorant of such specific list types.
Could they be added as extension methods from some Dictionary or Reversal classes?



_Code quote:_
```Smalltalk
		/// <summary>
		/// Gets the filtered and sorted dictionary entries.
		/// </summary>
		public int[] GetDictionaryFilteredAndSortedEntries(DictionaryPublicationDecorator decorator)
		{
			// Filters and Sorts the list, then populates the virtual cache with the result.
			m_list.FilterAndSortList();

			// Gets the entries from the virtual cache.
			int[] retEntries = decorator.VecProp(Cache.LangProject.LexDbOA.Hvo, VirtualFlid);

			return retEntries;
		}

		/// <summary>
		/// Gets the filtered and sorted entries for the specified reversal.
		/// </summary>
		/// <param name="reversalGuid">The Guid that identifies the reversal.</param>
		public int[] GetReversalFilteredAndSortedEntries(Guid reversalGuid, DictionaryPublicationDecorator decorator, DictionaryConfigurationModel revConfig)
```

PR #814, opened 2026-04-06
Consolidate native and managed test reporting

papeh:
* Ensure that native tests are built by build.ps1 when requested
* Run both native and managed tests in test.ps1 by default
* Include stderr test logs in artifacts (stdout logs are already included)

Why Invoke-CppTest.ps1 builds native tests but build.ps1 did not remains a mystery to me.

This change [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/814) :grin:


jasonleenaylor (commented):
Devin had some good observations, I was going to say LGTM but it specifically had questions about if we allow a specific managed test project to be built without building TestGeneric as well.

PR #820, opened 2026-04-08
Installer: add prerequisites alias for vcredists

johnml1135:
## Summary
- add a backward-compatible `prerequisites` package-group alias that forwards to `vcredists`
- apply the alias in the tracked redistributables include files used by legacy WiX 3 and WiX 6 installer authoring
- prevent stale bundle references from breaking installer builds when older authoring still refers to `prerequisites`

## Verification
- `./Build/Agent/check-and-fix-whitespace.ps1`
- `./Build/Agent/commit-messages.ps1`
- `.\build.ps1 -BuildInstaller -Configuration Release`

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/820)

### DIFF
--- a/FLExInstaller/Redistributables.wxi
+++ b/FLExInstaller/Redistributables.wxi
@@ -27,6 +27,10 @@
 			<PackageGroupRef Id="redist_vc15to19" />
 			<PackageGroupRef Id="FlexBridgeInstaller" />
 		</PackageGroup>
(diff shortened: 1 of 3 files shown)


papeh (changes requested):
I think we can just change the main PackageGroup Id when genericinstaller supports the more generic name.

PR #826, opened 2026-04-09
LT-22482: localize HCSynthByGloss

AndyBlack:
This adds English (by default), French, German, and Spanish for the 13 strings used in the HcSynthByGloss tools.
The EXE puts the locale code as its first argument.
The DLL add a property to set the locale code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/826)


papeh (changes requested):
That's great that you already have localizations for de, es, and fr. Rather than checking them into the repo, they should be uploaded to Crowdin. I've sent you an invitation.

Src/Utilities/HCSynthByGloss/HCSynthByGloss/HCSynthByGloss.cs
```diff
@@ -19,58 +19,72 @@
...
-						"HCSynthByGloss -h HC.xml_file -g gloss_file -o output (-t (-s))"
-					);
-					Console.WriteLine("\t-t = turn on tracing");
-					Console.WriteLine(
-						"\t-s = show the tracing result in the system default web browser; -s is only valid when also using -t"
-					);
+					Console.WriteLine(HCSynthByGlossStrings.ksUsage);
+					Console.WriteLine(HCSynthByGlossStrings.ksCommandLineUsage);
+					Console.WriteLine(HCSynthByGlossStrings.ksTurnOnTracing);
+					Console.WriteLine(HCSynthByGlossStrings.ksShowTracing);
```
papeh: If these are always used together, I think it would be best to combine them into a single string.

Src/Utilities/HCSynthByGloss/HCSynthByGloss/HCSynthByGloss.cs
```diff
@@ -19,58 +19,72 @@
...
+using System.Globalization;
+using System.Threading;
 
 namespace SIL.HCSynthByGloss
 {
 	class HCSynthByGloss
 	{
 		static void Main(string[] args)
 		{
+			const int iLocale = 0;
```
papeh, line 32: Clever use of constants to make command line args meaningful!

Src/Utilities/HCSynthByGloss/HCSynthByGlossLib/HCSynthByGlossStrings.resx
```diff
@@ -0,0 +1,159 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ksCommandLineUsage" xml:space="preserve">
+    <value>HCSynthByGloss locale -h HC.xml_file -g gloss_file -o output (-t (-s))</value>
+  </data>
+  <data name="ksCouldNotFind" xml:space="preserve">
+    <value>Could not find </value>
```
papeh: To support languages with different word order, you could use `string.Format(` with
```
<data name="ksCouldNotFind" xml:space="preserve">
    <value>Could not find {0} '{1}'.</value>
    <comment>{0} is 'HC XML file' or 'Gloss file'. {1} is the path to (name of?) the file.</comment>
</data>
```

Src/Utilities/HCSynthByGloss/HCSynthByGlossLib/Synthesizer.cs
```diff
@@ -186,7 +187,7 @@ private bool CheckForDuplicates(Morpher morpher, StringBuilder sb, List<Morpheme
 			}
 			if (duplicateFound)
 			{
-				sb.Append(" synthesis may not work.");
```
papeh, line 189: This sentence could also be combined into a single string.

AndyBlack: The preceding message is within a foreach loop and can include multiple glosses after the ksDuplicateGlossFoundFor string.  We shoujd leave this as it is.

papeh: How about
```
var duplicateBuilder = new StringBuilder();
foreach (Morpheme morph in morphemes)
{
    . . .
    if (duplicateGloss != null)
        duplicateBuilder.AppendFormat("'{0}'; ", morph.Gloss);
}
var duplicates = duplicateBuilder.ToString().Trim();
if (!string.IsNullOrEmpty(duplicates))
{
    sb.AppendFormat(HCSynthByGlossStrings.ksDuplicateGlossFoundForXSythesisMayNotWork, duplicates);
}

File: Src/Utilities/HCSynthByGloss/HCSynthByGlossLib/HCSynthByGlossStrings.resx [hunk 5]
```diff
@@ -136,24 +132,26 @@
...
   </data>
   <data name="ksProcessingResult" xml:space="preserve">
     <value>Processing result: </value>
   </data>
-  <data name="ksShowTracing" xml:space="preserve">
-    <value>\t-s = show the tracing result in the system default web browser; -s is only valid when also using -t</value>
+  <data name="ksQuoteBegin" xml:space="preserve">
+    <value>'</value>
+  </data>
+  <data name="ksQuoteEnd" xml:space="preserve">
```
papeh: I would replace these two with `<value>'{0}'; </value>` so that the string is more clearly a unit

Src/Utilities/HCSynthByGloss/HCSynthByGlossLib/HCSynthByGlossLib.csproj
```diff
@@ -46,4 +46,29 @@
...
     </Compile>
   </ItemGroup>
+  <ItemGroup>
+    <Compile Update="HCSynthByGlossStrings.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>HCSynthByGlossStrings.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
+  <ItemGroup>
```
papeh, line 49: Only the main resx and designer need to be included. When FieldWorks builds localizations, it will automatically include these.

Src/Utilities/HCSynthByGloss/HCSynthByGlossDll/HCSynthByGlossDll.cs
```diff
@@ -55,7 +60,7 @@ public string SetGlossFile(string value)
 
 		public string Process()
 		{
-
+			Thread.CurrentThread.CurrentUICulture = CultureInfo.GetCultureInfo(LocaleCode);
```
papeh: This shouldn't be necessary, since FLEx handles localization centrally.

Src/Utilities/HCSynthByGloss/HCSynthByGlossLib/HCSynthByGlossLib.csproj
papeh: Oops, should have kept the main resx and removed the language-specific ones.

papeh (approved):

PR #828, opened 2026-04-10
Fix LT-22475: The tab order is wrong for "Create and Edit" button

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22475.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/828)


papeh (changes requested):
I think a better place to set the tab order would be in the resx file.

jtmaxwell3:
@papeh: We can't do it in the resx file since the Create and Edit button is only included in some contexts.  When it is added, the downstream buttons have to have their TabIndex values incremented.

papeh:
Tab orders are allowed to be nonconsecutive, so hiding the button should not cause a problem.

papeh (approved):

PR #831, opened 2026-04-13
Fix LT-22486: Click Inserts Invisible Space crashes

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22486.  InvisibleSpaceCursor.cur was not accessible to GetCursor() because it wasn't marked as an EmbeddedResource.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/831)


AndyBlack (approved):
lgtm

PR #833, opened 2026-04-13
Fix LT-19561: "Use Default Folder" button was truncated

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-19561 by moving the button to the left so that it isn't clipped by the window.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/833)


AndyBlack (approved):
lgtm

PR #836, opened 2026-04-14
LT-21517: Implement DialogInsertItemInVector in Pub/Sub system

mark-sil:
Add a ReturnObject class to return a ‘handled’ value, which the Publisher needs to know. All of the Subscribers only handle a specific className or className and Id combination so it did not matter that two of the classes are derived from DlgListenerBase, which has ColleaguePriority.High.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/836)

### DIFF
--- a/Src/Common/FwUtils/PublisherParameterObject.cs
+++ b/Src/Common/FwUtils/PublisherParameterObject.cs
@@ -8,7 +8,7 @@ namespace SIL.FieldWorks.Common.FwUtils
 {
 	public sealed class PublisherParameterObject
 	{
-		public PublisherParameterObject(string message, object newValue = null)
+		public PublisherParameterObject(string message, object data = null)
 		{
 			if (string.IsNullOrWhiteSpace(message))
 			{
@@ -16,10 +16,25 @@ namespace SIL.FieldWorks.Common.FwUtils
 			}
 
 			Message = message;
-			NewValue = newValue;
+			Data = data;
 		}
 
 		public string Message { get; }
-		public object NewValue { get; }
+		public object Data { get; }
+	}
+
+	/// <summary>
+	/// Convenience class for use when we want to pass a return value back through the PublisherParameterObject.
+	/// </summary>
+	public class ReturnObject
+	{
+		public ReturnObject(object data)
+		{
+			Data = data;
+			ReturnValue = false;
+		}
+
+		public object Data { get; set; }
+		public bool ReturnValue { get; set; }
 	}
 }
--- a/Src/LexText/LexTextControls/EntryDlgListener.cs
+++ b/Src/LexText/LexTextControls/EntryDlgListener.cs
@@ -75,7 +75,7 @@ namespace SIL.FieldWorks.LexText.Controls
 			}
 			// Only handle "LexEntry" class.
 			string className = XmlUtils.GetOptionalAttributeValue(command.Parameters[0], "className");
-			if (className == null || className != "LexEntry")
+			if (className != "LexEntry")
 			{
 				return;
 			}
--- a/Src/LexText/LexTextControls/EntryDlgListener.cs
+++ b/Src/LexText/LexTextControls/EntryDlgListener.cs
@@ -7,6 +7,8 @@
 using System;
 using System.Diagnostics;
 using System.Windows.Forms;
+using SIL.FieldWorks.Common.FwUtils;
+using static SIL.FieldWorks.Common.FwUtils.FwUtils;
 using SIL.LCModel;
 using SIL.LCModel.Infrastructure;
 using XCore;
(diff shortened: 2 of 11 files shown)


Src/LexText/LexTextControls/EntryDlgListener.cs
```diff
@@ -32,28 +34,51 @@ protected override string PersistentLabel
...
+				return;
+			}
+			// Return if already handled by another Subscriber.
+			if (retObj.ReturnValue)
+			{
+				return;
+			}
+			// Only handle "LexEntry" class.
+			string className = XmlUtils.GetOptionalAttributeValue(command.Parameters[0], "className");
 			if (className == null || className != "LexEntry")
```
papeh: While you're in the area, `className == null` is redundant to `className != "LexEntry"`

Src/LexText/Morphology/MasterPhonFeatDlgListener.cs
```diff
@@ -57,27 +60,51 @@ public MasterPhonFeatDlgListener()
...
+			string className = XmlUtils.GetOptionalAttributeValue(command.Parameters[0], "className");
 			if ((className == null) || (className != "FsClosedFeature"))
-				return false;
-			if (className == "FsClosedFeature" && (argument as XCore.Command).Id != "CmdInsertPhonologicalClosedFeature")
-				return false;
+			{
+				return;
+			}
+			// Only handle "FsClosedFeature" for "CmdInsertPhonologicalClosedFeature".
+			if (className == "FsClosedFeature" && command.Id != "CmdInsertPhonologicalClosedFeature")
```
papeh: we just checked `className == "FsClosedFeature"`.

If you want to clean up, you could combine these two if statements.

Src/LexText/LexTextControls/EntryDlgListener.cs
```diff
@@ -32,28 +34,51 @@ protected override string PersistentLabel
...
 			CheckDisposed();
 
-			Debug.Assert(argument != null && argument is XCore.Command);
-			string className = XmlUtils.GetOptionalAttributeValue(
-				(argument as Command).Parameters[0],
-				"className");
+			if (!(obj is ReturnObject retObj) ||
+				!(retObj.Data is Command command))
+			{
+				Debug.Assert(false, "Received unexpected object type.");
```
papeh, line 68: nit: could use `Debug.Fail` or even a yellow-box crash

Src/Common/FwUtils/PublisherParameterObject.cs
```diff
@@ -8,18 +8,33 @@ namespace SIL.FieldWorks.Common.FwUtils
...
+	public class ReturnObject
+	{
+		public ReturnObject(object data)
+		{
+			Data = data;
+			ReturnValue = false;
+		}
+
+		public object Data { get; set; }
+		public bool ReturnValue { get; set; }
```
papeh, line 38: Would `Handled` be a better name?

papeh (commented):
> @papeh reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/836)* status: all files reviewed, 4 unresolved discussions (waiting on mark-sil).

Src/Common/FwUtils/PublisherParameterObject.cs
mark-sil: In this case I agree, Handled is a better name, but my intent is for this to be generic so the next time it is used it also makes sense, which might be something other than Handled.

Src/LexText/LexTextControls/EntryDlgListener.cs
mark-sil: The old code had an Assert to check Command, I think leaving it as an Assert but also checking ReturnObject is sufficient. It doesn't seem like it needs to change.

Src/LexText/LexTextControls/EntryDlgListener.cs
mark-sil: Done.

Src/LexText/Morphology/MasterPhonFeatDlgListener.cs
mark-sil: Done.

mark-sil (commented):
> @mark-sil made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/836)* status: 6 of 11 files reviewed, 4 unresolved discussions (waiting on papeh).

Src/Common/FwUtils/PublisherParameterObject.cs
papeh: I'm pretty sure `Handled` is the only case, but `ReturnValue` is acceptable.

papeh (approved):

PR #847, opened 2026-04-21
LT-22425: Fix testSetTextEmpty for Release builds

papeh:
## Summary
Sleep 1ms at the beginning of the MockTextStoreACPSink ctor to prevent an "unknown exception" in the test for VwTextStore.SetTextEmpty in Release builds (Debug builds are fine).

## Also
* Add a -SkipManagedTests switch to build.ps1
* Split tests into a separate execution step for installer builds for consistency with CI builds and more manageable log sizes

Another option is to `cout << "";`, which also spends enough time doing effectively nothing to cause the test to pass, and doesn't require two extra `#include`s, but sleeping rather than couting when the goal is to kill time feels more principled to me.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/847)

### DIFF
--- a/.github/workflows/base-installer-cd.yml
+++ b/.github/workflows/base-installer-cd.yml
@@ -161,7 +161,7 @@ jobs:
       - name: Run tests
         shell: powershell
         run: |
-          .\test.ps1 -Configuration Release -NoBuild -TestFilter 'TestCategory!=DesktopRequired'
+          .\test.ps1 -Configuration Release -NoBuild -SkipManaged
 
       - name: Scan Build Output
         shell: powershell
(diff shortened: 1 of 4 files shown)


papeh:
This works, and the logs look nice in the GHA web interface, but the downloaded build.log is missing the native logs. https://github.com/sillsdev/FieldWorks/actions/runs/24748859269 tests a tweak that I hope will fix this.

mark-sil (approved):
> @mark-sil reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/847)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on papeh).

PR #848, opened 2026-04-22
Fix problem with LT-22333: Click on field in Preview pane

jtmaxwell3:
This fixes a problem with the fix for https://jira.sil.org/browse/LT-22333 that was noticed by Marlon.  The Jump to field menu item should only appear in the Preview section of Lexicon Edit.  It shouldn't appear in the Dictionary view.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/848)


AndyBlack (approved):
lgtm

PR #858, opened 2026-04-30
Fix new problem with LT-11746: Make Preview aware of publication

jtmaxwell3:
This fixes a new problem with https://jira.sil.org/browse/LT-11746: Lexical relations shouldn't be shown if they are not in the current publication.  The solution was to pass in a DictionaryPublicationDecorator like the one in XhtmlDocView.cs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/858)

### DIFF
--- a/Src/xWorks/LcmXhtmlGenerator.cs
+++ b/Src/xWorks/LcmXhtmlGenerator.cs
@@ -51,7 +51,7 @@ namespace SIL.FieldWorks.XWorks
 			var xhtmlPath = Path.ChangeExtension(preferredPath, "xhtml");
 			try
 			{
-				SavePublishedHtmlWithStyles(entryHvos, clerk, publicationDecorator, entriesPerPage, configuration, propertyTable, xhtmlPath, progress, false, isLexEditPreviewOnly);
+				SavePublishedHtmlWithStyles(entryHvos, clerk, publicationDecorator, entriesPerPage, configuration, propertyTable, xhtmlPath, progress, isLexEditPreviewOnly: isLexEditPreviewOnly);
 			}
 			catch (IOException ioEx)
 			{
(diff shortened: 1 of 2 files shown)


jasonleenaylor:
*[`Src/xWorks/XhtmlRecordDocView.cs` line 170 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/858#-Orsps-U0HqJvicCDLKS:-Orsps-U0HqJvicCDLKT:b-tibmex) ([raw file](https://github.com/sillsdev/FieldWorks/blob/ff35b4a76d2fa05dabf4d6f0526b8242adddc696/Src/xWorks/XhtmlRecordDocView.cs#L170)):*
> ```Smalltalk
> 				var configuration = new DictionaryConfigurationModel(configurationFile, Cache);
> 				PublicationDecorator.Refresh();
> 				var xhtmlPath = LcmXhtmlGenerator.SavePreviewHtmlWithStyles(new [] { cmo.Hvo }, Clerk, PublicationDecorator, configuration, m_propertyTable);
> ```

Did you check and see if there are any visual changes after this beyond the current bug fix?
We explicitly passed null here before and I think we have some code that expected that for the single entry preview styling. That was probably overloading the parameter, so the fix isn't to undo this but to make that single entry view explicit.

jtmaxwell3:
I didn't see any other changes, but if there were other changes then I expect that we would want them.  Adding the publication parameter makes the function call used by the preview to be identical to the function call used by the dictionary display.  So the preview display should look like the dictionary display, which is what we want.

jasonleenaylor:
*[`Src/xWorks/XhtmlRecordDocView.cs` line 170 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/858#-Orsps-U0HqJvicCDLKS:-Os2zUMlEAc_RsyKDgy7:b-qcpt9w) ([raw file](https://github.com/sillsdev/FieldWorks/blob/ff35b4a76d2fa05dabf4d6f0526b8242adddc696/Src/xWorks/XhtmlRecordDocView.cs#L170)):*

Did you check and see if there are any visual changes after this beyond the current bug fix?
We explicitly passed null here before and I think we have some code that expected that for the single entry preview styling. That was probably overloading the parameter, so the fix isn't to undo this but to make that single entry view explicit.
</blockquote></details>



After getting past my dependency trouble I was able to build this branch locally and confirm that there are things we don't want in the single preview that were controlled by the lack of the publication decorator. Letter headers at least. I feared highlighting but I don't see that happening.

jasonleenaylor (approved):
> @jasonleenaylor reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/858)* status: all files reviewed, 1 unresolved discussion (waiting on jtmaxwell3).

PR #865, opened 2026-05-05
Closed: superseded by #908

johnml1135:
Superseded by #908, which presents the final non-render build/test/installer cleanup as a fresh PR with a clean description and review surface.

This PR is closed to avoid continuing review against the older, noisy discussion thread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/865)

### DIFF
--- a/.github/instructions/build.instructions.md
+++ b/.github/instructions/build.instructions.md
@@ -10,6 +10,11 @@ This file documents the **supported** build workflow for FieldWorks.
 
 FieldWorks is **Windows-first** and **x64-only**. Use the repo scripts so build ordering (native before managed) is correct.
 
+## Non-Windows hosts
+- Linux and macOS are supported for editing, code search, documentation, specs, and agent workflows only.
+- Do not attempt to build, test, or run installer/setup flows on non-Windows hosts.
+- `build.ps1` and `test.ps1` intentionally fail fast on non-Windows hosts so automation notices the environment mismatch.
+
 ## Quick start (PowerShell)
 ```powershell
 # Full traversal build (Debug/x64 defaults)
--- a/.github/instructions/csharp.instructions.md
+++ b/.github/instructions/csharp.instructions.md
@@ -1,0 +1,72 @@
+---
+description: 'Guidelines for building C# applications'
+applyTo: '**/*.cs'
+---
+
+# C# Development
+
+## Scope
+- FieldWorks managed code is a Windows desktop codebase built on .NET Framework 4.8 and native dependencies.
+- Treat this file as guidance for existing FieldWorks C# projects, not for ASP.NET or service-oriented .NET applications.
+- On Linux and macOS, limit work to editing, search, docs, and specs. Do not claim to have built or run managed binaries there.
+
+## Language Version And Features
+- Use the repo default C# language version unless a project explicitly overrides it. Today that default is C# 8.0 via `Directory.Build.props`.
+- Do not introduce C# features that require a newer language version without first updating repo-wide build policy.
+- Nullable reference types are disabled by default. Only use nullable annotations and nullable-flow assumptions in projects that opt in explicitly.
+- Prefer syntax that is already common in the touched area. Consistency with nearby code is more important than using the newest available syntax.
+
+## General Instructions
+- Make only high-confidence suggestions when reviewing code changes.
+- Fix the root cause when practical, but keep edits narrow and compatible with existing behavior.
+- Handle edge cases and exception paths explicitly. Do not swallow exceptions without a documented reason.
+- Treat native interop, COM, registry-free COM, file formats, and serialization boundaries as high-risk areas that need extra care.
+
+## Naming Conventions
+- Follow PascalCase for types, methods, properties, events, and public members.
+- Use camelCase for locals and private fields.
+- Prefix interfaces with `I`.
+- Keep namespaces aligned with the existing project root namespace instead of inventing new top-level naming schemes.
+
+## Formatting And Style
+- Apply the formatting rules defined in `.editorconfig` and match surrounding code.
+- Prefer block-scoped namespaces. Do not default to file-scoped namespaces in this repo.
+- Keep using directives simple and consistent with the file you are editing.
+- Use `nameof` instead of string literals when referencing member names.
+- Use pattern matching where it improves clarity and is supported by the project language version. Do not force newer syntax into older-looking code.
+
+## Documentation And Comments
+- Public APIs should have XML documentation comments.
+- Add code comments only when intent, invariants, or interop constraints are not obvious from the code itself.
+- Do not add boilerplate comments to every method.
+
+## Project Conventions
+- Most managed projects here are SDK-style `.csproj` files targeting `net48`.
+- Keep `GenerateAssemblyInfo` disabled where the project relies on linked `CommonAssemblyInfo.cs`.
+- Preserve project-specific build settings such as warnings-as-errors, x64 assumptions, WinExe/WindowsDesktop settings, and registration-free COM behavior.
+- When adding new files, update the project file only if the specific project format requires it.
+
+## Desktop, UI, And Localization
+- FieldWorks is a desktop application. Favor guidance relevant to WinForms, WPF, dialogs, view models, threading, and long-running UI work.
+- UI-affecting code must respect the UI thread. Avoid blocking calls that can freeze the application.
+- Keep user-visible strings in `.resx` resources and follow existing localization patterns. Do not hardcode new UI strings.
+- Preserve designer compatibility for WinForms and avoid edits that break generated code patterns.
+
+## Nullability And Defensive Code
+- Because nullable reference types are usually disabled, write explicit null checks at public entry points and interop boundaries when required by the surrounding code.
+- Prefer `is null` and `is not null` checks when adding new null checks.
+- Do not pretend the compiler will enforce null-state safety unless the project has opted into nullable analysis.
+
+## Testing
+- For behavior changes and bug fixes, add or update tests when practical.
+- Follow nearby test naming and structure conventions. Do not add `Arrange`, `Act`, or `Assert` comments unless the existing file already uses them.
+- Prefer fast, deterministic NUnit tests for managed code.
+- Use `./test.ps1` on Windows to run tests, and `./build.ps1` when you need a supporting build first. Do not recommend ad-hoc `dotnet test` or `msbuild` commands as the normal path for this repo.
+
+## Build And Validation
+- Use `./build.ps1` for builds and `./test.ps1` for tests in normal repo workflows.
+- Avoid changing build, packaging, COM, or registry behavior without checking the existing build instructions and affected tests.
+- Treat compiler warnings as actionable unless the repo already documents a specific exception.
+
+## What Not To Assume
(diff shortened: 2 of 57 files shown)


johnml1135:
**Copilot rationale note 1/4: why this PR exists**

This PR is intentionally the non-render half of the split from the render-speedup branch. I reviewed it as real code only and ignored the agent/skills/prompt updates.

The theme is not a single product feature. It is stabilization work around several recent FieldWorks changes that were previously riding along with the rendering branch:

- native Views/TestViews build prerequisites can exist but be stale after rebases or branch switches;
- SDK/WPF build behavior can be tripped up by stale `_wpftmp` files and missing solution metadata;
- dependency and whitespace helper scripts needed small reliability cleanups;
- ScrChecks is still a hidden runtime/test dependency even after recent project layout changes;
- installer/WiX package management has different constraints from the main app projects;
- TonePars tests need to validate generated command semantics rather than local absolute paths.

The xWorks dictionary/delete-routing changes that were previously discussed here have been removed from this PR. The underlying DeleteRecord architecture issue on main is now tracked separately as [LT-22514](https://jira.sil.org/browse/LT-22514).

So the main value of PR #865 is: keep the render PR focused, while preserving the non-render build/test/installer/TonePars fixes that make current mainline development and tests less brittle.

johnml1135:
**Copilot rationale note 2/4: build/test reliability changes**

The build/test pieces are here because FieldWorks is a mixed native/managed Windows repo, and simple existence checks are not enough for generated native artifacts.

`Build/Agent/FwBuildHelpers.psm1`, `build.ps1`, and `Build/scripts/Invoke-CppTest.ps1` add and use timestamp-based freshness checks for Views/TestViews prerequisites. This matters because files like `Views.dll`, `views.lib`, generated FwKernel headers, and Views object files can exist but still be stale relative to native sources or IDL/interface changes. After a branch switch or rebase, stale native artifacts produce confusing C++/COM build errors. This PR makes the build refresh native prerequisites when they are out of date, instead of assuming existing artifacts are valid.

`Directory.Build.props` and `Directory.Build.targets` handle two repo-wide MSBuild irritants: setting default `SolutionDir`/`SolutionName` when projects are built outside the solution, and excluding/deleting stale WPF `_wpftmp` temp projects before markup compilation. That makes SDK-style project evaluation and WPF builds less sensitive to interrupted builds and ad-hoc project invocation.

`Build/Agent/Verify-FwDependencies.ps1` wraps required/missing/optional query results as arrays so PowerShell scalar-vs-array behavior does not corrupt counts or classifications. That improves the dependency preflight script’s reliability.

`Build/Src/FwBuildTasks/CollectTargets.cs` restores ScrChecks as an explicit generated target dependency for affected tests. ScrChecks is not always discoverable as a normal project reference, but `ScrChecks.dll` still needs to land under `DistFiles/Editorial Checks`. This connects to the recent ScrChecks project-layout changes and prevents hidden runtime/test dependency loss.

johnml1135:
**Copilot rationale note 3/4: installer and TonePars changes**

The installer changes are here because FLExInstaller does not behave like an ordinary app project. WiX custom actions, DTF packages, and patch installer versioning have constraints that should not be accidentally governed by the same package-management assumptions as the main FieldWorks projects.

`FLExInstaller/Directory.Packages.props` and the WiX custom action project changes isolate installer package handling so the installer can keep the package/version behavior it needs while the rest of the repo continues moving toward shared package conventions. The patch installer workflow/targets adjustments are in the same family: make patch build/version semantics clearer and less fragile.

The TonePars change in `ToneParsInvokerTests.cs` is a deterministic-test fix. Recent work brought PCPATR/TonePars into this repo, and the tests compare generated command/log content. Those generated files can contain absolute Windows paths, which vary by developer machine and worktree. Normalizing absolute path prefixes lets the test validate the meaningful generated content rather than failing because one machine used a different checkout path.

johnml1135:
**Copilot rationale note 4/4: xWorks dictionary/headword/delete behavior**

The xWorks changes are the largest user-facing part of this cleanup. For reviewers less familiar with FieldWorks: xWorks coordinates dictionary/reversal views, record lists, filters, publication-aware decorators, and UI commands. A dictionary entry can be shown differently depending on the selected dictionary configuration, publication, and writing system. Homograph numbers distinguish entries with the same headword. Reversal indexes are dictionary-like views from the gloss/translation side.

The headword-number changes move custom homograph digits into the dictionary configuration model instead of deriving them transiently from the selected writing system each time. That is important because FLEx has multiple dictionary/reversal configurations, while the underlying homograph configuration used by views is effectively shared. Persisting custom digits per configuration makes the dialog round-trip the user’s settings and validates that custom digits are either empty or a complete set of ten.

`DictionaryPublicationDecorator.Refresh()` now detects when publication filtering or homograph numbering actually changes and notifies registered roots. This matters because dictionary views and exports are filtered through this decorator. Without notification, recalculated visible entries or numbers can leave the UI/export layer stale.

The delete-record changes make command ownership more explicit. Recent history moved delete handling through pub/sub, but delete is context-sensitive in xWorks. The wrong component handling `DeleteRecord` can mean doing nothing in the active view, or worse, deleting the wrong level of reversal data. The updated `RecordClerk`/`XmlDocView` behavior returns handled/not-handled more carefully so reversal-entry deletion can fall through to the clerk that actually owns it.

Local validation for this cleanup branch passed: build succeeded and the full test run completed with `4150 total`, `4090 passed`, and `60 skipped`. One review note: the whitespace helper simplification should be checked for UTF-8/BOM expectations, since PowerShell `Set-Content -Encoding utf8` behavior can vary between Windows PowerShell and newer PowerShell.

johnml1135:
**Copilot update:** the xWorks dictionary/DeleteRecord routing changes have been removed from this PR. The DeleteRecord pub/sub architecture issue on `main` is now tracked separately in [LT-22514](https://jira.sil.org/browse/LT-22514), with the recommended scoped command-router path and test plan. PR #865 should now be reviewed as build/test/installer/TonePars cleanup only.

jasonleenaylor (commented):
> @jasonleenaylor reviewed 23 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/865)* status: 23 of 57 files reviewed, all discussions resolved.

___
*[`Directory.Build.props` line 89 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/865#-OsbND4c8z9Bgn0v3D9o:-OsbND4c8z9Bgn0v3D9p:b-rlrrv1) ([raw file](https://github.com/sillsdev/FieldWorks/blob/daf921a47fa3c7ab1f7372b5a81a740b06c7250d/Directory.Build.props#L89)):*
> ```XML
> 		-->
> 		<SolutionDir Condition="'$(SolutionDir)' == '' Or '$(SolutionDir)' == '*Undefined*'">$(MSBuildThisFileDirectory)</SolutionDir>
> 		<SolutionName Condition="'$(SolutionName)' == '' Or '$(SolutionName)' == '*Undefined*'">FieldWorks</SolutionName>
> ```

This one makes me nervous for the future single project focused rapid debugging situations, but I don't think my nervousness is enough to hold this back. It will probably still work for us to build a single project without it trying to restore and rebuild everything. Maybe the command line build will now use a different resolution path, but it will probably be fine.
___
*[`Directory.Build.targets` line 35 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/865#-OsbNr538wcYy860y6Mw:-OsbNr538wcYy860y6Mx:bxswswq) ([raw file](https://github.com/sillsdev/FieldWorks/blob/daf921a47fa3c7ab1f7372b5a81a740b06c7250d/Directory.Build.targets#L35)):*
> ```XML
> 	-->
> 	<Target Name="_CleanStaleWpfTempProjects"
> 		BeforeTargets="MarkupCompilePass2ForMainAssembly"
> ```

Huh, that's a new one for me. Neat!
___
*[`Build/Agent/FwBuildHelpers.psm1` line 477 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/865#-OsbOsT_BbHU3q52sU_D:-OsbOsT_BbHU3q52sU_E:bonp5lh) ([raw file](https://github.com/sillsdev/FieldWorks/blob/daf921a47fa3c7ab1f7372b5a81a740b06c7250d/Build/Agent/FwBuildHelpers.psm1#L477)):*
> ```PowerShell
> }
> 
> function Get-NewestWriteTimeUtc {
> ```

I am very curious why this would ever be useful, I'll try to look out for it. Did the AI just generate it for completeness with the OldestWriteTime?
___
*[`Build/Src/FwBuildTasks/CollectTargets.cs` line 0 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/865#-OsbPMUR5_npZ2x-hYA4:-OsbPMUS6KlCYRYYcxQI:b-8n0cle) ([raw file](https://github.com/sillsdev/FieldWorks/blob/daf921a47fa3c7ab1f7372b5a81a740b06c7250d/Build/Src/FwBuildTasks/CollectTargets.cs#L0)):*
I'm not entirely sure that we need CollectTargets anymore in the new build system. Worth some investigation. Removal of it should be in a follow up commit so as not to make this one bigger and harder to review.

johnml1135:
Addressing Jason's review points:

- Directory.Build.props: I left the SolutionDir/SolutionName fallback in place. It only applies when MSBuild leaves those properties undefined, so solution/VS builds keep their existing values. The intent is to stop the nested-project restore warning without changing the normal single-project workflow.
- Directory.Build.targets: `_CleanStaleWpfTempProjects` is there to recover from interrupted WPF builds that leave `*_wpftmp*` files behind and break the next markup-compile pass.
- Build/Agent/FwBuildHelpers.psm1: `Get-NewestWriteTimeUtc` is used by `Test-ViewsNativeArtifactsStale`, which is called from both `build.ps1` and `Build/scripts/Invoke-CppTest.ps1` to decide when Views/FwKernel native artifacts need to be refreshed. So that helper is intentional rather than generated-for-completeness dead code.
- Build/Src/FwBuildTasks/CollectTargets.cs: agreed this is legacy/orthogonal to this PR. I removed the ScrChecks/CollectTargets additions in `f14d46882` to keep this PR focused.

johnml1135:
Addressing Jason's review points:

- Directory.Build.props: I left the SolutionDir/SolutionName fallback in place. It only applies when MSBuild leaves those properties undefined, so solution/VS builds keep their existing values. The intent is to stop the nested-project restore warning without changing the normal single-project workflow.
- Directory.Build.targets: `_CleanStaleWpfTempProjects` is there to recover from interrupted WPF builds that leave `*_wpftmp*` files behind and break the next markup-compile pass.
- Build/Agent/FwBuildHelpers.psm1: `Get-NewestWriteTimeUtc` is used by `Test-ViewsNativeArtifactsStale`, which is called from both `build.ps1` and `Build/scripts/Invoke-CppTest.ps1` to decide when Views/FwKernel native artifacts need to be refreshed. So that helper is intentional rather than generated-for-completeness dead code.
- Build/Src/FwBuildTasks/CollectTargets.cs: agreed this is legacy/orthogonal to this PR. I removed the ScrChecks/CollectTargets additions in `f14d46882` to keep this PR focused.

jasonleenaylor (commented):
> @jasonleenaylor reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/865)* status: 19 of 57 files reviewed, all discussions resolved.

PR #874, opened 2026-05-08
Fix more problems with LT-22357

jtmaxwell3:
This fixes more problems with https://jira.sil.org/browse/LT-22357 by widening the scope of try ... catch expressions and adding more guards.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/874)

### DIFF
--- a/Src/Common/Controls/FwControls/ProgressDialogWithTask.cs
+++ b/Src/Common/Controls/FwControls/ProgressDialogWithTask.cs
@@ -255,20 +255,20 @@ namespace SIL.FieldWorks.Common.Controls
 			{
 				CheckDisposed();
 
-				if (m_fCreatedProgressDlg && m_synchronizeInvoke.InvokeRequired)
+				try
 				{
-					try
+					if (m_fCreatedProgressDlg && m_synchronizeInvoke.InvokeRequired)
 					{
 						m_synchronizeInvoke.Invoke((Action<string>)(s => m_progressDialog.Message = s), new[] { value });
 					}
-					catch (Exception e)
-					{
-						// Fixes LT-22357.
-						Debug.Assert(false);
-					}
+					else
+						m_progressDialog.Message = value;
 				}
-				else
-					m_progressDialog.Message = value;
+				catch (Exception e)
+				{
+					// Fixes LT-22357.
+					Debug.Assert(false);
+				}
 			}
 		}
 
@@ -361,20 +361,21 @@ namespace SIL.FieldWorks.Common.Controls
 			{
 				CheckDisposed();
 
-				if (m_fCreatedProgressDlg && m_synchronizeInvoke.InvokeRequired)
+				try
 				{
-					try
+					if (m_fCreatedProgressDlg && m_synchronizeInvoke.InvokeRequired)
 					{
 						m_synchronizeInvoke.Invoke((Action<int>)(i => m_progressDialog.Minimum = i), new object[] { value });
 					}
-					catch (Exception e)
-					{
-						// Fixes LT-22357.
-						Debug.Assert(false);
-					}
+					else
+						m_progressDialog.Minimum = value;
+
 				}
-				else
-					m_progressDialog.Minimum = value;
+				catch (Exception e)
+				{
+					// Fixes LT-22357.
+					Debug.Assert(false);
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Src/Common/Controls/FwControls/ProgressDialogWithTask.cs
```diff
@@ -829,10 +830,13 @@ private void m_progressDialog_FormClosing(object sender, FormClosingEventArgs e)
 
 		private void m_worker_RunWorkerCompleted(object sender, RunWorkerCompletedEventArgs e)
 		{
-			if (m_fCreatedProgressDlg && m_synchronizeInvoke.InvokeRequired)
-				m_synchronizeInvoke.Invoke((Action)m_progressDialog.Close, null);
-			else
-				m_progressDialog.Close();
+			if (m_progressDialog != null)
```
johnml1135, line 833: drafted with the help of GPT 5.5: The new `m_progressDialog != null` guard helps, but this is still a check-then-use race: the dialog/control can be disposed after the null check and before `Invoke` / `Close`. Could this close path use the same scoped teardown-safe helper as the property setters, or at least catch the same expected teardown exceptions here too?

Src/xWorks/XhtmlDocView.cs
```diff
@@ -1527,6 +1527,10 @@ private object SaveConfiguredXhtml(IThreadedProgress progress, object[] args)
 				progress.Message = xWorksStrings.ksObtainingEntriesToDisplay;
 			var configuration = new DictionaryConfigurationModel(configurationFile, Cache);
 			publicationDecorator.Refresh();
+			if (Clerk.IsDisposed)
```
johnml1135, line 1530: drafted with the help of GPT 5.5: Could we capture `Clerk` once and use that same reference for the disposed check, `VirtualFlid`, and the later `SavePreviewHtmlWithStyles` call? As written we check `Clerk.IsDisposed`, then later re-read `Clerk.VirtualFlid`, then pass `Clerk` again, which makes the teardown lifetime harder to reason about.

Src/Common/Controls/FwControls/ProgressDialogWithTask.cs
```diff
@@ -255,20 +255,20 @@ public virtual string Message
...
 					}
-					catch (Exception e)
-					{
-						// Fixes LT-22357.
-						Debug.Assert(false);
-					}
+					else
+						m_progressDialog.Message = value;
+				}
+				catch (Exception e)
```
johnml1135, line 267: drafted with the help of GPT 5.5: Optional: would it be worth adding a focused regression test around the progress-dialog wrapper with a synchronizing control/object that tears down mid-update? Even if the full Dictionary View import flow is hard to automate, a small test here would lock in that expected teardown exceptions are ignored while unexpected ones still surface.

johnml1135, line 267: drafted with the help of GPT 5.5: Could we narrow this catch to the expected teardown exceptions instead of catching all `Exception`? The evidence here seems pretty specific: `CheckDisposed()` in this class throws `ObjectDisposedException`, `Dispose(bool)` nulls out `m_progressDialog`/`m_worker`, and the WinForms `Control.BeginInvoke` docs explicitly call out `InvalidOperationException` when no appropriate window handle can be found. The `Control.Invoke` / `InvokeRequired` docs also note exceptions if no appropriate handle exists or if the UI thread is no longer active. Unexpected exceptions here would still be useful to surface.

Src/xWorks/XhtmlDocView.cs
```diff
@@ -1527,6 +1527,10 @@ private object SaveConfiguredXhtml(IThreadedProgress progress, object[] args)
...
+			{
+				return null;
+			}
 			var entriesToPublish = publicationDecorator.GetEntriesToPublish(m_propertyTable, Clerk.VirtualFlid);
```
johnml1135, line 1534: drafted with the help of GPT 5.5: In this same method we still do `progress.Position++`, but `Position` is not covered by the new LT-22357 handling and `++` here is a cross-thread read/modify/write. Would it be cleaner to switch callers like this to `Step(1)` and then make `ProgressDialogWithTask.Step` use the same teardown-safe helper as `Message` / `Minimum` / `Maximum`? That would collapse this to one marshaled UI operation instead of a getter+setter pair.

johnml1135 (commented):
Draft review comments

Src/Common/Controls/FwControls/ProgressDialogWithTask.cs
johnml1135: It should catch 99% of the stuff - good enough.

Src/xWorks/XhtmlDocView.cs
johnml1135: Clerk doesn't change outside - it is only disposed.  Good enough.

Src/Common/Controls/FwControls/ProgressDialogWithTask.cs
johnml1135: We want to catch all exceptions.  The Debug.Assert(false) catches the issue for the programmer.

johnml1135: Unneeded.

Src/xWorks/XhtmlDocView.cs
johnml1135: It may not display properly, but it shouldn't crash.

johnml1135 (approved):
:lgtm:

> @johnml1135 reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/874#-:-OsRxhXy2QUFYw0ydasO:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on jtmaxwell3).

PR #875, opened 2026-05-11
LT 22373: Fix reversal numbers

aror92:
* Fix width of "Number even a single sense" message to prevent
the final e being cut off.
* Fix position of "number even a single sense" box to prevent
box border being cut off.
* Fix title of the configuration to say "Reversal Number
Configuration" instead of "Sense Number Configuration"
when configuring referenced senses in the reversal.
* Fix reversal number handling in GetSenseNumber
* Fix disappearing sense nums if reversals are toggled on/off

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/875)

### DIFF
--- a/Src/FwCoreDlgs/FwCoreDlgControls/ConfigSenseLayout.Designer.cs
+++ b/Src/FwCoreDlgs/FwCoreDlgControls/ConfigSenseLayout.Designer.cs
@@ -54,78 +54,78 @@ namespace SIL.FieldWorks.FwCoreDlgControls
 			this.m_chkSenseParagraphStyle = new System.Windows.Forms.CheckBox();
 			this.m_grpSenseNumber.SuspendLayout();
 			this.SuspendLayout();
-			//
+			// 
 			// m_cbNumberFont
-			//
+			// 
 			resources.ApplyResources(this.m_cbNumberFont, "m_cbNumberFont");
 			this.m_cbNumberFont.DropDownStyle = System.Windows.Forms.ComboBoxStyle.DropDownList;
 			this.m_cbNumberFont.FormattingEnabled = true;
 			this.m_cbNumberFont.Name = "m_cbNumberFont";
-			//
+			// 
 			// m_cbNumberStyle
-			//
+			// 
 			resources.ApplyResources(this.m_cbNumberStyle, "m_cbNumberStyle");
 			this.m_cbNumberStyle.DropDownStyle = System.Windows.Forms.ComboBoxStyle.DropDownList;
 			this.m_cbNumberStyle.FormattingEnabled = true;
 			this.m_cbNumberStyle.Name = "m_cbNumberStyle";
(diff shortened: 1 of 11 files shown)


jasonleenaylor:
*[`Src/xWorks/DictionaryDetailsController.cs` line 78 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/875#-OsMru_Y61jxho8qVuey:-OsMru_Y61jxho8qVuez:b9blw3z) ([raw file](https://github.com/sillsdev/FieldWorks/blob/757755f8d72af50aabbafbc20f6bc360782cfaf5/Src/xWorks/DictionaryDetailsController.cs#L78)):*
> ```Smalltalk
> 			m_configModel = model;
> 			m_node = node;
> 			bool isReversal = DictionaryConfigurationDlg.GetTopLevelNode(node).Label == "Reversal Entry";
> ```

I feel lilke we already have an IsReversal method or property somewhere that should be usable here. We shouldn't need to make GetTopLevelNode public and do logic here. If you can't find an IsReversal method that we can use here then we should make one. I would prefer to isolate any string comparison in the config into a clear method that shows why we care about it.

aror92 (commented):
> @aror92 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/875)* status: 0 of 11 files reviewed, all discussions resolved (waiting on jasonleenaylor).

___
*[`Src/xWorks/DictionaryDetailsController.cs` line 78 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/875#-OsMru_Y61jxho8qVuey:-OsR7gDz6W_1e2MNsdOz:b-q5r608) ([raw file](https://github.com/sillsdev/FieldWorks/blob/757755f8d72af50aabbafbc20f6bc360782cfaf5/Src/xWorks/DictionaryDetailsController.cs#L78)):*

I feel lilke we already have an IsReversal method or property somewhere that should be usable here. We shouldn't need to make GetTopLevelNode public and do logic here. If you can't find an IsReversal method that we can use here then we should make one. I would prefer to isolate any string comparison in the config into a clear method that shows why we care about it.
</blockquote></details>

This makes sense. You're right, there is a model.IsReversal property. Updated to use that instead. I was using GetTopLevelNode in ConfiguredLcmGenerator as well, but it's not needed there because we already have a node list and can access the root node directly. Removed that usage too, and reverted GetTopLevelNode back to private.

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 11 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/875#-:-OsS31ok4jYfPIZJPmWA:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on aror92).

PR #877, opened 2026-05-12
Fix LT-22508: text not enabled; text deletion issues

jtmaxwell3:
This is a second attempt to fix https://jira.sil.org/browse/LT-22508.  The problem was that some stale cached data was producing wrong results.  I tried to fix this in InterlinearTextsRecordClerk, but this didn't always work.    The new fix is in ConcDecorator.RemoveNotification.  I had to not only clear the local cache data (m_interestingTexts) but also the data stored in InterestingTextsDecorator.  I also had to replace m_interestingTexts with InterestingTextsList so that m_interestingTexts would be initialized at the appropriate time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/877)

### DIFF
--- a/Src/LexText/Interlinear/InterlinearTextsRecordClerk.cs
+++ b/Src/LexText/Interlinear/InterlinearTextsRecordClerk.cs
@@ -308,10 +308,6 @@ namespace SIL.FieldWorks.IText
 			{
 				m_langProjectTextsFlid = Cache.MetaDataCacheAccessor.GetFieldId("LangProject", "Texts", true);
 			}
-			if (tag == m_langProjectTextsFlid && (cvIns > 0 || cvDel > 0))
-			{
-				InterestingTextsDecorator.ClearInterestingTextsList(m_propertyTable);
-			}
 		}
 
 		internal abstract class CreateAndInsertStText : RecordList.ICreateAndInsert<IStText>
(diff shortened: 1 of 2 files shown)


jasonleenaylor (approved):
> @jasonleenaylor reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/877)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

___
*[`Src/xWorks/ConcDecorator.cs` line 564 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/877#-OsSBn6P9ScvQgB3oovl:-OsSBn6P9ScvQgB3oovm:b-pqg03l) ([raw file](https://github.com/sillsdev/FieldWorks/blob/37962f006a0090562e4d22787f4007ac1946793d/Src/xWorks/ConcDecorator.cs#L564)):*
> ```Smalltalk
> 				{
> 					m_interestingTexts = InterestingTextsDecorator.GetInterestingTextList(Mediator, PropTable, m_services);
> 					m_interestingTexts.InterestingTextsChanged += m_interestingTexts_InterestingTextsChanged;
> ```

I know this was just moved and isn't new code, but we should verify that InterestingTextsDecorator.GetInterestingTextList can never return null. If it can we might need to return the guard above, and add a guard here. If it can't then reslove this without changes.

jtmaxwell3:
InterestingTextsDecorator.GetInterestingTextList never returns null.

PR #879, opened 2026-05-12
Fix LT-22509: Newtonsoft.Json Method not found error

jasonleenaylor:
- Newtonsoft.Json doesn't change assembly versions on minor version bumps. 13.0.4 added a method and we started using it in libpalaso. If a version less than that is installed in the GAC our local version would not be loaded. This patch enforces using our local copy by our application.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/879)


johnml1135:
Initial AI review (unfiltered):


Assessment

* The WiX 6 GAC-servicing comment is the important one. The PR adds a GAC assembly component in CustomComponents.wxi:73-78, but the WiX 6 project still only shows cabinet-cache linker options in FieldWorks.Installer.wixproj:34. I did not find a visible WiX 6 equivalent of -fv / SetMsiAssemblyNameFileVersion, so Copilot’s concern there is valid.
* The bigger problem is actually cross-repo drift in the legacy path. Installer.legacy.targets:159-166 copies PatchableInstallerHeatExclude.xml into PatchableInstaller/BaseInstallerBuild/heat-exclude.xml, and AGENTS.md:13-14 says legacy buildMsi.bat passes -fv. But in the external genericinstaller repo, the actual base MSI and patch scripts do not pass -fv, and the checked-in legacy KeyPathFix.xsl files only rewrite KeyPath; they do not load heat-exclude.xml. So that copy step is currently inert unless genericinstaller changes too.
* The MASTERBUILDDIR guard pattern is sensible only if the intent is “full/base installer fix, not patch fix.” The external patch flow builds master and update from separate defines, so those guards intentionally exclude the new Newtonsoft components from the update side. If LT-22509 must be fixed via MSP, this is still a real gap.
The two XSL comments on KeyPathFix.xsl:39-85 are reasonable but non-blocking. The O(N^2) lookup and duplicated exclusion logic are maintainability/perf issues, not the main correctness risk.
* The MakeDir comment on Installer.legacy.targets:159-166 is low value. Adding it is harmless hardening, but it is not the real issue. The real issue is that the external legacy repo does not currently consume the copied file.

Recommendations

1. Keep the WiX 6 MsiAssemblyName/FileVersion comment open until there is proof. Either add the WiX 6 setting explicitly or inspect the built MSI tables and confirm the GAC assembly row includes file-version metadata.
2. Treat the legacy PatchableInstaller path as incomplete in this PR alone. Either land matching changes in the external genericinstaller repo, or remove/soften the claims in AGENTS.md:13-14 and in review discussion so the PR does not imply legacy behavior that does not exist.
3. Decide whether the supported fix path is full MSI / major upgrade only, or MSP patch too. If patch support matters, the current MASTERBUILDDIR-guarded component pattern is not enough.
4. Defer the two XSL refactors and the MakeDir tweak unless someone is already touching those files. They are follow-up cleanup, not merge blockers.

Net: Copilot’s WiX 6 servicing comment is the one to act on, the MakeDir comment is minor, and the external genericinstaller inspection exposed a larger issue Copilot did not call out: the legacy installer behavior described by this PR is not implemented in the installer repo it depends on.

I can do a manual review after this.

jasonleenaylor:
From Devin:
Based on WiX 4+ migration notes (which WiX 6 inherits): the -fv switch was removed from the linker in WiX 4. The behavior it enabled — emitting fileVersion into the MsiAssemblyName table — became the default for <File Assembly=".net"> components. So your NewtonsoftJsonGac component in FLExInstaller/wix6/Shared/Common/CustomComponents.wxi:73-79 should produce a MsiAssemblyName row with fileVersion automatically.

The generic installer repo has already been updated with those in the most recent commit (https://github.com/sillsdev/genericinstaller/commit/173675fbc98b6f4065b4cccde2c0687b5fad44d8)

I don't think this type of action (gac registration) can run on a patch in the way that we are generating them from the genericinstaller. It may even be outside of the typical patch workflow for windows installer where only file diffs are expected in an .msp (by my understanding)

I think there is a misleading wix6 comment, but otherwise this code seems to be correct to me.

johnml1135:
I think we have 3 things to do:
1. Let's update the wix6 comment to no longer be misleading
2. Let's confirm the MSI tables are made in wix6 or open a JIRA issue to confirm
3. I believe we are saying "we can't update this on a patch"  - I'm ok with that, just want to confirm.

johnml1135 (changes requested):
Please fix the issues identified.

PR #880, opened 2026-05-12
Detect stale FwBuildTasks bootstrap outputs

johnml1135:
## Summary
- rebuild `FwBuildTasks` when the bootstrapped output is older than its source or package/props inputs
- keep the freshness decision in `build.ps1` to avoid paying for a second MSBuild bootstrap invocation on the common up-to-date path
- clarify `-ForceBuildBuildTasks` help text

## Origin
The change started from build warnings in the main build when `RegFree` processing tried to load managed metadata and hit an assembly mismatch in the bootstrapped `FwBuildTasks` output.

## Root cause
`FwBuildTasks` was bootstrapped outside the main traversal build and the existing guard only checked whether `BuildTools/FwBuildTasks/<Configuration>/FwBuildTasks.dll` existed. That meant stale bootstrap outputs could survive package or build-input changes, so old dependency copies could remain in `BuildTools/FwBuildTasks/<Configuration>/` and produce warnings even though the main build itself was otherwise current.

## Solution
Add a targeted bootstrap freshness check in `build.ps1` that rebuilds `FwBuildTasks` when:
- the bootstrap output is missing
- the companion config is missing
- any `Build/Src/FwBuildTasks` source/project/targets input is newer than the bootstrapped DLL
- key repo-level props/targets/package inputs are newer than the bootstrapped DLL

This stays as a cheap script-level preflight instead of always invoking a second MSBuild bootstrap build, which keeps the common up-to-date path faster while still fixing the stale-output case.

## Validation
- `CI: Full local check`
- `./build.ps1`

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/880)

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -71,7 +71,7 @@
 	Skips running tests for native C++ components. Does not skip building these tests. Implied by SkipNative.
 
 .PARAMETER ForceBuildBuildTasks
-	Buld FwBuildTasks even if FwBuildTasks.dll already exists. BuildTasks are rarely updated, so the default is not to build them.
+	Build FwBuildTasks even if FwBuildTasks.dll already exists and is newer than its source/package inputs.
 
 .PARAMETER BuildInstaller
 	If set, builds the installer via Build/InstallerBuild.proj after the main build.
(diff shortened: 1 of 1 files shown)


johnml1135:
Addressed the review notes in `3bfa7be`.

- switched the new bootstrap path checks to `Test-Path -LiteralPath`
- replaced the recursive `Get-ChildItem -Include` scan with a single recursive file enumeration filtered by extension
- added an explicit `Build/Src/FwBuildTasks` source-root check so a missing path fails with a clear error instead of a generic enumeration failure

Validation run locally:
- `./build.ps1`
- `powershell.exe -NoProfile -ExecutionPolicy Bypass -File .\Build\Agent\commit-messages.ps1`
- `powershell.exe -NoProfile -ExecutionPolicy Bypass -File .\Build\Agent\check-and-fix-whitespace.ps1`

jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/880#-:-OsbMW_A2pqaPeqZil9A:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #896, opened 2026-05-19
Fix LT-21617: Sorting does not work for Exception Features

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-21617.  RecordList.ListAlreadySorted was implemented incorrectly, and in any case IsSorted wasn't being updated when a list was reverse sorted.  ListAlreadySorted isn't called anywhere else.  It is amazing that this code worked at all.  My guess is that IsSorted was true for most possibility lists, so the old implementation of ListAlreadySorted worked.  But for some reason it wasn't true for Exception Features, and so Exception Features couldn't be sorted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/896)

### DIFF
--- a/Src/xWorks/RecordList.cs
+++ b/Src/xWorks/RecordList.cs
@@ -209,7 +209,7 @@ namespace SIL.FieldWorks.XWorks
 		{
 			get
 			{
-				return !(m_owningObject as ICmPossibilityList).IsSorted;
+				return (m_owningObject as ICmPossibilityList).IsSorted;
 			}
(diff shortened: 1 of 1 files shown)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/896)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #911, opened 2026-05-25
Fix new case for LT-11746: Make Preview aware of publication choice

jtmaxwell3:
This fixes a new case in https://jira.sil.org/browse/LT-11746 that was reported in the comments by Kevin Warfel that is related to minor entries.  A warning message should be displayed for Lexeme-based and Hybrid-based configurations as well as Root-based when a minor entry is not to be shown.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/911)


AndyBlack (approved):
LGTM

PR #913, opened 2026-05-26
Fix LT-22530: Error message when cutting in Concordance view

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22530.  It was crashing when there wasn't an action handler.  I decided to call DeleteSelection non-undoably in this case since the user can also delete the text by typing into the selection.  Not that changing the text does anything.  If the user wants to start a new search, they need to use "Specify concordance criteria...".

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/913)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/913)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #927, opened 2026-06-03
Fix LT-22529: both allo and variant generators get wrong title

AndyBlack:
When starting both the Allomorph and Variant Generator utilties at the same time, the title of one of them had the title for the other.
The solution is to remove the formTitle variable and just use the form's actual title.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/927)

### DIFF
--- a/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenForm.cs
+++ b/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenForm.cs
@@ -2,23 +2,15 @@
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
-using Microsoft.Win32;
 using SIL.AlloGenModel;
 using SIL.AlloGenService;
-using SIL.AllomorphGenerator;
 using SIL.FieldWorks.Common.FwUtils;
 using SIL.LCModel;
-using SIL.LCModel.Infrastructure;
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
-using System.Data;
-using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
 using System.Windows.Forms;
 using XCore;
 
--- a/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs
+++ b/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs
@@ -3,32 +3,24 @@
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
 using Microsoft.Win32;
-using PtxUtils;
 using SIL.AlloGenModel;
 using SIL.AlloGenService;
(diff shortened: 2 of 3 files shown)


mark-sil (changes requested):
> @mark-sil reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/927)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on AndyBlack).

___
*[`Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs` line 6 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/927#-OuHRtHb2h13DMKUL4b8:-OuHRtHb2h13DMKUL4b9:bwhauqh) ([raw file](https://github.com/sillsdev/FieldWorks/blob/a127fd7f905bb0fa2159f4dc1dab4abcc956fddb/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs#L6)):*
> ```Smalltalk
> 
> using Microsoft.Win32;
> using PtxUtils;
> ```

Is this needed?
___
*[`Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs` line 1394 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/927#-OuHSGjH4mMTp3BRuQuV:-OuHSGjH4mMTp3BRuQuW:b42vw9u) ([raw file](https://github.com/sillsdev/FieldWorks/blob/a127fd7f905bb0fa2159f4dc1dab4abcc956fddb/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs#L1394)):*
> ```Smalltalk
> 					ActiveForm.Text += kStar;
> 			}
> 		}
> ```

Do you know why this method is using the static 'ActiveForm.Text' instead of using Text for 'this'? I think we would always be getting the correct one if we don't use the static.

When the Text already end with kStar, and there are ChangesMade, then we are removing and re-adding kStar. If this condition can exist, then we might be getting an un-necessary flicker.

Would something like this work:
protected void ShowChangeStatusOnForm()
{
    const string kStar = "*";
    bool hasStar = Text.EndsWith(kStar);
    if (ChangesMade && !hasStar)
    {
        Text += kStar;
    }
    else if (!ChangesMade && hasStar)
    {
        Text = Text.Substring(0, Text.Length - 1);
    }
}

AndyBlack:
@mark-sil Thank you, Mark.  Those were very helpful comments.

mark-sil (commented):
> @mark-sil reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/927)* status: all files reviewed, 2 unresolved discussions (waiting on AndyBlack).

mark-sil (approved):
:lgtm:

> @mark-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/927#-:-OuITkzx4lnKeDi0iEi7:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on AndyBlack).

PR #931, opened 2026-06-05
Fix LT-22539: Loss of focus from gloss field

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22539, https://jira.sil.org/browse/LT-22534, and https://jira.sil.org/browse/LT-22541.    I discovered that MoveFocusBoxIntoPlace caused the secondary cache to be cleared, which meant that any user edits were lost.  So I replaced it with UpdateField and EstablishDefaultSense.  EstablishDefaultSense is an existing function which updates the sense name and msa in the secondary cache and the display.  FocusBoxController.UpdateField suppresses the focus change to fix LT-22539.  SandboxBase.UpdateField fixes LT-22534 and LT-22541.  I put all of these fixes in the same commit to make sure that they all worked together.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/931)

### DIFF
--- a/Src/Common/FwUtils/EventConstants.cs
+++ b/Src/Common/FwUtils/EventConstants.cs
@@ -33,7 +33,6 @@ namespace SIL.FieldWorks.Common.FwUtils
 		public const string RefreshCurrentList = "RefreshCurrentList";
 		public const string RefreshInterlin = "RefreshInterlin";
 		public const string RefreshPopupWindowFonts = "RefreshPopupWindowFonts";
-		public const string RefreshSandbox = "RefreshSandbox";
 		public const string ReloadAreaTools = "ReloadAreaTools";
 		public const string RemoveFilters = "RemoveFilters";
 		public const string RestoreScrollPosition = "RestoreScrollPosition";
--- a/Src/Common/FwUtils/EventConstants.cs
+++ b/Src/Common/FwUtils/EventConstants.cs
@@ -33,6 +33,7 @@ namespace SIL.FieldWorks.Common.FwUtils
 		public const string RefreshCurrentList = "RefreshCurrentList";
 		public const string RefreshInterlin = "RefreshInterlin";
 		public const string RefreshPopupWindowFonts = "RefreshPopupWindowFonts";
+		public const string RefreshSandbox = "RefreshSandbox";
 		public const string ReloadAreaTools = "ReloadAreaTools";
 		public const string RemoveFilters = "RemoveFilters";
 		public const string RestoreScrollPosition = "RestoreScrollPosition";
--- a/Src/LexText/Interlinear/FocusBoxController.cs
+++ b/Src/LexText/Interlinear/FocusBoxController.cs
@@ -47,7 +47,7 @@ namespace SIL.FieldWorks.IText
 			get { return m_sandbox.IsDirty; }
 		}
 
-		private bool suppressFocusChange = false;
+		private bool m_suppressFocusChange = false;
 
 		// There is no logical reason for other buttons ever to get the focus. But .NET helpfully focuses the link words button
 		// as we hide the focus box. And in some other circumstance, which I can't even figure out, it focuses the menu button.
@@ -265,7 +265,7 @@ namespace SIL.FieldWorks.IText
 			}
 			finally
 			{
-				if (!suppressFocusChange)
+				if (!m_suppressFocusChange)
 				{
 					Focus();
 				}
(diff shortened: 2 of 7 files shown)


jasonleenaylor:
*[`Src/LexText/Interlinear/SandboxBase.cs` line 1667 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/931#-OuxZrSe5zhGOrVGfSQH:-OuxZrSe5zhGOrVGfSQI:b-2mhwfr) ([raw file](https://github.com/sillsdev/FieldWorks/blob/67c0eff440ec6805253336d1db79d20d8a6d0140/Src/LexText/Interlinear/SandboxBase.cs#L1667)):*
> ```Smalltalk
> 			}
> 			ICmObject hvoObject = Caches.MainCache.ServiceLocator.GetInstance<ICmObjectRepository>().GetObject(hvo);
> 			if (hvoObject is ILexSense lexSense && lexSense.Owner is ILexEntry lexEntry)
> ```

This wont' cover updates to Subsenses, which may or may not matter in practice. I think there is an 'Entry' property on ILexSense that will chase up the tree to an Entry if you think that edge case is worth covering.

johnml1135:
I defer to @jasonleenaylor for finishing the reviews on this PR.

johnml1135:
@jtmaxwell3 - don't feel you need to address this feedback - I had asked Claude about the PR and was going to review it more manually.  Here is Claude's review if you are interested:

### Concern 2 — Part-of-speech changes made directly on the MSA might not refresh

What it means: When something changes, FieldWorks broadcasts a PropChanged("object X, field Y changed"). The new code handles the part-of-speech case like this:

if (hvoObject.Owner is IMoMorphSynAnalysis msa) { ... }

That only fires when the changed object is something owned by an MSA (the grammatical-info record). But editing a part-of-speech is often a change on the MSA object itself (its PartOfSpeech is an atomic reference property). In that case the changed object is the MSA, and its owner is the LexEntry — so the is IMoMorphSynAnalysis test is false and the refresh is skipped.

Realistic concern: Moderate, and the most substantive of the three. The author manually verified LT-22541, so the specific reported scenario clearly works — meaning some edit path does flow through an MSA-owned object. The risk is the adjacent cases: a different way of editing the part-of-speech (e.g., from a different dialog or a direct reference change) could silently fail to update the Focus Box display, reproducing exactly the stale-display symptom this PR set out to kill. It wouldn't lose data or crash — the worst case is "I changed the POS and the interlinear didn't refresh until I clicked away." Worth either confirming which PropChanged patterns FieldWorks actually emits for POS edits, or just broadening the check to hvoObject is MSA || hvoObject.Owner is MSA defensively. Cheap to fix, real to leave.

---
### Concern 6 — The test stand-in (MockSandbox.UpdateField) throws instead of doing nothing

What it means: Tests use a fake sandbox, MockSandbox, in place of the real one. The new interface method was stubbed as:

public void UpdateField(int hvo, int flid) { throw new NotImplementedException(); }

Because InterlinDocForAnalysis.PropChanged now routes through FocusBox.UpdateField → m_sandbox.UpdateField, any test that drives a gloss/sense edit through this document will hit that throw.

Realistic concern: Low today, latent landmine tomorrow. I confirmed the current suite (4,251 tests, 0 failures) never triggers a PropChanged against the mock, so nothing breaks right now. But it's a tripwire: the next person who writes a perfectly reasonable test exercising focus-box refresh gets a confusing NotImplementedException from deep inside production code, not an assertion failure. The fix is one line — make it a no-op ({ }). Leaving it is harmless until it isn't, and when it bites it wastes someone's afternoon. Low severity, near-zero fix cost.

---
### Concern 7 — No regression test for the new UpdateField logic

What it means: SandboxBase.UpdateField is the heart of all three fixes (preserve focus, preserve edits, refresh sense/MSA), yet there's no automated test asserting "edit a gloss/sense → secondary cache refreshes and user edits survive." There's an established SandboxBaseTests suite where such a test would belong.

Realistic concern: Low immediate, moderate long-term. The code works now (manually verified), so there's no live defect. The risk is durability: this is a subtle, cache-interaction bug that was already regressed once. With no test pinning the behavior, a future refactor of MoveFocusBoxIntoPlace, EstablishDefaultSense, or the PropChanged routing could silently reintroduce the exact "lost edits / stolen focus" symptom — and nobody would notice until a user reports it again. This is the classic "bug fixed without a test" situation: the cost isn't paid today, it's paid the next time someone touches this area. Given the bug's history (three JIRA tickets), the case for a pinning test here is stronger than usual.

---
Bottom line: None of the three block correctness of the reported scenarios — the PR does what it claims. 2 is the only one with a plausible path to a real user-visible miss (a POS edit that doesn't refresh); 6 and 7 are test-hygiene debt that protect future-you rather than fix present behavior. If you address only one before merge, make it 2; 6 is a free one-liner worth grabbing alongside it.

PR #958, opened 2026-06-17
LT-22568: Part 2 – Scope Pub/Sub to a Main Window

mark-sil:
- Move these Subscribe calls to places where we can get the scope: ItemDataModified ConfigureCustomFields
- Added PropertyTable to VwBaseVc and TextsFilterItem so a scope could be provided in additional Publish() calls.
- Add support for subscribers in a new app window.
- Null is still valid for a scope, but require that it is provided.
- SFMImport is currently the only null (process-wide) scope. All other messages have a scope for all of the publishers and subscribers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/958)

### DIFF
--- a/Build/Installer.legacy.targets
+++ b/Build/Installer.legacy.targets
@@ -327,8 +327,12 @@
 			<!-- Installed in Program Files\Common Files -->
 			<BinFiles Remove="@(CustomInstallFiles)" />
 			<!-- Files used to customize the generic patchable installer -->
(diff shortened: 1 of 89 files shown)


aror92 (approved):
> @aror92 reviewed 68 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/958)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mark-sil).

PR #962, opened 2026-06-24
LT-22565: Add a pictures node to classified dictionary

aror92:
Add pictures node to Classified Dictionary. All info for the pictures node is the same as the pictures node in the regular Dictionary, except for the field name. We use the field name PicturesOS to directly reference the sense's pictures property, since in the Classified Dictionary pictures are children of senses instead of lexical entries.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/962)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/962)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on aror92).

PR #964, opened 2026-06-24
LT-22625: Add the WinForms to Avalonia conversion foundation

johnml1135:
## The WinForms to Avalonia conversion foundation

[LT-22625](https://jira.sil.org/browse/LT-22625)

This is the foundational change for converting FieldWorks WinForms dialogs and
views to Avalonia. It delivers three things: a documented, repeatable **process**
for performing a conversion, the **support code** that process targets, and
**example conversions** that prove both are usable.

Default behavior is unchanged. `UIMode` defaults to `Legacy` and both gate
predicates fail closed, so with no action FieldWorks does exactly what `main`
does. Open **Tools -> Options -> Interface** and set **New UI (preview)** to
*New*, and the converted views render in Avalonia, live, without a restart.

---

## Support code

The reusable machinery. Nothing in this section is specific to a converted
view or dialog, and a new conversion should add nothing here.

### View definitions to Avalonia controls

`Src/Common/FwAvalonia/` (77 files), `Src/xWorks/Avalonia/` (22 files)

- `ViewDefinition/ViewDefinitionCompiler` compiles the existing
  `Configuration/**/*.xml` Parts and Layout files into an immutable,
  fingerprinted snapshot cached by (class, layout, type, fingerprint). Migrating
  a view edits no layout XML.
- `Composer/DetailComposer` walks that snapshot against LCModel and produces a
  `DetailModel` plus an `IDetailEditContext`, applying per-project
  `ViewDefinitionOverride` files. Each node carries a `HostRouting` value
  (product / preview / unsupported), so real LCModel wiring and preview-host
  sample wiring can never be silently confused.
- `Detail/DataTree` renders the model as a `UserControl`, with the owned
  controls beside it (`FwFieldControls`, `FwOptionChooser`,
  `FwStructuredTextField`, `DetailMenuFlyout`, `DetailFocusMemory`).
- `Plugins/SlicePlugins` holds `ISlicePlugin`, keyed by the legacy layout's
  `class=` string. `SlicePluginRegistry.Register` throws on a double claim: one
  owner per legacy class, no silent overwrite.

### The gate: which framework renders a tool

`Src/Common/FwAvalonia/UIFramework.cs` and friends,
`Src/Common/FwUtils/UIModeGates.cs`

- `UIFramework { Legacy, Avalonia }` is the value `UIFrameworkResolver.Resolve`
  returns, `UIFrameworkSelectionService` wraps in its per-host decision, and
  `RecordEditView` stores to choose which of its two children to build and show.
- `LexiconFeatureCatalog` is the single list of tools that have a working
  Avalonia implementation, with the display name, description, and grouping a
  settings UI shows. `UIFrameworkRegistry` is built from it, so the gate and
  the user-visible list cannot drift apart.
- `UIFrameworkResolver.Resolve` answers, for one tool, whether Legacy WinForms
  or Avalonia renders it. Precedence: an unregistered tool is always Legacy;
  otherwise an explicit override wins (tests use this); otherwise the persisted
  `UIMode` decides. Only a case-insensitive "New" selects Avalonia, so null,
  blank, or an unrecognized value falls back to Legacy. The type has no
  Avalonia dependency, so the decision is unit-testable without a UI runtime.
- `UIFrameworkSelectionService.Decide` classifies each host's behavior under
  the global switch (supported / explicit legacy fallback), and per-tool
  opt-outs (`UIModeDisabledTools`) are honored by the hosting layer, so New
  mode is not all-or-nothing.
- `UIMode` is a PropertyTable string defaulting to `Legacy`. Call sites branch
  through a `[MethodImpl(MethodImplOptions.NoInlining)]` helper so the Avalonia
  assemblies are never loaded on the legacy path.

### Dialog hosting

`Src/Common/FwAvaloniaDialogs/` (57 files),
`Src/Common/FwAvalonia/AvaloniaDialogHost.cs`

- `AvaloniaDialogHost.ShowModal` hosts an Avalonia `UserControl` inside a
  WinForms-owned modal `Form`, which is what makes coexistence possible: a
  converted dialog can be launched from unconverted WinForms code.
- `AvaloniaDialogLauncher<TState, TViewModel, TPayload>` is the template method
  every conversion fills in (build state, create view model, create view,
  apply). It collapses a non-accepted result to cancelled in one place.
- `DialogViewModelBase` supplies `RequestClose`, `Accepted`, `CanOk`,
  `ApplyChanges`, and the generated OK/Cancel commands.
- `DialogTheme` and the shared sizing tokens keep converted dialogs visually
  consistent without per-dialog styling.
- Strings live in `FwAvaloniaDialogsStrings.resx`. Localization stays in resx.

### Preview host

`Src/Common/FwAvaloniaPreviewHost/` (7 files)

Renders a converted dialog or view standalone, so layout work does not
require launching FieldWorks and opening a project.

---

## Example conversions

Proof that the support code and the process work. These are the deliverable's
evidence, not its purpose.

### Detail views (4 tools)

Enrolled in `LexiconFeatureCatalog`: `lexiconEdit`, `lexiconEditPopup`,
`notebookEdit`, `posEdit`. Each renders through the compiler and composer
above, from the unmodified layout XML.

### Dialog bodies (9)

`AddNewSenseDlg`, `ChooserDialog`, `CreateFeatureDialog`, `EntryGoDialog`,
`FeatureChooserDialog`, `InsertEntryDlg`, `LexOptionsDlg`, `MessageBox`,
`MsaCreatorDlg`.

### Product wiring (14 launchers)

`Src/LexText/**/Avalonia/` (39 files). One launcher per legacy dialog entry
point, each selected at its call site by the resolved framework.

The reuse pattern matters more than the count: `EntryGoDialog` is one dialog body
re-skinned by six launchers (`LcmGoToEntryDialogLauncher`,
`LcmLinkEntryOrSenseDialogLauncher`, `LcmLinkAllomorphDialogLauncher`,
`LcmLinkMsaDialogLauncher`, `LcmMergeEntryDialogLauncher`,
`LcmAddAllomorphDialogLauncher`). Only the title, prompts, search filter, and
on-OK behavior differ. Three members of that legacy family are not yet
converted (`ReversalEntryGoDlg`, `RecordGoDlg`, `WordformGoDlg`); each needs a
launcher, not a view.

---

## The process

This is what makes the change foundational rather than four screens and some
dialogs.

### Agent procedure (15 skills added, 3 updated)

`.claude/skills/` (37 files). `convert-dialog` and `convert-slice` drive one
conversion through analysis, developer alignment, test planning, exemplar
mapping, design, and implementation. The middle stages are gates: an interactive
question that ends the turn, not a paragraph to read past. Analysis is read-only,
so it is safe to run while a developer has the live application open.

Supporting skills cover the areas a conversion routinely gets wrong:
`dialog-update` (keeping a converted pair in sync), `create-integration-test`,
`fieldworks-code-commenting` (the C# comment standard, referenced from
`AGENTS.md`), `fieldworks-localization-review`,
`fieldworks-semantic-render-parity`, `fieldworks-ui-wiring-review`,
`fieldworks-uia2-parity-testing`, `fieldworks-managed-netfx-review`,
`fieldworks-migration-scope-review`, `fieldworks-test-coverage`.

### Developer guides

`Docs/migration/` (4 files): `avalonia-migration-overview.md`,
`migrate-a-dialog.md`, `migrate-a-slice-type.md`, and `adjust-the-layout.md` --
the hand-tuning path for a developer who wants to adjust a converted layout in
Visual Studio rather than accept what the conversion produced.

### Shared language

`CONTEXT.md` carries the project glossary the conversion vocabulary is
grounded in. Terms that accumulated conflicting senses during the work were
renamed rather than documented around: the framework choice is `UIFramework`,
product-vs-preview wiring is `HostRouting`, and the glossary records the
retirement of the words they replaced.

### Durable contracts and lessons

`openspec/specs/` (41 files) holds the contracts a future change must not break.
`Docs/lessons/avalonia-migration/` (7 files) is an indexed set of lesson cards
recording reusable observations and decision boundaries. The cards constrain
future research; they do not authorize restoring retired code.

---

## Unfinished work is visible, not absent

Two mechanisms, with different scopes. Inside a converted detail view, a slice
the Avalonia path cannot yet render composes as a labeled `Unsupported` row
with a stable automation id, never a silent omission -- grep those rows and you
have the remaining slice work. This is why the enrolled tools can ship
before every field kind exists. Everything else that is unconverted -- dialogs
without a launcher, tools not in the registry -- shows nothing new at all: it
stays entirely on the WinForms path via the gate, so there is no half-rendered
state anywhere.

---

## Tests

139 new test files, 1,590 test methods, +34,895 lines. Headless Avalonia tests
for views and view models, composer tests against a real LCModel cache, gate and
fallback tests, automation-id convention tests, and semantic snapshot comparisons
against the WinForms views.

---

## Scope

476 files, +82,359 / -470, against `main`. Tests are counted separately from the
source rows below, not within them.

| Area | Files | Added |
| --- | --- | --- |
| Tests (all areas) | 149 | 34,895 |
| `Src/Common/FwAvalonia` (core) | 77 | 15,715 |
| `Src/Common/FwAvaloniaDialogs` | 57 | 8,372 |
| `Src/xWorks` (hosting, composer) | 22 | 6,712 |
| `Src/LexText` (launchers, gates) | 39 | 6,362 |
| `.claude/skills` (process) | 37 | 4,788 |
| Build and solution configuration | 11 | 2,129 |
| `openspec/specs` | 41 | 1,111 |
| `Docs/migration` | 4 | 587 |
| `Src` (gate call sites elsewhere) | 22 | 558 |
| `Docs/lessons` | 9 | 547 |
| `Src/Common/FwAvaloniaPreviewHost` | 7 | 508 |
| `Docs` (other) | 1 | 75 |

The 470 deletions are gate call sites and comment cleanup. No legacy WinForms
view or dialog is removed by this change.

---

## Verification

CI on this head is the source of truth. Locally: whole-solution build clean, and
the Avalonia and composer test projects green.

To exercise it by hand, set **New UI (preview)** to *New* and confirm the four
tools render and the converted dialogs open; then set it back to *Legacy* and
confirm the WinForms path is untouched.

---

## What is deleted when the conversion is complete

**Avalonia-in-WinForms hosting.** Exists only to put Avalonia content inside a
WinForms parent.

`AvaloniaDialogHost.cs`, `AvaloniaHostControlBase.cs`, `DetailHostControl.cs`,
`InputKeyClaimingAvaloniaHost.cs`, `FinalizerSafeSynchronizationContext.cs`,
`FwAvaloniaRuntime.cs`, `Seams/IXCoreCommandBridge.cs`,
`Src/xWorks/Avalonia/Hosting/RecordEditView.Avalonia.cs`

**Framework selection.** Chooses Legacy or Avalonia per tool; one framework
leaves nothing to choose.

`UIFramework.cs`, `UIFrameworkResolver.cs`, `UIFrameworkRegistry.cs`,
`UIFrameworkSelectionService.cs`, `LexiconFeatureCatalog.cs`,
`EditControlFactory.cs`, `Src/Common/FwUtils/UIModeGates.cs`, the `UIMode`
setting and its Options-dialog editor, and every `NoInlining` call site

**Legacy managed UI.** Every WinForms dialog and control, plus the record-edit
stack behind them, replaced by the Avalonia views and dialog bodies.

163 `Form` subclasses, 84 `UserControl` subclasses, 236 `.Designer.cs` files and
their `.resx` companions, plus `Src/Common/Controls/DetailControls` (88 files,
`DataTree` and the `Slice` hierarchy), `Src/Common/Controls/XMLViews` (66),
`Src/Common/RootSite` (44), `Src/Common/SimpleRootSite` (42), and `Src/XCore`
(96, the mediator and `XWindow`)

**Native Views library.** The C++ COM engine that draws every legacy field;
Avalonia renders and shapes its own text, which
`lexical-edit-font-decommissioning` already requires.

`Src/views` entire -- 136 C++ files (`VwRootBox`, `VwEnv`, `VwGraphics`, the
lazy-box layout, `lib/GraphiteEngine`, `lib/UniscribeEngine`), the `Views.idh`
COM interfaces, and the STA constraint the render path imposes on the process

**`FieldWorks.Main` changes rather than goes.** Its
`Application.EnableVisualStyles()` and `Application.Run()` become
`BuildAvaloniaApp().StartWithClassicDesktopLifetime(args)` -- the call the preview
host already makes -- so Avalonia owns the message loop. The
`Application.ThreadException` hook becomes a dispatcher handler and
`Application.DoEvents` moves to `Dispatcher.UIThread`.

Runtime XML layout parsing retires separately, view by view as parity gates pass,
leaving the typed view definitions as the runtime source
(`lexical-edit-view-definition`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/964)


johnml1135:
@C:/Users/johnm/AppData/Local/Temp/claude/C--Users-johnm-Documents-repos-FieldWorks-worktrees-010-advanced-entry-view/2b14e829-7fba-4b8e-9d29-78ed6e0d2c6c/scratchpad/pr964-audit-final.md

Src/xWorks/RecordEditView.cs
```diff
@@ -356,6 +460,82 @@ private bool ShouldSuppressFocusChange(RecordNavigationInfo rni)
...
+			m_dataEntryForm.BringToFront();
+		}
+
+		private void AttachLegacySurfaceToPanel()
+		{
+			if (m_dataEntryForm == null || m_panel == null)
+				return;
+
+			if (!m_panel.Controls.Contains(m_dataEntryForm))
+				m_panel.Controls.Add(m_dataEntryForm);
```
johnml1135, line 525: Disclosure (low severity, no behavior change): m_dataEntryForm now stays nested inside m_panel rather than being reparented as a sibling. This adds one level to the UI-Automation tree for Legacy users. Visually identical (Dock=Fill either way) — noted for the record.

Src/xWorks/RecordEditView.cs
```diff
@@ -356,6 +460,82 @@ private bool ShouldSuppressFocusChange(RecordNavigationInfo rni)
...
+			get
+			{
+				var persistContext = XmlUtils.GetOptionalAttributeValue(m_configurationParameters, "persistContext");
+				return string.IsNullOrEmpty(persistContext)
+					? m_vectorName + ".DataTree"
+					: m_vectorName + "." + persistContext + ".DataTree";
+			}
+		}
+
+		private void EnsureLegacySurfaceInitialized()
```
johnml1135: Disclosure (low severity, no behavior change): legacy DataTree setup moved from Init() to lazy first-record-show via EnsureLegacySurfaceInitialized(). Same code, later timing — the surface is now built when the first record is shown rather than during Init().

Src/Common/FieldWorks/App.config
```diff
@@ -29,6 +29,17 @@ Only entries that auto-gen cannot produce are kept below.
...
 </dependentAssembly>
+<!-- FwAvalonia brings System.Diagnostics.DiagnosticSource (CPM-pinned in
+  Directory.Packages.props) via a project reference. FieldWorks.exe does not reference
+  that package directly, so auto-generated redirects do not emit this entry and net48
+  fails to bind Avalonia's 8.0.0.0 reference to the copied assembly at runtime.
+  newVersion must match the assembly version of the deployed DLL (9.0.2 package =
+  assembly 9.0.0.2). Keep this until the host resolves DiagnosticSource directly
+  through the main package graph. -->
+<dependentAssembly>
+<assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
```
johnml1135, line 40: Disclosure (low severity): new process-wide binding redirect for System.Diagnostics.DiagnosticSource, added because FwAvalonia pulls it in. Versions are CPM-pinned solution-wide so there is no conflict today; noting it because it affects assembly resolution for everyone, Legacy included.

PR #970, opened 2026-06-25
Fix LT-22586: Normalize Word export content to NFC

jasonleenaylor:
* Normalize the OpenXML <w:t> run text in place (new DocFragment.NormalizeText) and the letter-header string
* Add regression tests GenerateWordDocForEntry_GeneratesNFC and GenerateLetterHeaderDocFragment_GeneratesNFC

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/970)

### DIFF
--- a/Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs
+++ b/Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs
@@ -18,6 +18,7 @@ using SIL.FieldWorks.Common.Widgets;
 using SIL.LCModel;
 using SIL.LCModel.Core.KernelInterfaces;
 using SIL.LCModel.Core.Text;
+using SIL.LCModel.Core.WritingSystems;
 using SIL.LCModel.DomainServices;
 using SIL.WritingSystems;
 using SIL.TestUtilities;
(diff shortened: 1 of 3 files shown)


Src/xWorks/xWorksTests/LcmWordGeneratorTests.cs
```diff
@@ -599,6 +600,65 @@ public void GenerateWordDocForEntry_LineBreaksInBeforeContentWork()
...
+			var nfdHeadword = CustomIcu.GetIcuNormalizer(FwNormalizationMode.knmNFD).Normalize("자ㄱㄴ시"); // Korean
+			entry.CitationForm.set_String(wsKo, TsStringUtils.MakeString(nfdHeadword, wsKo));
+			var storedHeadword = entry.CitationForm.get_String(wsKo);
+			Assert.That(storedHeadword.get_IsNormalizedForm(FwNormalizationMode.knmNFD), "Should be NFDecomposed in memory");
+			Assert.That(storedHeadword.Text.Length, Is.EqualTo(6), "NFD headword should have 6 codepoints");
+
+			//SUT
+			var result = ConfiguredLcmGenerator.GenerateContentForEntry(entry, mainEntryNode, null, DefaultSettings, 0) as DocFragment;
+			Assert.That(result, Is.Not.Null, "Results should have been generated");
+			var tsResult = TsStringUtils.MakeString(result.ToString(), wsKo);
```
hahn-kev: is there a reason we convert it into a ts string before testing?

jasonleenaylor: No, not really. It has a symmetry with the code setting the string in the model but it doesn't add any value. It is correct code with an unnecessary wrapper. I'll simplify it since you asked and made me think about it. :)

PR #975, opened 2026-06-29
Fix LT-22589: Crash on clicking Item button in custom list

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22589.  A PropChanged was called when the Clerk was disposed.  This shouldn't happen.  I fixed it so that at least the PropChanged doesn't crash.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/975)

### DIFF
--- a/Src/xWorks/XWorksViewBase.cs
+++ b/Src/xWorks/XWorksViewBase.cs
@@ -233,6 +233,11 @@ namespace SIL.FieldWorks.XWorks
 			}
 		}
 
+		internal bool ClerkIsDisposed()
+		{
+			return m_clerk?.IsDisposed == true;
+		}
+
 		public IPaneBar MainPaneBar
 		{
 			get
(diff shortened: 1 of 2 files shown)


AndyBlack (approved):

jasonleenaylor:
*[`Src/xWorks/XhtmlRecordDocView.cs` line 191 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/975#-OwJCzwa7zWntXZhJLHk:-OwJCzwa7zWntXZhJLHl:bovooe6) ([raw file](https://github.com/sillsdev/FieldWorks/blob/e10fdf903c9d545d319befcd7a92b2c477fe30af/Src/xWorks/XhtmlRecordDocView.cs#L191)):*
> ```Smalltalk
> 		public void PropChanged(int hvo, int tag, int ivMin, int cvIns, int cvDel)
> 		{
> 			if (ClerkIsDisposed())
> ```

Combining this test with the if below would be fine, the commit comment will retain the purpose so losing the comment is ok. Not a big deal though.



_Code quote:_
```Smalltalk
			if (ClerkIsDisposed())
```

PR #978, opened 2026-07-01
chore: fix VersionInfoProvider bugs, drop stale RegFree.targets entry, thin opsx prompts

johnml1135:
This branch pulls three unrelated fixes out of #964 (the Avalonia migration
branch), where they had been bundled in by mistake: a set of
`VersionInfoProvider` bugs (frozen copyright year, wrong assembly
resolution, a placeholder leak into the About box, and a `FormatException`
out of a property getter), a stale build-file reference to
a retired COM class, and 11 `.github/prompts/opsx-*.prompt.md` files
rewritten to delegate to their matching `.claude/skills/openspec-*` skill
instead of repeating the same instructions inline.

The obvious question for a three-in-one bundle like this is why it isn't
three PRs -- it's the size of the diff that argues against splitting further,
not the relatedness of the changes: each piece is small, self-contained, and
independently testable, and none of them are Avalonia-migration-specific, so
keeping #964 scoped to the migration mattered more than keeping this cleanup
maximally atomic. This has already been through one review round: a prior
pass flagged that the retired-COM-class cleanup only covered
`RegFree.targets` and missed `BuildInclude.targets`/`mkall.targets`; that gap
is now closed (verified below), and #989, the PR that would have collided
with the `RegFree.targets` line, was closed as a duplicate.

**Where to look**

- `VersionInfoProvider.cs` -- the bug fixes, pinned by 6 new tests in
  `VersionInfoProviderTests.cs` that link the real generated
  `CommonAssemblyInfo.cs`, so they exercise the shipped attribute shapes
  rather than a fake.
  - `MajorVersion` previously printed the literal placeholder
    `MAJOR.MINOR STABILITY` (and a trailing space) when the informational
    version had fewer parts than expected. It could not throw -- the
    placeholder array was always length 5 -- so this is a display bug, not
    a robustness fix.
  - `ParseInformationalVersion` separately *could* throw:
    `Convert.ToInt32(versionParts[1])` raises `FormatException` on a
    non-numeric second token, out of a property getter the splash screen
    and About box both read. Now `int.TryParse`, matching the idiom
    already used by `BaseBuildNumber` in the same file.
- `Build/mkall.targets` -- the one remaining stale `ManagedVwWindow` CLSID
  exclusion. `main` still carries this entry at `Build/mkall.targets:46`;
  after this branch, `ManagedVwWindow` appears in no `*.targets` file.
- `.github/prompts/opsx-*.prompt.md` (11 files) -- each now delegates to its
  matching `.claude/skills/openspec-*/SKILL.md`; every target skill
  directory was confirmed to exist.

**Deliberately not here**

- `kDefaultCopyrightString`/`kSensitiveCopyrightString` moved from `const`
  to `static readonly` (needed to compute the year at runtime) is a
  binary-compat change for any out-of-repo consumer that inlined the old
  `const` -- confirmed no in-repo caller does.
- The sensitive-mode copyright strip
  (`.Replace("SIL International", string.Empty)`) is literal, not
  pattern-based: fine against today's fixed template, but would leak a
  future suffix (e.g. a city name) into "sensitive" output. Flagged in
  review, left as a documented trade-off rather than churning the diff
  further.

Extracted from #964, not stacked on it. Rebased onto `main` today (58
commits behind). `./build.ps1` and
`./test.ps1 -TestProject "Src/Common/FwUtils/FwUtilsTests" -TestFilter "VersionInfoProviderTests"`
both clean on the rebased tree: `Test Run Successful. Total tests: 6`.

---

<details>

This PR went through a real review round before this rebase: a prior pass
found a MEDIUM issue (the `ManagedVwWindow.dll` cleanup was incomplete) and
two LOW nits, and a follow-up commit fixed the MEDIUM one. That follow-up
commit's SHA (and the original bug-finding commit's SHA) are cited in the
PR's older comments but are no longer reachable history after this rebase
-- trust the grep in "Where to look" above over the old SHAs, which is why
this description states the fix as a verified-in-tree fact rather than a
commit reference.

</details>

<details>

**Why `RegFree.targets` shows no diff against `main`.** The original version
of this branch removed a stale `ManagedVwWindow.dll` entry from
`Build/RegFree.targets`. A separate PR (#989) touched the same line
independently; #989 was closed as a duplicate once this branch's fix was
confirmed to cover the same ground, and by the time this branch was rebased
onto the current `main`, that line's content already matched -- so the
rebase folded it to a no-op automatically. Nothing was lost; there is
simply nothing left to show.

**Why the MEDIUM issue needed a second commit.** The first pass only edited
`RegFree.targets`, whose `ManagedVwWindow.dll` fallback entry is gated by
`Condition="'@(ManagedComAssemblies)' == ''"`. `FieldWorks.exe` never hit
that fallback: `Src/Common/FieldWorks/BuildInclude.targets` populated
`ManagedComAssemblies` explicitly, with its own `ManagedVwWindow.dll` entry,
so the main executable kept emitting a "Could not find file" warning
regardless of the `RegFree.targets` edit. The follow-up removed that
`BuildInclude.targets` entry and the matching stale CLSID exclusion in
`Build/mkall.targets`.

**Why the two LOW nits were left alone.** Both are pre-existing-pattern
trade-offs rather than bugs introduced by this branch: the `const` ->
`static readonly` change is required to compute the copyright year at
runtime, and a repo-wide grep confirmed no consumer depends on it being a
compile-time constant. The literal-string copyright strip matches the
current fixed template exactly; making it pattern-based would be defending
against a template change that has not happened, at the cost of a more
fragile regex today.

</details>

<details>

- `./build.ps1` succeeds on the rebased branch.
- `./test.ps1 -TestProject "Src/Common/FwUtils/FwUtilsTests" -TestFilter
  "VersionInfoProviderTests"` -- `Test Run Successful. Total tests: 6`,
  covering the copyright-year, assembly-resolution, and trailing-whitespace
  fixes.
- `git grep ManagedVwWindow origin/main -- '*.targets'` returns
  `Build/mkall.targets:46`; the same grep on this branch returns nothing.
  The deletion is doing real work.
- Every `.github/prompts/opsx-*.prompt.md` file's target skill name resolves
  to an existing `.claude/skills/openspec-*/` directory (checked all 11).

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/978)


jasonleenaylor (commented):
Looks good, Devin noted that ManagedVwWindow.dll was removed from RegFree.targets defaults but is still present in FieldWorks.exe BuildInclude.targets

> @jasonleenaylor reviewed 14 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/978#-:-OwZ_t6J-dCOkk-qVJf7:b-nj3kv9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

johnml1135:
## Code review (automated deep review, verified against the code)

Overall: **the core VersionInfoProvider bug fixes are correct and well-tested** — verified each against the real generated attribute shapes (`ApplicationVersion` now reads `m_assembly` rather than always the entry assembly; the trailing-space leak from the empty `$FWBETAVERSION` case is fixed by the `IsNullOrWhiteSpace` guard; `MajorVersion` no longer emits `"9.3 STABILITY"`/`"9.3 "`; copyright years are computed at runtime). The new `VersionInfoProviderTests` link the real `CommonAssemblyInfo.cs`, so they exercise the shipped attribute shapes. The 11 opsx stubs delegate to skill names that all map to existing `.claude/skills/openspec-*/` dirs. Nice.

A few points:

### MEDIUM — the RegFree cleanup is only half-done, and overlaps PR #989
This PR removes `ManagedVwWindow.dll` from `Build/RegFree.targets`, but that entry sits in the fallback `ItemGroup` gated by `Condition="'@(ManagedComAssemblies)' == ''"`. For **FieldWorks.exe** the list is populated explicitly in `Src/Common/FieldWorks/BuildInclude.targets:10` (which still contains `<ManagedComAssemblies Include="$(OutDir)ManagedVwWindow.dll" />`), so the fallback never applied to the main executable — it will keep emitting the `Could not find file ...ManagedVwWindow.dll` warning. The `retire-linux-era-view-shims` change (task 5.1, marked done) names **both** files; only `RegFree.targets` was done. Suggest also removing `BuildInclude.targets:10` (and the stale CLSID-exclusion comment at `Build/mkall.targets:46`).

Also note: **this RegFree.targets edit is the same line PR #989 removes** (there it's replaced with an explanatory comment). The two PRs will conflict / one becomes a no-op — worth deciding which one carries the RegFree change and dropping it from the other. The removal *is* genuinely useful for LCMBrowser/GenerateHCConfig/UnicodeCharEditor/ComManifestTestHost, which hit the fallback path.

### LOW / nit — `const` → `static readonly` is a (minor) binary-compat change
`VersionInfoProvider.cs:23-28`: `kDefaultCopyrightString`/`kSensitiveCopyrightString` are `public` and moved from `const` to `static readonly`. Necessary to compute the year at runtime, and I confirmed the only in-repo references are within `VersionInfoProvider` + the new test (no const-context/attribute/`case` usages), so it's safe here. Flagging only for public-API hygiene — an out-of-repo consumer that inlined the old const would need a recompile.

### LOW — sensitive-mode copyright strip is literal
`VersionInfoProvider.cs:305`: `copyRight.Replace("SIL International", string.Empty).TrimEnd()`. Strips cleanly against today's fixed template (`Copyright (c) 2002-$YEAR SIL International`), but brittle if that string ever gains a trailing suffix (e.g. `, Dallas`), which would leak into "sensitive" output. Future fragility only.

### Process note
The test-plan checkboxes (`build.ps1`, `test.ps1 FwUtilsTests`) are still unchecked, so passing tests aren't yet evidenced in the description.

johnml1135:
Fixed the MEDIUM issue in 5b964370d: removed the remaining `ManagedVwWindow.dll` entry from `BuildInclude.targets` (which pre-populates `ManagedComAssemblies` for `FieldWorks.exe`, bypassing the `RegFree.targets` fallback) and the matching stale CLSID exclusion in `mkall.targets`. The RegFree.targets overlap with #989 is moot now — #989 was closed as a duplicate.

Leaving the two LOW nits (const→static readonly, literal copyright strip) as documented trade-offs rather than churning the diff further; flag if you'd like those addressed too.

jasonleenaylor (changes requested):
The three `VersionInfoProvider` fixes are right and the tests pin them against the
real generated attributes rather than a fake, which is the correct instinct. Three
things to clear before this merges.

**1. Drop the `AGENTS.md` hunk.**

`main` already carries an equivalent bullet at `AGENTS.md:15` ("Commit messages must
pass `gitlint` ... See `.claude/skills/commit-messages/SKILL.md`"), and the context
line this hunk anchors on has since become `- Build with `.\build.ps1 -CommentHygiene``.
As it stands the hunk is a guaranteed conflict and redundant once resolved.

**2. Comment hygiene.**

Three added `//` blocks exceed the 200-character cap in
`.claude/skills/fieldworks-code-commenting/SKILL.md`, and none of the regions reach
the McCabe 11+ threshold that would earn the 600-character exemption:

- `VersionInfoProvider.cs` `MajorVersion` — 241 chars
- `VersionInfoProvider.cs` `ApplicationVersion` — 204 chars
- `VersionInfoProvider.cs` `CopyrightString` — 213 chars

Four more trip banned content categories:

- `kDefaultCopyrightString` doc: "Computed so it can never ship frozen at the year
  somebody last edited this file" — who last edited the file is repo history, not the
  member's contract. The contract is "current year".
- `VersionInfoProviderTests.cs`: "they were once hardcoded \"2002-2021\" and shipped
  stale for years" — absence narration.
- `CopyrightString`: "in EITHER mode" only parses against the old code, and "we don't
  have to update the splash screen each year" names a caller. The second half is
  inherited text, but this block is being rewritten anyway.
- `ApplicationVersion_ReportsTheProvidersAssembly_NotTheEntryAssembly`: the trailing
  "must report THAT assembly's version" restates the method name. The first half
  ("under a test runner the entry assembly is testhost") is the part worth keeping.

For contrast, the `FWBETAVERSION` comment in `ParseInformationalVersion` is the model:
147 characters, states a fact the code cannot show, survives reimplementation.

`build.ps1 -CommentHygiene` will fail on the three over-cap blocks.

**3. Re-read the PR body against what the diff still does.**

The body will land in history, and two claims in it no longer hold:

- The Evidence section says grepping for `ManagedVwWindow` "returns nothing, in any
  `*.targets` file, on this branch or on `main`." On `main` at `17cdfba35`,
  `git grep ManagedVwWindow` returns `Build/mkall.targets:46`. True of the branch,
  false of `main` — which matters in your favour, since it means that deletion is
  still doing real work rather than being the no-op the sentence implies.

- The defensive-parsing framing overstates the diff. `MajorVersion` could not throw
  before: the placeholder array was always length 5, so the old failure mode was
  printing literal `MAJOR.MINOR STABILITY` in the About box, not a crash. Meanwhile
  the one line that can actually throw is untouched —
  `ParseInformationalVersion`: `int date = Convert.ToInt32(versionParts[1]);`. An
  informational version whose second token is not an integer throws `FormatException`
  out of a property getter the splash screen and About box call, one case label away
  from the whitespace guard you did add. Either switch it to `int.TryParse` or narrow
  the body's wording so the claim matches the change — your call which.

johnml1135:
All three points taken, and the third one changed the diff rather than just the
prose. Rebased onto `main` (was 34 behind); the review-response changes are in a
single commit on top so you can diff just that.

**1. `AGENTS.md` hunk -- dropped.**

You were right that it was both redundant and the conflict. `main`'s bullet is a
strict superset of mine (it adds the workflow path, the blank-line rule, the
heredoc note, and the skill pointer), and my hunk anchored on
`- Build with `.\build.ps1`.`, which is now
`- Build with `.\build.ps1 -CommentHygiene``. The commit that added it is gone
from the branch, and the follow-up ASCII-punctuation commit that touched the same
lines was resolved to `main`'s file. `AGENTS.md` no longer appears in the diff and
the PR is no longer CONFLICTING.

**2. Comment hygiene -- fixed, and the gate now passes.**

I measured the three you flagged and got 241 / 214 / 204, matching your numbers.
All three are rewritten, along with the four banned-content ones:

- `MajorVersion` 241 -> 151. Drops "index defensively" (see point 3).
- `CopyrightString` 214 -> 73. Drops "in EITHER mode" and the splash-screen
  caller reference.
- `ApplicationVersion` 204 -> 80. Drops the `InternalProductVersion` collaborator
  naming.
- `kDefaultCopyrightString` doc: "somebody last edited this file" was repo history,
  as you said. Now states the contract: a fallback, through the current year, for
  an assembly with no copyright attribute.
- Test: "they were once hardcoded 2002-2021 and shipped stale for years" deleted
  outright -- absence narration with nothing left over once removed.
- Test: the trailing "must report THAT assembly's version" restated the method
  name; kept only the testhost half, which is the part that isn't obvious.

`build.ps1 -CommentHygiene` reports `comment-hygiene: clean` on the branch. It
also auto-rewrapped two of my rewritten lines, which are included in the commit.

**3. Body claims -- both corrected, and you were right about which way the
`ManagedVwWindow` one cuts.**

`git grep ManagedVwWindow origin/main -- '*.targets'` does return
`Build/mkall.targets:46`. The body now says exactly that, and says the same grep
on this branch returns nothing, so the deletion is doing real work instead of
reading as a no-op. Thanks -- that sentence was undercutting my own change.

On the parsing framing: you're right that old `MajorVersion` could not throw. The
placeholder array was always length 5, so `versionParts[4]` was always in range,
and the failure mode was printing the literal `MAJOR.MINOR STABILITY` into the
About box. So "non-defensive parsing" was the wrong noun for it -- it's a
placeholder leak, and the body now calls it that.

You offered a choice on `Convert.ToInt32(versionParts[1])` and I took the fix
rather than the reword. It's a real `FormatException` out of a property getter the
splash screen and About box both read, it's one case label away from the guard I
did add, and `BaseBuildNumber` seventy lines up in the same file already uses
`int.TryParse` -- so the guard is the local idiom, not a new pattern. Now:

```cs
// A non-numeric token here would otherwise throw out of a property getter
// that the splash screen and About box read.
int.TryParse(versionParts[1], out var date);
if (date > 0)
```

`date` stays 0 on a parse failure, which the existing `if (date > 0)` already
treats as "no date", so the fallback path is unchanged. The body now lists this as
a fourth fix rather than folding it into the third.

Src/Common/FwUtils/VersionInfoProvider.cs
```diff
@@ -219,12 +225,17 @@ private static void ParseInformationalVersion(Assembly assembly, out string prod
...
+					// informational version; don't let it leak into the parsed display strings.
+					if (!string.IsNullOrWhiteSpace(versionParts[2]))
+						productType = " " + versionParts[2].Trim();
 					goto case 2;
 				}
 				case 2:
 				{
-					int date = Convert.ToInt32(versionParts[1]);
+					// A non-numeric token here would otherwise throw out of a property getter
+					// that the splash screen and About box read.
```
jasonleenaylor, line 237: Item 2 asked you to drop consumer references, and you dropped both of the ones I named -- this is a third, added with the guard. It's a weaker case than the others, since it justifies the `TryParse` rather than describing a member's contract in its own summary, and the guard exists because I asked for it. Flagging only so the pattern doesn't creep back: the reason the guard matters is that a throw escapes a property getter, which stands without naming who calls it.

jasonleenaylor (approved):
Everything from the first round is addressed, and item 3 went further than I asked.

The rebase resolved the conflict, and `AGENTS.md` is out of the diff entirely rather than reworked -- which was the right call, since `main`'s bullet is a strict superset.

On item 3, I said the body wording or `int.TryParse` was your choice. You did both: `Convert.ToInt32(versionParts[1])` is now guarded, which fixes the one line that could actually throw, and the body reframes the `MajorVersion` bug as a placeholder leak rather than non-defensive parsing. The `ManagedVwWindow` sentence now says the grep hits `mkall.targets:46` on `main` and nothing here, so the deletion reads as doing work instead of undercutting itself.

On item 2, I ran `Build/Agent/comment-hygiene.ps1 -BaseRef b8f546358` rather than measuring by eye: `comment-hygiene: clean`. The longest comment line the PR adds is 86 characters, and `kDefaultCopyrightString` now states its contract instead of the file's edit history.

One nit inline, not worth another round.

jasonleenaylor (approved):
> @jasonleenaylor reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/978)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #979, opened 2026-07-01
Fix RealDataTestsBase cross-worktree test collisions

johnml1135:
## Summary
- `RealDataTestsBase` (backing `RenderBaselineTests`, `RenderBenchmarkTestsBase`, and subclasses) resolved its scratch test-project directory via `FwDirectoryFinder.ProjectsDirectory`, a per-machine registry value that's correct for the shipping app (one shared Projects folder machine-wide) but means every git worktree of this repo collides on the same physical `integration_test_data` folder when running these tests.
- Symptom: `System.InvalidOperationException: Refusing to delete '...\integration_test_data' because the test sentinel file '.fieldworks-real-data-test-project' is missing` — one worktree's test run deletes/rewrites another's project mid-run.
- Fix (scoped entirely to this test fixture, `FwDirectoryFinder`'s production API is untouched):
  1. `DbDirectory()` now resolves under this checkout's own `SourceDirectory/../DistFiles/Projects`, matching the pattern already used by other real-cache test fixtures in this repo (`ConfiguredXHTMLGeneratorTests`, `RecordListTests`, etc.).
  2. The project name is suffixed with a short hash of the worktree's `SourceDirectory`, because `FwNewLangProjectModel.CheckForUniqueProjectName` — correct, unrelated production logic for the real "New Project" wizard — checks name uniqueness against the *shared* `ProjectsDirectory` regardless of where this fixture's files live, so a fixed name still collided across worktrees. The suffix is stable within a worktree (preserving the mutex-guarded "reusable project" perf intent) but unique across worktrees.

Found and fixed while running CI locally as part of an unrelated PR #964 review — see that PR's `PR_964_review.md` §15 for the full investigation trail.

## Test plan
- [x] `RenderBaselineTests` + `RealDataTestsBaseCleanupTests` (9 tests): went from 8/9 failing with the collision error to 9/9 passing.
- [ ] Full `./test.ps1` CI-equivalent run

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/979)

### DIFF
--- a/Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs
+++ b/Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs
@@ -1,5 +1,7 @@
 using System;
 using System.IO;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading;
 using NUnit.Framework;
 using SIL.FieldWorks.Common.FwUtils;
@@ -20,10 +22,21 @@ namespace SIL.FieldWorks.Common.RootSites.RootSiteTests
 	[TestFixture]
 	public abstract class RealDataTestsBase
 	{
-		private const string ReusableProjectName = "integration_test_data";
-		private const string ProjectMutexName =
-			@"Local\FieldWorks.RealDataTests.integration_test_data";
+		// The project name must be unique per worktree, so that multiple worktrees of the same repo
+		// can run tests in parallel without colliding on the same project directory.
+		private static readonly string ReusableProjectName = "integration_test_data_" + WorktreeSuffix();
+		private static readonly string ProjectMutexName =
+			@"Local\FieldWorks.RealDataTests." + ReusableProjectName;
 		private const string TestProjectSentinelFileName = ".fieldworks-real-data-test-project";
+
+		private static string WorktreeSuffix()
+		{
+			using (var sha = SHA1.Create())
+			{
+				var hash = sha.ComputeHash(Encoding.UTF8.GetBytes(FwDirectoryFinder.SourceDirectory));
+				return BitConverter.ToString(hash, 0, 4).Replace("-", string.Empty).ToLowerInvariant();
+			}
+		}
 
 		protected FwNewLangProjectModel m_model;
 		protected LcmCache Cache;
@@ -138,7 +151,11 @@ namespace SIL.FieldWorks.Common.RootSites.RootSiteTests
 
 		protected string DbDirectory(string name)
 		{
-			return Path.Combine(FwDirectoryFinder.ProjectsDirectory, name);
+			// Deliberately NOT FwDirectoryFinder.ProjectsDirectory - allow for multiple worktrees
+			// of the same repo to run tests in parallel without colliding on the same project directory.
+			var worktreeProjectsDirectory = Path.Combine(
+				Path.GetDirectoryName(FwDirectoryFinder.SourceDirectory), "DistFiles", "Projects");
+			return Path.Combine(worktreeProjectsDirectory, name);
 		}
 
 		private void AcquireProjectMutex()
--- a/Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs
+++ b/Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs
@@ -151,11 +151,13 @@ namespace SIL.FieldWorks.Common.RootSites.RootSiteTests
 
 		protected string DbDirectory(string name)
 		{
-			// Deliberately NOT FwDirectoryFinder.ProjectsDirectory - allow for multiple worktrees
-			// of the same repo to run tests in parallel without colliding on the same project directory.
-			var worktreeProjectsDirectory = Path.Combine(
-				Path.GetDirectoryName(FwDirectoryFinder.SourceDirectory), "DistFiles", "Projects");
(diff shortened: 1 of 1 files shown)


jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/979#-:-OwZbMxZEpo3fwZCfc-Z:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

jasonleenaylor (commented):
A commit comment that contains all the comments also present in the code is a bit much. Please make sure we commit a much more results focused comment rather than details of the solution and maybe shorten the code comments as well.

> @jasonleenaylor made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/979#-:-OwZclo172sWlim0Q_US:bslccnn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

___
*[`Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs` line 169 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/979#-OwZcex64txbLFiEVNaM:-OwZcex710MRBIAOxywd:b-1bgekx) ([raw file](https://github.com/sillsdev/FieldWorks/blob/7a61c4f8cb9e592544b97b42d7c09eb445e489b9/Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs#L169)):*
> ```Smalltalk
> 			// delete or rewrite another's mid-run). Anchor to this checkout's own DistFiles/Projects
> 			// instead - the same SourceDirectory-derived, registry-free pattern already used by other
> 			// real-cache test fixtures in this repo (e.g. ConfiguredXHTMLGeneratorTests, RecordListTests).
> ```

This comment is awefully long and wordy, wouldn't mind a pass at tightening it up.

johnml1135:
Cleaned up comments to be simpler.

johnml1135:

### HIGH — `DbDirectory()`'s worktree-local repoint doesn't match where the project actually gets validated/created, so the fix doesn't close the collision it targets

`RealDataTestsBase.cs:152-158` repoints `DbDirectory()` to a worktree-local path (`Path.GetDirectoryName(FwDirectoryFinder.SourceDirectory)\DistFiles\Projects`) instead of `FwDirectoryFinder.ProjectsDirectory`. But `EnsureSafeProjectDirectory` (line 346-353+), which every actual-creation path (`GetProjectDirectory`, called from `TestSetup` at line 82 right after `CreateNewLangProj`) is validated against, still hardcodes the expected directory as:

```csharp
var expectedProjectDirectory = NormalizePath(
    Path.Combine(FwDirectoryFinder.ProjectsDirectory, ReusableProjectName));
```

`FwDirectoryFinder.ProjectsDirectory` resolves from a machine-wide registry value (or `%CommonApplicationData%\SIL\FieldWorks\Projects` by default) — it can't be per-worktree. So the actual project created by `CreateNewLangProj` still lands under the shared, machine-wide `ProjectsDirectory` (that's what the validation in `EnsureSafeProjectDirectory` requires it to do, or `GetProjectDirectory` throws). The new worktree-suffixed `ReusableProjectName` (from the other hunk) does prevent same-name collisions across worktrees at that shared location, so that part is genuinely useful. But `DbDirectory()`'s worktree-local rerouting is now dead for the creation path and only affects the pre-create cleanup call at line 51 (`m_projectDirectory = DbDirectory(m_dbName); ... DeleteProjectDirectory(m_projectDirectory)` before `CreateNewLangProj` runs).

Concretely: if a prior test run crashes mid-setup and leaves an orphaned project directory at the real location (`ProjectsDirectory\integration_test_data_<hash>`), the next run's pre-create cleanup checks the wrong (worktree-local, never-populated) path and won't find or delete it — `CreateNewLangProj` then runs against a dirty leftover directory instead of a clean one.

Suggest either dropping the `DbDirectory()` change (the worktree suffix on `ReusableProjectName` already does the collision-avoidance work) or making `EnsureSafeProjectDirectory`'s expected path consistent with `DbDirectory()`'s.

johnml1135:
Fixed in 19dc4a30d. Traced it further: `CreateNewLangProj` → `LcmCache.CreateNewLangProj` always creates the project under `dirs.ProjectsDirectory` (`ILcmDirectories.ProjectsDirectory`, which is just `FwDirectoryFinder.ProjectsDirectory`) — it can't be redirected to a worktree-local path without deeper plumbing than this class has. So `EnsureSafeProjectDirectory`'s hardcoded expectation was actually the correct/necessary one; `DbDirectory()` was the one that had drifted.

Reverted `DbDirectory()` to `FwDirectoryFinder.ProjectsDirectory`. The collision-avoidance this PR is actually after still holds: `ReusableProjectName`'s worktree-hash suffix gives each worktree a distinct subdirectory name (and mutex name) under that same shared directory, so parallel worktrees no longer collide — and now the pre-create cleanup path checks the same location the project actually gets created/validated against.

jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/979#-:-OxREWwu42JZ42JgZBv1:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #988, opened 2026-07-02
Fix RegFree manifest-file race under parallel MSBuild

johnml1135:
Parallel FieldWorks builds can now generate registration-free COM manifests without racing when multiple MSBuild workers target the same output file. `RegFree.Execute()` serializes each manifest's complete read-modify-write operation across processes while unrelated manifests remain independent.

The key review question is whether the lock is narrow enough to preserve build parallelism but broad enough to protect every manifest mutation. The mutex is keyed by the normalized manifest path and encloses the load, update, write, and delete paths. Its name uses a deterministic digest because .NET string hash codes are process-specific. An abandoned mutex is treated as acquired, matching the platform contract, so the existing task error handling can report any damaged manifest instead of leaking an unhandled exception.

## Where to look

- `RegFree.Execute()` owns the mutex for the complete manifest transaction and releases it in `finally`.
- `ManifestLockName()` creates the same lock identity in every worker without globally serializing different manifests.
- `AbandonedMutexException` handling preserves ownership and lets normal manifest validation and error reporting continue.
- `RegFreeConcurrencyTests` runs 12 tasks against one manifest and verifies every result plus the final XML content.

## Deliberately not here

- No global build lock or serialization between unrelated manifest files.
- No COM registration, registry workaround, retry loop, or change to manifest contents.

## Verification

`./test.ps1 -TestProject Build/Src/FwBuildTasks/FwBuildTasksTests -StartedBy agent` passed locally: 146 managed tests passed, 3 skipped, and the repository script's 31 native smoke tests passed. The full FieldWorks suite was not rerun locally after the rebase; CI remains the full gate.

---

<details>

This focused branch carried no Markdown research or working notes to preserve or delete. The decision record lives here because the source comments intentionally retain only the local concurrency invariants a maintainer needs while reading the code.

</details>

<details>

- The lock covers the entire manifest transaction, not only `XmlWriter.Create()`. A worker must not read another worker's partially updated state or overwrite changes made after its own read.
- The lock is path-specific. Builds can still generate different manifests concurrently.
- The normalized full path is converted to a deterministic MD5 digest for the mutex name. `string.GetHashCode()` was rejected because its value can differ between worker processes. MD5 is used only as a stable identifier, not for security.
- `AbandonedMutexException` is a successful acquisition signal with warning semantics. Continuing under ownership allows the existing XML load and task error handling to detect and report a manifest left incomplete by the previous holder.

</details>

<details>
<summary><b>Regression evidence</b></summary>

The original failure was an intermittent `IOException` while parallel workers wrote the same manifest during PR #964; rerunning the same CI job passed, which isolated the problem to timing rather than deterministic input.

The regression test compiles a COM-visible assembly, launches 12 concurrent `RegFree.Execute()` calls against one manifest, requires every call to succeed, loads the result as XML, and verifies the expected `clrClass`. During development it failed in 3 of 3 runs with the mutex removed and passed in 5 of 5 runs with the mutex restored. The final squashed commit passed the complete `FwBuildTasksTests` project locally: 146 passed and 3 skipped.

</details>

- - -
[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/988)

### DIFF
--- a/Build/Src/FwBuildTasks/FwBuildTasksTests/RegFreeConcurrencyTests.cs
+++ b/Build/Src/FwBuildTasks/FwBuildTasksTests/RegFreeConcurrencyTests.cs
@@ -1,0 +1,119 @@
+// Copyright (c) 2026 SIL International
+// This software is licensed under the LGPL, version 2.1 or later
+// (http://www.gnu.org/licenses/lgpl-2.1.html)
+
+using System;
+using System.CodeDom.Compiler;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using System.Xml;
+using Microsoft.CSharp;
+using NUnit.Framework;
+using SIL.FieldWorks.Build.Tasks;
+
+namespace SIL.FieldWorks.Build.Tasks.FwBuildTasksTests
+{
+	/// <summary>
+	/// Verifies concurrent <see cref="RegFree"/> executions produce a valid shared manifest.
+	/// </summary>
+	[TestFixture]
+	public sealed class RegFreeConcurrencyTests
+	{
+		private const string AsmNamespace = "urn:schemas-microsoft-com:asm.v1";
+
+		[Test]
+		public void Execute_ConcurrentInvocationsTargetingSameManifest_AllSucceedAndProduceValidXml()
+		{
+			var tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N"));
+			Directory.CreateDirectory(tempDir);
+			var assemblyPath = Path.Combine(tempDir, "RegFreeConcurrencyTestAssembly.dll");
+			var manifestPath = Path.Combine(tempDir, "Shared.manifest");
+
+			try
+			{
+				CompileComVisibleAssembly(assemblyPath);
+
+				const int concurrentInvocations = 12;
+				var results = new bool[concurrentInvocations];
+				var tasks = new Task[concurrentInvocations];
+				for (var i = 0; i < concurrentInvocations; i++)
+				{
+					var index = i;
+					tasks[index] = Task.Run(() =>
+					{
+						var regFree = new RegFree
+						{
+							BuildEngine = new FwBuildTasks.TestBuildEngine(),
+							Executable = assemblyPath,
+							Output = manifestPath,
+							ManagedAssemblies = new Microsoft.Build.Utilities.TaskItem[]
+							{
+								new Microsoft.Build.Utilities.TaskItem(assemblyPath)
+							},
+							Platform = "x64"
+						};
+						results[index] = regFree.Execute();
+					});
+				}
+
+				Task.WaitAll(tasks);
+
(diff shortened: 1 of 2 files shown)


johnml1135:

Overall: the mutex approach is sound and the regression test (12 concurrent `Execute()` calls, verified to actually fail without the fix) is convincing evidence. One correctness gap worth closing before merge:

### MEDIUM — `WaitOne()` isn't wrapped for `AbandonedMutexException`, so a crashed prior holder turns a handled build error into an unhandled one

`RegFree.cs`:
```csharp
using (var manifestLock = new Mutex(false, ManifestLockName(manifestFile)))
{
    manifestLock.WaitOne();
    try
    {
        ...
    }
    catch (Exception e)
    {
        Log.LogErrorFromException(e, true, true, null);
        return false;
    }
    finally
    {
        manifestLock.ReleaseMutex();
    }
}
```

`WaitOne()` sits outside the `try`. If a previous MSBuild worker holding this mutex is killed mid-write (a build cancellation, `taskkill`, OOM), the next waiter's `WaitOne()` throws `AbandonedMutexException` — and since that's before the `try`, it propagates out of `Execute()` as an unhandled exception instead of going through the existing `Log.LogErrorFromException` path, and `ReleaseMutex()` in the `finally` never runs either (the `finally` only guards the `try` block it's attached to). MSBuild will still fail the task, but with a raw unhandled-exception task failure instead of the clean, logged error this PR is otherwise adding.

Note that `AbandonedMutexException` actually means ownership *was* granted, so the fix is a `catch` around `WaitOne()` that treats it as a successful acquire (and probably logs that the manifest content from the crashed writer should be treated as suspect, since it could reflect a torn write) — not a `return false`. Worth comparing with how PR #979 handles the same category of mutex (`AcquireProjectMutex` there does exactly this: `try { m_projectMutex.WaitOne(); } catch (AbandonedMutexException) { }`).

johnml1135:
Fixed in 73ea066b6: wrapped `WaitOne()` in a `catch (AbandonedMutexException)` that proceeds as a normal acquire (ownership is still granted on that exception), so a crashed prior holder now produces the intended logged build error instead of an unhandled task failure. Verified the existing `RegFreeConcurrencyTests` still passes.

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/988#-:-OxghI5m3A4NJzjh_uan:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

jasonleenaylor:
@copilot resolve the merge conflicts in this pull request

jasonleenaylor (approved):
> @jasonleenaylor reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/988)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #906, opened 2026-05-21
Retire Linux-era view shims

johnml1135:
## Summary
- remove the retired Linux-era managed Views shims `ViewInputManager` and `ManagedVwWindow`, including their solution and build wiring
- keep the Windows-native RootSite path in place by using `VwTextStore`/`IViewInputMgr` and HWND client-rectangle geometry while removing stale `IVwWindow`/`VwWindow` interop exposure
- update reg-free manifest inputs, architecture/OpenSpec docs, and task tracking for the shim retirement

## Validation
- `./build.ps1`
- `./test.ps1 -SkipManaged -TestProject TestViews`
- `./test.ps1 -TestProject FwBuildTasksTests -TestFilter "FullyQualifiedName~RegFreeCreator"`
- `./test.ps1 -TestProject SimpleRootSiteTests`
- `CI: Whitespace check`
- Manual smoke in rebuilt FieldWorks app covering lexical edit view typing, keyboard switching, IME/composition, selection/PageUp/PageDown, and launch/manifest sanity

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/906)

PR #989, opened 2026-07-03
Build: drop retired ManagedVwWindow.dll from RegFree managed COM list

johnml1135:
## Summary

`ManagedVwWindow.dll` was retired with the Linux-era view shims (#906). That cleanup updated `Build/mkall.targets` (excludes its CLSID) and `Build/Installer.legacy.targets` (lists it as removed-since-last-base), but **missed `Build/RegFree.targets`**, which still listed the DLL in `ManagedComAssemblies`.

Because the file no longer exists, the `RegFree` task emits a warning for every EXE that imports the target:

```
Build\RegFree.targets(91,3): warning : Failed to process managed assembly
  ...\Output\Debug\ManagedVwWindow.dll: Could not find file
  '...\Output\Debug\ManagedVwWindow.dll'.
  [Src\UnicodeCharEditor\UnicodeCharEditor.csproj]
  [Src\LCMBrowser\LCMBrowser.csproj]
```

## Change

Remove the stale `<ManagedComAssemblies Include="$(OutDir)ManagedVwWindow.dll" />` entry (replaced with an explanatory comment). The remaining managed COM assemblies (`FwUtils.dll`, `SimpleRootSite.dll`) are real and unaffected.

## Impact

- Clears the warning from both `UnicodeCharEditor` and `LCMBrowser` builds (one root cause, surfaced once per EXE).
- No behavior change: the DLL doesn't exist, so it was never actually being processed — only warned about.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/989)

### DIFF
--- a/Build/RegFree.targets
+++ b/Build/RegFree.targets
@@ -66,7 +66,7 @@
 		<!-- Explicitly list managed assemblies that expose COM types -->
 		<ManagedComAssemblies Include="$(OutDir)FwUtils.dll" />
 		<ManagedComAssemblies Include="$(OutDir)SimpleRootSite.dll" />
-		<ManagedComAssemblies Include="$(OutDir)ManagedVwWindow.dll" />
+		<!-- ManagedVwWindow.dll was retired with the Linux-era view shims (#906); do not process it. -->
 	</ItemGroup>
 	<ItemGroup>
 		<ManagedComAssemblies Remove="$(OutDir)*.resources.dll" />


johnml1135:

### MEDIUM — the fallback-list removal doesn't reach `FieldWorks.exe`, so the warning this PR targets keeps firing for the main product

`Build/RegFree.targets:68` (the `<ManagedComAssemblies Include="$(OutDir)ManagedVwWindow.dll" />` line) sits in an `ItemGroup` gated by `Condition="'@(ManagedComAssemblies)' == ''"` — it's only a fallback used when nothing has pre-populated `ManagedComAssemblies`. For `FieldWorks.exe` specifically, `Src/Common/FieldWorks/BuildInclude.targets:10` pre-populates that list explicitly and still includes `<ManagedComAssemblies Include="$(OutDir)ManagedVwWindow.dll" />`, so the fallback in `RegFree.targets` never applies to the main executable's build — it will keep emitting `Could not find file ...ManagedVwWindow.dll` for `FieldWorks.exe` after this PR merges. The change is real and useful for the other projects that do hit the fallback (LCMBrowser, GenerateHCConfig, UnicodeCharEditor, ComManifestTestHost), just incomplete for the main product. Suggest also removing the `BuildInclude.targets:10` entry (and checking `Build/mkall.targets:46` for a related stale CLSID-exclusion comment).

### Note — overlaps with #978

#978 removes this exact same `RegFree.targets` line outright (no replacement comment); this PR replaces it with an explanatory comment instead. Whichever merges first will make the other a no-op/conflict on this one line — worth deciding which PR carries this change and dropping it from the other.

johnml1135:
Close as a duplicate of #978

PR #990, opened 2026-07-03
Add cross-platform code coverage tooling + CI wiring

johnml1135:
## Summary
- The existing `/EnableCodeCoverage` MSBuild flag (`Build/SetupInclude.targets`) depends on a VS-Enterprise-only "Code Coverage" data collector and silently fails to collect on Community/Build Tools installs (confirmed live: `vstest.console.exe` reports `Unable to find a datacollector with friendly name 'Code Coverage'`).
- Add `coverlet.collector` instead, which registers the cross-platform "XPlat Code Coverage" collector and works under plain `vstest.console.exe` on any VS edition.
- Add a `-Coverage` switch to `test.ps1` that collects coverage (resolving coverlet's `TestAdapterPath` from the centrally-pinned package version, same pattern as the existing NUnit adapter lookup) and renders a summary/HTML report via a local ReportGenerator dotnet tool (`.config/dotnet-tools.json`).
- Wire `-Coverage` into `CI.yml`'s test step and upload the rendered report as a build artifact.
- `test.ps1` now runs `dotnet tool restore` before invoking ReportGenerator so a fresh checkout/CI runner doesn't just warn and skip the report.
- Add the `fieldworks-test-coverage` skill (how to check that new/changed lines are actually exercised, not just that a test file exists nearby) and reference it from `execute-implement`/`verify-test` and the base rubric's `test_coverage` criterion.

## Test plan
- [x] `.\build.ps1` — full solution build, clean
- [x] `.\test.ps1 -TestProject "Src/Common/FwUtils/FwUtilsTests" -Coverage` on this branch (built against plain `main`, not any feature branch) — 375/375 tests passed, coverage collected and summary/HTML report rendered correctly
- [x] Confirmed the previous `/EnableCodeCoverage` flag does *not* work on a VS 2022 Community install (motivates the coverlet switch)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/990)

### DIFF
--- a/test.ps1
+++ b/test.ps1
@@ -45,12 +45,10 @@
 	Skips acquiring/releasing the same-worktree lock because the parent build already owns it.
 
 .PARAMETER Coverage
-	Collect code coverage by wrapping vstest.console.exe with the dotnet-coverage tool (Microsoft
-	CodeCoverage CLI; free on any Visual Studio edition). Collection uses dynamic, in-memory
-	instrumentation, so nothing in the shared Output/<Configuration> folder is rewritten on disk and
-	parallel testhosts remain safe. Writes a merged coverage.cobertura.xml under
-	Output/<Configuration>/TestResults, then renders a summary + HTML report via the local
-	ReportGenerator tool (Output/<Configuration>/TestResults/CoverageReport).
+	Collect code coverage by wrapping vstest.console.exe with the dotnet-coverage tool. Its dynamic,
+	in-memory instrumentation leaves the shared Output/<Configuration> assemblies untouched, so
+	parallel testhosts stay safe. Writes coverage.cobertura.xml under Output/<Configuration>/TestResults,
+	then renders a summary + HTML report via ReportGenerator. If coverage can't be collected, the run fails.
 
 .EXAMPLE
 	.\test.ps1
@@ -304,6 +302,7 @@ $cleanupArgs = @{
 }
 
 $testExitCode = 0
+$script:coverageFailed = $false
 
 try {
 	if (-not $SkipWorktreeLock) {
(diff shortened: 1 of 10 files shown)


johnml1135:
## Code review + fixes (pushed in 6ca2c6b33)

Reviewed the coverage wiring at medium effort. Found and fixed three issues in `test.ps1` / `CI.yml`; verified one suspected issue was a non-issue.

### Fixed

1. **Stale coverage files skewed the local report.** `$resultsDir` is never cleaned and the report step globs `coverage*.cobertura.xml`, so per-assembly files left behind by the `-1` retry path would be merged into a later run's report. Now `coverage*.cobertura.xml` is deleted before collection.

2. **Coverage-tooling failure aborted the whole test run.** Wrapping vstest in `dotnet tool run dotnet-coverage` meant that if `dotnet tool restore` failed (no `dotnet` on PATH, offline, etc.) the tests never ran and the step failed. `test.ps1` now checks the restore result and falls back to running tests *without* coverage instead of turning a tooling problem into a test failure.

3. **Codecov could receive partial coverage.** The upload pointed at the primary `coverage.cobertura.xml`, which holds only the failed primary run when the multi-assembly `-1` retry path fires (real data then lives in per-assembly files). `test.ps1` now also emits a merged `CoverageReport/Cobertura.xml` via ReportGenerator, and CI uploads that merged file. (Rebased on top of @jasonleenaylor's `452ffe74e`, keeping the v7 SHA-pin / `flags` / `name`.)

### Verified, no change needed

- **`-1` retry workaround under `-Coverage`.** Confirmed empirically that `dotnet-coverage collect` propagates the wrapped process's exit code (including `-1`) for managed children, so the multi-assembly retry still triggers.

### One thing to double-check

`fail_ci_if_error: true` (from `452ffe74e`) now interacts with fix #2: if the coverage-tooling fallback ever engages on the CI runner, no `Cobertura.xml` is produced and the upload step would fail CI. On the runner `dotnet tool restore` should always succeed, so this is only a concern if the runner's `dotnet`/tooling changes — flagging it so it's a conscious choice.

test.ps1
```diff
@@ -695,10 +695,31 @@ try {
...
+			# Remove stale coverage artifacts from prior runs. $resultsDir is not cleaned
+			# between runs, and the report step globs coverage*.cobertura.xml, so leftover
+			# per-assembly files would otherwise be merged into this run's report.
+			Get-ChildItem -Path $resultsDir -Filter "coverage*.cobertura.xml" -ErrorAction SilentlyContinue |
+				Remove-Item -Force -ErrorAction SilentlyContinue
+
 			# Restore the local tool manifest (cheap no-op once restored) so dotnet-coverage and
-			# ReportGenerator are available on fresh checkouts/CI runners.
-			& dotnet tool restore 2>&1 | Out-Null
+			# ReportGenerator are available on fresh checkouts/CI runners. If the tooling can't be
```
jasonleenaylor: Is this for CI or for AI? If you ask for coverage and can't get it we really should error I think.

test.ps1
```diff
@@ -822,7 +843,10 @@ try {
 				$coverageReportDir = Join-Path $resultsDir "CoverageReport"
 				$coverageReportsGlob = Join-Path $resultsDir "coverage*.cobertura.xml"
 				try {
-					& dotnet tool run reportgenerator "-reports:$coverageReportsGlob" "-targetdir:$coverageReportDir" "-reporttypes:TextSummary;Html" 2>&1 |
+					# Also emit a single merged Cobertura (CoverageReport/Cobertura.xml) so downstream
```
jasonleenaylor: Change comment to 'Emit a single merged Cobertura report for Codecov'

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/990#-:-OxCri9l7sn_naSGool2:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #993, opened 2026-07-06
Fix for LT-22563

AndyBlack:
This turned out to be a continuation of [LT-22422](https://jira.sil.org/browse/LT-22422) "XAmple can incorrectly report success for a variant referring to an entry with multiple MSA"  The fix there took care of  Try a Word, but not what fed into the parser filer.  This fixes that.
The result from XAmple can include MSI elements like these:
<MSI DbRef='19938.0.14573'/>
<MSI DbRef='19938.0.9208'/>
which have three parts: the lex entry hvo, the lex ref hvo. and the msa hvo.  We were ignoring the msa hov when passing the XAmple parsing result to the parser filer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/993)

### DIFF
--- a/Src/LexText/ParserCore/XAmpleParser.cs
+++ b/Src/LexText/ParserCore/XAmpleParser.cs
@@ -245,9 +245,11 @@ namespace SIL.FieldWorks.WordWorks.Parser
 			//       required slots in affix templates.  The parser filer can ignore these.
 			// 3. <MSI DbRef="y"... and y is an hvo for a LexEntry.
 			//       The LexEntry is a variant form for the first set of LexEntryRefs.
-			// 4. <MSI DbRef="y"... and y is an hvo for a LexEntry followed by a period and an index digit.
-			//       The LexEntry is a variant form and the (non-zero) index indicates
-			//       which set of LexEntryRefs it is for.
+			// 4. <MSI DbRef="y.i.z"... and y is an hvo for a LexEntry, i is an index to the
+			//       LexEntryRef, and z is the hvo of the specific MSA (of that LexEntryRef's
+			//       target entry) that was used. The LexEntry is a variant form; i indicates
+			//       which set of LexEntryRefs it is for, and z indicates which of that entry's
+			//       MSAs to use, since it can have more than one (see LT-22422/LT-22563).
 			ICmObject objForm;
 			if (!cache.ServiceLocator.GetInstance<ICmObjectRepository>().TryGetObject(int.Parse(formHvo), out objForm))
 			{
--- a/Src/LexText/ParserCore/XAmpleParser.cs
+++ b/Src/LexText/ParserCore/XAmpleParser.cs
@@ -262,7 +262,7 @@ namespace SIL.FieldWorks.WordWorks.Parser
 			}
 
 			// Irregulary inflected forms can have a combination MSA hvo: the LexEntry hvo, a period, and an index to the LexEntryRef
-			Tuple<int, int, int> msaTuple = ProcessMsaHvo(msaHvo);
+			Tuple<int, int, int> msaTuple = ParserXmlWriterExtensions.ProcessMsaHvo(msaHvo);
 			ICmObject objMsa;
 			if (!cache.ServiceLocator.GetInstance<ICmObjectRepository>().TryGetObject(msaTuple.Item1, out objMsa))
 			{
@@ -307,15 +307,6 @@ namespace SIL.FieldWorks.WordWorks.Parser
 			// if it is anything else, we ignore it
 			morph = null;
 			return true;
-		}
-
-		private static Tuple<int, int, int> ProcessMsaHvo(string msaHvo)
-		{
-			string[] msaHvoParts = msaHvo.Split('.');
-			// the msa hvo has one part or three parts separated by a period.
-			// in the latter case, it is the lex entry hvo, the lex ref hvo, and the msa hvo
-			return Tuple.Create(int.Parse(msaHvoParts[0]), msaHvoParts.Length == 3 ? int.Parse(msaHvoParts[1]) : 0,
-				msaHvoParts.Length == 3 ? int.Parse(msaHvoParts[2]) : 0);
 		}
 
 		public XDocument ParseWordXml(string word)
--- a/Src/LexText/ParserCore/XAmpleParser.cs
+++ b/Src/LexText/ParserCore/XAmpleParser.cs
@@ -261,7 +261,8 @@ namespace SIL.FieldWorks.WordWorks.Parser
 				return true;
 			}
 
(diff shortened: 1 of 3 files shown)


Src/LexText/ParserCore/XAmpleParser.cs
```diff
@@ -262,7 +262,7 @@ private static bool TryCreateParseMorph(LcmCache cache, XElement morphElem, out
 			}
 
 			// Irregulary inflected forms can have a combination MSA hvo: the LexEntry hvo, a period, and an index to the LexEntryRef
-			Tuple<int, int> msaTuple = ProcessMsaHvo(msaHvo);
+			Tuple<int, int, int> msaTuple = ParserXmlWriterExtensions.ProcessMsaHvo(msaHvo);
```
johnml1135, line 268: Nit: the doc comment above this method (lines 236-250, cases 3 and 4) still describes the old 1-part (`y`) and 2-part (`y.index`) MSI `DbRef` encodings. `ProcessMsaHvo` now only special-cases exactly 3 dot-separated parts (`lexEntryHvo.refIndex.msaHvo`); anything else collapses to refIndex=0/msaHvo=0. Worth updating the comment so it matches the format actually produced since LT-22422, or a future maintainer may reintroduce a 2-part-aware branch that no longer applies.

johnml1135:
General comment (not diff-anchored — this file isn't touched by this PR, so GitHub won't let me attach it to a line):

**`Src/Utilities/pcpatrflex/ToneParsFLExDll/ToneParsInvoker.cs`, `TryCreateParseMorph`/`ProcessMsaHvo` (~line 799-891)**

This class's `TryCreateParseMorph`/`ProcessMsaHvo` is explicitly labeled "from XAmpleParser.cs" (see the comment at line 729: *"ConvertParserFilerResultXmlToParseResult(), TryCreateParseMorph(), and ProcessMsaHvo() are from XAmpleParser.cs ... Ideally, we'd expose them from XAmpleParser.cs"*). It still has the old 2-tuple `ProcessMsaHvo` that discards the MSA hvo — the exact bug this PR fixes for XAmple. Since `ToneParsInvoker` is a live, separately-tested class (see `ToneParsInvokerTests.cs`), the PC-PATR/tone-parser path will still reproduce the "valid parse not in the list" symptom from LT-22563/LT-22422 for entries with multiple MSAs. Might be worth either applying the same fix there, or (per the existing TODO) finally exposing `ProcessMsaHvo`/`TryCreateParseMorph` from `XAmpleParser.cs` so there's one implementation.

johnml1135:
General comment (not diff-anchored — no test files are touched by this PR):

**Missing test coverage for the fixed code path**

Neither `XAmpleParserTests.cs` nor `ParseFilerProcessingTests.cs` exercises `TryCreateParseMorph`/`ProcessMsaHvo` against a 3-part MSI `DbRef` (e.g. `19938.0.14573`) resolving to an `IMoStemMsa`. This is exactly the kind of format-parsing detail that regressed silently once already (LT-22422 fixed the writer side but not the parser-filer side, which is what led to LT-22563). A regression test mirroring the JIRA acceptance test — a variant LexEntry with multiple MSAs parsed through `ParseWordXml`/`TryCreateParseMorph` — would guard against this reappearing in a future refactor.

johnml1135:
All thee comments AI generated - human review forthcoming.

Src/LexText/ParserCore/XAmpleParser.cs
AndyBlack: Good point.  Done.

AndyBlack:
> General comment (not diff-anchored — this file isn't touched by this PR, so GitHub won't let me attach it to a line):
> 
> **`Src/Utilities/pcpatrflex/ToneParsFLExDll/ToneParsInvoker.cs`, `TryCreateParseMorph`/`ProcessMsaHvo` (~line 799-891)**
> 
> This class's `TryCreateParseMorph`/`ProcessMsaHvo` is explicitly labeled "from XAmpleParser.cs" (see the comment at line 729: _"ConvertParserFilerResultXmlToParseResult(), TryCreateParseMorph(), and ProcessMsaHvo() are from XAmpleParser.cs ... Ideally, we'd expose them from XAmpleParser.cs"_). It still has the old 2-tuple `ProcessMsaHvo` that discards the MSA hvo — the exact bug this PR fixes for XAmple. Since `ToneParsInvoker` is a live, separately-tested class (see `ToneParsInvokerTests.cs`), the PC-PATR/tone-parser path will still reproduce the "valid parse not in the list" symptom from LT-22563/LT-22422 for entries with multiple MSAs. Might be worth either applying the same fix there, or (per the existing TODO) finally exposing `ProcessMsaHvo`/`TryCreateParseMorph` from `XAmpleParser.cs` so there's one implementation.

I will create a separate JIRA issue for this.

AndyBlack:
> **Missing test coverage for the fixed code path**
I wish I knew how to create unit tests for these kinds of things.  I've never been able to figure out how to do it; hence, there aren't any.  Do you have ideas on how it could be done?

johnml1135:
Added a regression test for this in c350a5807: `ParseFilerProcessingTests.TryCreateParseMorph_IrregularVariantWithMultipleMsas_UsesMsaFromDbRef`.

It builds an irregularly inflected variant entry ("creb") of a base entry ("seek") that has *two* MSAs, invokes the private `XAmpleParser.TryCreateParseMorph` (via reflection) with a hand-built `<Morph><MoForm DbRef="..."/><MSI DbRef="lexEntryHvo.0.msaHvo"/></Morph>` element pointing at the second MSA — mimicking what XAmple echoes back — and asserts the resulting `ParseMorph.Msa` is that exact second MSA, not the variant's first/main sense (which is what the pre-fix code would have produced via `GetMainOrFirstSenseOfVariant`). Confirmed it exercises the fixed path by tracing it against the still-unfixed duplicate of this logic in `ToneParsInvoker.cs` (flagged in an earlier comment), which would return the wrong (first) MSA for this same setup.

Src/LexText/ParserCore/XAmpleParser.cs
johnml1135: Updated case 4 in `f6e1099cf` to describe the current 3-part `y.i.z` (lexEntryHvo.refIndex.msaHvo) format.

johnml1135 (approved):
Looks good - just make sure the new JIRA issue is linked.

johnml1135 (approved):
> @johnml1135 reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/993)* status: all files reviewed, 1 unresolved discussion (waiting on AndyBlack).

johnml1135:
New JIRA issue: https://jira.sil.org/browse/LT-22607

PR #994, opened 2026-07-06
Fix LT-22591: Crash in SFM import Modify Mapping dialog

jasonleenaylor:
This broke during the modernization commit because there was a file to class mismatch and the modern build code expects them to match

Rename the files to match the class so the default SDK naming produces the name the runtime requests. This also aligns with the localization build (ProjectLocalizer), which names satellite resources after the resx file: once the translated resx files are renamed on the Crowdin side, this dialog's translations will display for the first time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/994)


jtmaxwell3 (approved):
> @jtmaxwell3 partially reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/994)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jasonleenaylor).

PR #995, opened 2026-07-07
LT-22570: Fix WS order mismatch after restoring a project

aror92:
Deserialization can produce different WS ordering in the currentWS and allWS lists.

After restoring Sena3 (and some other projects) from backup, the WS order no longer matches between Lexicon Edit and Format Writing Systems.

Previously, the working WS list was built from currentWS, with any unselected WS appended from allWS. This preserved the order shown in Lexicon Edit after a restore, but reordered unselected WS.

That behavior was later changed to build the working list from allWS so unselected WS would retain their original order. However, this reintroduced the WS ordering discrepancy after restoring from backup.

Build a synchronized WS order that satisfies both requirements by preserving the currentWS order for selected WS while maintaining the posotion of unselected WS from allWS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/995)

### DIFF
--- a/Src/FwCoreDlgs/FwWritingSystemSetupModel.cs
+++ b/Src/FwCoreDlgs/FwWritingSystemSetupModel.cs
@@ -274,7 +274,7 @@ namespace SIL.FieldWorks.FwCoreDlgs
 			}
 
 			// If there are no common elements, return the list of all WS.
-			return (List<CoreWritingSystemDefinition>)allWS;
+			return allWS.ToList();
 
 		}
 
--- a/Src/FwCoreDlgs/FwWritingSystemSetupModel.cs
+++ b/Src/FwCoreDlgs/FwWritingSystemSetupModel.cs
@@ -212,13 +212,70 @@ namespace SIL.FieldWorks.FwCoreDlgs
 		private List<WSListItemModel> BuildWorkingList(ICollection<CoreWritingSystemDefinition> allForType, IList<CoreWritingSystemDefinition> currentForType)
 		{
 			var list = new List<WSListItemModel>();
-			// Don't reorder the writing systems here even though the lexical data fields are reordered (cf. LT-22136).
-			var wssInPreferredOrder = allForType;
+
+			// Because of the deserialization process, when restoring a project from backup,
+			// it is possible for the order of writing systems to differ between the all and current lists,
+			// even when they contain all the same elements.
+			// Synchronize the order of selected writing systems between the all and current lists (LT-19664 and LT-22570),
+			// while preserving the order of writing systems that are unselected in Format Writing Systems (cf. LT-22136).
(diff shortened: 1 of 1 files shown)


jasonleenaylor:
*[`Src/FwCoreDlgs/FwWritingSystemSetupModel.cs` line 277 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/995#-OwyE1h12zTobbsJDw6d:-OwyE1h12zTobbsJDw6e:bjy2u9i) ([raw file](https://github.com/sillsdev/FieldWorks/blob/91f9d635a4592f18bfad6fdf53c401904a75e996/Src/FwCoreDlgs/FwWritingSystemSetupModel.cs#L277)):*
> ```Smalltalk
> 
> 			// If there are no common elements, return the list of all WS.
> 			return (List<CoreWritingSystemDefinition>)allWS;
> ```

This isn't a safe return, if allWs *is* a List then the parameter type should be changed and no cast is needed, if it isn't a List then this will crash and you should do something like new List<cwsd>(allWS)

aror92 (commented):
> @aror92 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/995)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on jasonleenaylor).

___
*[`Src/FwCoreDlgs/FwWritingSystemSetupModel.cs` line 277 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/995#-OwyE1h12zTobbsJDw6d:-OwyGbtC6uVjv_L6NFp2:b-1o0c5f) ([raw file](https://github.com/sillsdev/FieldWorks/blob/91f9d635a4592f18bfad6fdf53c401904a75e996/Src/FwCoreDlgs/FwWritingSystemSetupModel.cs#L277)):*

This isn't a safe return, if allWs *is* a List then the parameter type should be changed and no cast is needed, if it isn't a List then this will crash and you should do something like new List<cwsd>(allWS)
</blockquote></details>

Thanks for pointing this out. allWS is an ICollection<CoreWritingSytemDefiniton>. I replaced the cast with allWS.ToList().

(I just tried to test execution of this revised line of code by deselecting all writing systems, but that's not permitted in FieldWorks, so in the end this line shouldn't get hit, because there should always be overlap between allWs and currentWs.)

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/995#-:-OwyKBQX56rpHiht47Y_:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on aror92).

PR #1000, opened 2026-07-08
Chore: Consolidate dependabot bumps (nuget-minor, actions/checkout, actions/cache, actions-minor)

johnml1135:
## Summary
- Consolidates 5 open dependabot PRs into one branch/PR so they build and
  test together instead of as 5 separate merges: #997, #984, #982, #981, #980.
- Pulls in #979's fix for `RealDataTestsBase` cross-worktree collisions,
  which was needed to get a clean local CI run (see Test plan).
- Holds `Microsoft.Extensions.DependencyModel` at 9.0.16 instead of the
  9.0.17 that #984's nuget-minor group proposed (see below).

## Included dependabot bumps
- #997 Bump Microsoft.Build.Utilities.Core 18.6.3 -> 18.7.1 (+5 others)
- #984 Bump the nuget-minor group (10 updates): System.Drawing.Common,
  System.Reflection.Metadata, System.Resources.Extensions,
  System.Security.Permissions -> 9.0.17/10.0.9; SilLcm/SilLibPalaso version
  refresh in `Build/SilVersions.props`
- #982 Bump actions/checkout 6 -> 7 (all workflow files)
- #981 Bump actions/cache 5 -> 6 (copilot-setup-steps.yml)
- #980 Bump the actions-minor group (2 SHA-pinned action updates in
  CI.yml/base-installer-cd.yml)

## Why Microsoft.Extensions.DependencyModel stays at 9.0.16
`Directory.Packages.props` has carried a long-standing comment warning
that bumping this specific package breaks ICU initialization in the
.NET Framework test host. #984 bumped it to 9.0.17 anyway alongside the
other current-line packages. Running the full test suite locally
reproduced the exact failure the comment warns about:

```
System.TypeInitializationException : The type initializer for
'Icu.NativeMethods' threw an exception.
 ----> System.IO.FileLoadException : Could not load file or assembly
'Microsoft.Extensions.DependencyModel, Version=9.0.0.16, ...'
```

5 test fixtures (`HCWorkerServiceTests`, `ToWordAnalysisDto_*`) failed
`OneTimeSetUp` until this was reverted back to 9.0.16. All other
package bumps from the nuget-minor group are kept. #984 is being
closed with an explanatory comment rather than merged as-is, since
dependabot bundles it as one grouped update.

## Test plan
- [x] `.\build.ps1 -Configuration Debug -BuildTests` — 0 errors, 0 warnings
- [x] `.\test.ps1 -Configuration Debug -NoBuild -TestFilter 'TestCategory!=LongRunning&TestCategory!=ByHand&TestCategory!=SmokeTest&TestCategory!=DesktopRequired'`
      (same filter as `.github/workflows/CI.yml`) — 4298 tests, only 12
      failures, all pre-existing local `RenderBaselineTests`/
      `RenderBenchmark` pixel-diff mismatches (font/GPU rendering
      differences vs. the windows-2022 runner that captured the
      baselines), unrelated to this branch's changes and previously
      documented as a known local-environment gap.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1000)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: windows-2022
     steps:
       - name: Checkout Files
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         id: checkout
 
       - name: Build with tests
@@ -144,7 +144,7 @@ jobs:
           path: test-results
 
       - name: Publish Test Results
-        uses: EnricoMi/publish-unit-test-result-action@c950f6fb443cb5af20a377fd0dfaa78838901040
+        uses: EnricoMi/publish-unit-test-result-action@d0a4676d0e0b938bc201470d88276b7c74c712b3
         with:
           files: 'test-results/**/*.trx'
           check_name: NUnit Tests
--- a/.github/workflows/CommitMessage.yml
+++ b/.github/workflows/CommitMessage.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v6
+      uses: actions/checkout@v7
       with:
         fetch-depth: 0
     - name: Install dependencies
--- a/.github/workflows/base-installer-cd.yml
+++ b/.github/workflows/base-installer-cd.yml
@@ -62,14 +62,14 @@ jobs:
           echo "FW_BUILD_NUMBER=$combined" >> $env:GITHUB_ENV
 
       - name: Checkout Files
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         id: checkout
         with:
           ref: ${{ github.event.inputs.fw_ref || github.ref }}
           fetch-depth: 0
 
       - name: Checkout Helps
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         id: helps-checkout
         with:
           repository: 'sillsdev/FwHelps'
@@ -78,7 +78,7 @@ jobs:
           path: 'DistFiles/Helps'
 
       - name: Checkout PatchableInstaller
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         id: installer-checkout
         with:
           repository: 'sillsdev/genericinstaller'
@@ -87,7 +87,7 @@ jobs:
           path: 'PatchableInstaller'
 
       - name: Checkout Localizations
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
         id: loc-checkout
         with:
           repository: 'sillsdev/FwLocalizations'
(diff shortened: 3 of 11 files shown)


johnml1135:
Added breadcrumbs to the `Microsoft.Extensions.DependencyModel` pin comment in `Directory.Packages.props` so this doesn't need to be re-diagnosed the next time someone tries to bump it (amended into the same commit, force-pushed):

- **Exact failure signature** from reproducing this today: every test's `OneTimeSetUp` throws `TypeInitializationException` on `Icu.NativeMethods`, chained from a `FileLoadException` citing two conflicting requested versions of `DependencyModel` (our pin vs. a 2.0.4 floor). Future attempts can grep the `.trx` for `Icu.NativeMethods`/`DependencyModel` — zero hits means it's safe to bump.
- **Corrected/completed attribution of the 2.0.4 floor.** Verified via `Obj/**/project.assets.json` that it's `icu.net` *and* `SIL.LCModel.Core` (liblcm) both requesting 2.0.4 transitively — not just icu.net as originally noted. Also confirmed libpalaso (`SIL.Core`, `SIL.Core.Desktop`, `SIL.Windows.Forms`, `SIL.WritingSystems`) and Chorus don't reference this package at all, so they're not part of the conflict.
- **A lead on an actual fix**, not just "wait and retry": there's no `bindingRedirect` for this assembly anywhere in the repo, so the .NET Framework test host's auto-generated redirect isn't reconciling the two version requests. Adding an explicit `<dependentAssembly>` redirect to the NUnit test host config may unblock the bump directly instead of waiting on icu.net/liblcm to raise their floor.

No functional change — comment-only update to `Directory.Packages.props`.

johnml1135:
Trimmed the \`DependencyModel\` comment - it had ballooned to 28 lines, way out of proportion with every other pin comment in this block (1-4 lines each). It now carries just the load-bearing facts (who wants what floor, the exact exception to grep for, a one-line retest procedure) and points back to this PR thread for the full investigation instead of duplicating it inline.

While in there, also fixed two pre-existing stale comments unrelated to this pin but part of the same dependabot bump: \`System.Drawing.Common\` and \`System.Security.Permissions\` both said "Pin to 9.0.16" while actually pinned at 9.0.17.

Amended into the same commit and force-pushed - still one commit, still gitlint-compliant.

jasonleenaylor:
*[`Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs` at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/1000#-OxBmFCUB9d5mhfUr-H3:-OxBmFCUB9d5mhfUr-H4:b45auuz) ([raw file](https://github.com/sillsdev/FieldWorks/blob/04a37560ffcfe19854941b189748b656756f1102/Src/Common/RootSite/RootSiteTests/RealDataTestsBase.cs)):*
This does not belong in this PR.

PR #1002, opened 2026-07-09
LT-22607 TonePars use XAmpleParser code for result processing

AndyBlack:
Now that the TonePars utility is part of FLEx, we can have it use the same code that XAmple does when processing its result.

1. This change took the result processing of XAmpleParser and put it in a new public method.
2. ToneParsInvoker now gets an instance of XAmpleParser (or creates it, if needed).
3. The change removes the duplicated code from ToneParsInvoker.
4. ToneParsFLExForm passes in an instance of XAmpleParser.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1002)

### DIFF
--- a/Src/LexText/ParserCore/XAmpleParser.cs
+++ b/Src/LexText/ParserCore/XAmpleParser.cs
@@ -68,7 +68,7 @@ namespace SIL.FieldWorks.WordWorks.Parser
 		/// </summary>
 		/// <param name="originalName">The original name to be converted</param>
 		/// <returns>Converted name</returns>
-		internal static string ConvertNameToUseAnsiCharacters(string originalName)
+		public static string ConvertNameToUseAnsiCharacters(string originalName)
 		{
 			var sb = new StringBuilder();
 			char[] letters = originalName.ToCharArray();
@@ -168,13 +168,13 @@ namespace SIL.FieldWorks.WordWorks.Parser
 		{
 			CheckDisposed();
 
-			var results = new StringBuilder(m_xample.ParseWord(word));
-			ParseResult result = ProcessParseResults(ref results);
+			var results = m_xample.ParseWord(word);
+			ParseResult result = ProcessParseResults(results, m_cache);
 
 			return result;
 		}
 
-		public ParseResult ProcessParseResults(ref StringBuilder results)
+		public static ParseResult ProcessParseResults(String results, LcmCache cache)
 		{
 			results = results.Replace("DB_REF_HERE", "'0'");
 			results = results.Replace("<...>", "[...]");
@@ -202,7 +202,7 @@ namespace SIL.FieldWorks.WordWorks.Parser
 			}
 
 			ParseResult result;
-			using (new WorkerThreadReadHelper(m_cache.ServiceLocator.GetInstance<IWorkerThreadReadHandler>()))
+			using (new WorkerThreadReadHelper(cache.ServiceLocator.GetInstance<IWorkerThreadReadHandler>()))
 			{
 				var analyses = new List<ParseAnalysis>();
 				foreach (XElement analysisElem in wordformElem.Descendants("WfiAnalysis"))
@@ -212,7 +212,7 @@ namespace SIL.FieldWorks.WordWorks.Parser
 					foreach (XElement morphElem in analysisElem.Descendants("Morph"))
 					{
 						ParseMorph morph;
-						if (!TryCreateParseMorph(m_cache, morphElem, out morph))
+						if (!TryCreateParseMorph(cache, morphElem, out morph))
 						{
 							skip = true;
 							break;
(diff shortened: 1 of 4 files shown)


Src/Utilities/pcpatrflex/ToneParsFLExDll/ToneParsInvoker.cs
```diff
@@ -73,6 +75,11 @@ LcmCache cache
 			DatabaseName = ConvertNameToUseAnsiCharacters(cache.ProjectId.Name);
 			InitFileNames();
 			Queue = new IdleQueue { IsPaused = true };
+			XAmpleParser = xAmpleParser;
+			if ( XAmpleParser == null )
+			{
+				XAmpleParser = new XAmpleParser(Cache, null);
```
johnml1135: **Claude - Sonnet 5**: This fallback constructs a full `XAmpleParser` (native XAmple engine init + a `ParserModelChangeListener` registered on `Cache.DomainDataByFlid`), but `ToneParsInvoker` has no `IDisposable`/`Dispose` and never disposes it — a genuine resource leak. Worse, it's not an edge case: in `ToneParsFLExForm.cs`, `m_XAmpleParser` is only populated on the HermitCrab code path, so whenever XAmple is the active parser, every "Parse Text"/"Parse Segment"/"Show Log" click passes `null` here and leaks a new native handle + listener each time.

Separately: this fallback passes `dataDir = null`, and `XAmpleParser.Update()` does `Path.Combine(m_dataDir, "Configuration", "Grammar")` with no null guard. Nothing in `ToneParsInvoker` calls `Update()`/`Reset()` today so this is latent, but every other production caller of `new XAmpleParser(...)` passes a real `dataDir` — this is a new, unguarded landmine if that ever changes.

Src/LexText/ParserCore/XAmpleParser.cs
```diff
@@ -169,6 +169,13 @@ public ParseResult ParseWord(string word)
 			CheckDisposed();
 
 			var results = new StringBuilder(m_xample.ParseWord(word));
+			ParseResult result = ProcessParseResults(ref results);
+
+			return result;
+		}
+
+		public ParseResult ProcessParseResults(ref StringBuilder results)
```
johnml1135: **Claude - Sonnet 5**: The `ref` on `StringBuilder results` is functionally unnecessary — `StringBuilder.Replace` mutates in place and returns the same instance, and neither call site (`ParseWord` here, or `ToneParsInvoker.SaveResultsInDatabase`) reads `results` after the call. A plain `string results` parameter would be simpler and would avoid the caller in `ToneParsInvoker.cs` (`SaveResultsInDatabase`) having to allocate a throwaway `StringBuilder` purely to satisfy this signature.

Also: this method doesn't call `CheckDisposed()`, unlike `Update()`/`Reset()`/`ParseWord()`/`ParseWordXml()` on this same `DisposableBase`-derived class. Lower severity — this mirrors a pre-existing gap in the code this PR deletes (which had the same `CheckDisposed()` omitted, with a `//TODO: fix!` comment), and there's no demonstrated live path where a disposed instance is called through today — but now that this method is called from another assembly (`ToneParsInvoker`), it's worth closing for consistency with the rest of the public API.

johnml1135 (commented):
**Claude - Sonnet 5** — automated code review.

Two additional findings sit on lines outside this PR's diff hunks (pre-existing code untouched here), so they can't be attached as inline comments — noting them here instead:

1. **Reuse (`Src/Utilities/pcpatrflex/ToneParsFLExDll/ToneParsInvoker.cs` ~line 101)**: `ConvertNameToUseAnsiCharacters` is a byte-for-byte duplicate of `XAmpleParser.ConvertNameToUseAnsiCharacters`, explicitly commented as "borrowed from SIL.FieldWorks.WordWorks.Parser (ParserCore.dll)". Since this PR already dedupes `ProcessParseResults` against the same class, and `ToneParsInvoker` now holds a live `XAmpleParser` reference, this is a natural candidate to dedupe too (note: `XAmpleParser`'s version is `internal static`, so calling through would need an accessibility change).
2. **Cleanup (`Src/Utilities/pcpatrflex/ToneParsFLExDll/ToneParsInvoker.cs` line 22)**: `using System.Xml.Linq;` is now dead — it was the only consumer's (`XElement`/`Descendants`) namespace, and that code was deleted by this PR. No remaining use of any `System.Xml.Linq` type in the file.

johnml1135 (commented):
> @johnml1135 reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1002)* status: all files reviewed, 2 unresolved discussions (waiting on AndyBlack).

johnml1135 (approved):

Src/LexText/ParserCore/XAmpleParser.cs
johnml1135: :LGTM:

Src/Utilities/pcpatrflex/ToneParsFLExDll/ToneParsInvoker.cs
johnml1135: :LGTM:

PR #1005, opened 2026-07-10
LT-22351: Preserve default font features in Preview and Word export

johnml1135:
Persisted OpenType and Graphite defaults now flow through LibLCM's
`BaseStyleInfo` into both Dictionary Preview CSS and Word export. FieldWorks no
longer maintains a second `StyleInfo` loader for `ktptFontVariations`.

The main review question is whether removing that compatibility loader leaves
either export path without coverage. The tests use real persisted `IStStyle`
rules rather than an in-memory test double, and they verify both propagation
and precedence: a style's own features override writing-system defaults while
the font family can still fall back normally. The branch uses the published
`SIL.LCModel` 11.0.0-beta0178 package, which contains the authoritative load
path from [sillsdev/liblcm#388](https://github.com/sillsdev/liblcm/pull/388).

## Where to look

- `StyleInfo.cs` removes the FieldWorks-only compatibility loader.
- `CssGeneratorTests` exercises persisted defaults and CSS precedence.
- `LcmWordGeneratorTests` exercises the equivalent Word export behavior.
- The archived design records why LibLCM is now the authoritative load path.

## Deliberately not here

- No replacement FieldWorks fallback or duplicate font-feature parser.
- No .NET 10 migration; [LibLCM #394](https://github.com/sillsdev/liblcm/pull/394) repaired the current .NET 8 CI instead.

## Stack and verification

LibLCM #388 and #394 are merged, and beta0178 is published on NuGet.org. The
FieldWorks Debug build passes with 0 warnings and 0 errors. The four focused
xWorks tests covering persisted defaults and precedence pass against beta0178.

[LT-22351](https://jira.sil.org/browse/LT-22351)

---

<details>

The load-path decision remains in the archived design because it constrains
future style handling. Temporary investigation and proposal narration was
removed from this branch; the dependency sequence is recorded here because it
is specific to landing this pull request.

</details>

<details>
<summary><b>Decisions and evidence</b></summary>

- LibLCM owns deserialization of default `ktptFontVariations`; FieldWorks owns
  consumption in Preview CSS and Word export.
- Regression tests construct a real style with persisted rules, matching the
  production `LcmStyleSheet` path that the old test doubles bypassed.
- Coverage includes the sibling precedence case so the fix cannot silently
  replace a style's own features with writing-system defaults.
- Focused tests passed against the published beta0178 package containing
  LibLCM #388.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1005)

### DIFF
--- a/Build/SilVersions.props
+++ b/Build/SilVersions.props
@@ -12,7 +12,7 @@
 		=============================================================
 	-->
 	<PropertyGroup Label="SIL Ecosystem Versions">
-		<SilLcmVersion>11.0.0-beta0176</SilLcmVersion>
+		<SilLcmVersion>11.0.0-beta0178</SilLcmVersion>
 		<SilLibPalasoVersion>18.0.0-beta0030</SilLibPalasoVersion>
 		<SilLibPalasoL10nsVersion>18.0.0-beta0012</SilLibPalasoL10nsVersion>
 		<SilChorusVersion>6.0.0-beta0065</SilChorusVersion>
--- a/Src/FwCoreDlgs/FwCoreDlgControls/StyleInfo.cs
+++ b/Src/FwCoreDlgs/FwCoreDlgControls/StyleInfo.cs
@@ -40,24 +40,6 @@ namespace SIL.FieldWorks.FwCoreDlgControls
 		public StyleInfo(IStStyle style)
 			: base(style)
 		{
-			LoadDefaultFontFeatures(style);
-		}
-
-		private void LoadDefaultFontFeatures(IStStyle style)
-		{
-			if (style == null || style.Rules == null)
-				return;
-
-			for (int i = 0; i < style.Rules.StrPropCount; i++)
-			{
-				int tpt;
-				string value = style.Rules.GetStrProp(i, out tpt);
-				if (tpt == (int)FwTextPropType.ktptFontVariations)
-				{
-					m_defaultFontInfo.m_features.ExplicitValue = value;
-					return;
-				}
-			}
 		}
 
 		/// ------------------------------------------------------------------------------------
--- a/Src/xWorks/xWorksTests/CssGeneratorTests.cs
+++ b/Src/xWorks/xWorksTests/CssGeneratorTests.cs
@@ -2415,6 +2415,36 @@ namespace SIL.FieldWorks.XWorks
 		}
 
 		[Test]
+		public void GenerateCssForConfiguration_DefaultFontFeaturesFromPersistedStyleRules_ReachPreviewCss()
+		{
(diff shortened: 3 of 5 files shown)


johnml1135:
Note: CI on this PR is expected to be red until sillsdev/liblcm#388 is released and Build/SilVersions.props is bumped. The committed SilVersions.props still pins upstream 11.0.0-beta0173, which lacks the liblcm fix, so both the pre-existing gate test (SaveToDB_DefaultFontFeatures_RoundTripsThroughRules) and the new CssGenerator regression test in this PR will fail against that pin in CI. This is the expected pre-release state, not a broken branch -- see the DO NOT MERGE note above.

jasonleenaylor:
*[`openspec/changes/archive/2026-07-14-add-opentype-font-features/design.md` line 128 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/1005#-Ozvorz404dRv280HIan:-Ozvorz404dRv280HIao:bw5d6jg) ([raw file](https://github.com/sillsdev/FieldWorks/blob/f3830d7f4c03b98bcfb89b088f9b45970404abfd/openspec/changes/archive/2026-07-14-add-opentype-font-features/design.md#L128)):*
> ```Markdown
> **Alternatives considered:** Remove the `StyleInfo` adapter immediately. Rejected for this change because `SaveToDB_DefaultFontFeatures_RoundTripsThroughRules` failed after removal. Broader LCM dependency alignment could retire the adapter later with the same round-trip tests as the gate.
> 
> **Status update (LT-22351):** The gating condition has been satisfied. liblcm's `BaseStyleInfo.ProcessStyleRules` now loads default `ktptFontVariations` (sillsdev/liblcm#388), `SaveToDB_DefaultFontFeatures_RoundTripsThroughRules` passes through the authoritative `BaseStyleInfo` path, and the `StyleInfo.LoadDefaultFontFeatures` adapter was removed under LT-22351.
> ```

I doubt that keeping the old decisions in a persistent document is actually good for future ai or human understanding. Modifying this to instead show the status quo seems a better option.

jasonleenaylor (approved):
I think one of the tests is really extra, but it covers on the css side something thoroughly proven on the word side, it is possible they would diverge. It seems like we may need to verify that the style dialog handles the LCM changes, but maybe that is an acceptance test that needs to be put on a Jira issue that we add to the sprint?
:lgtm:

> @jasonleenaylor reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1005#-:-P-L21184F-BxNGlxfAE:bjl76sy)* status: 5 of 8 files reviewed, all discussions resolved.

PR #1006, opened 2026-07-10
LT-22610: Fix randomly missing WS labels — RefreshDisplay must always reconstruct

johnml1135:
Fixes [LT-22610](https://jira.sil.org/browse/LT-22610) — "Sometimes the WS labels are not visible (random?)" (regression in 9.3.9).

## What happened

FieldWorks views are drawn from a "root box" — a native object that builds what you see on screen by running a view constructor. The render-speedup work in #724 taught the root box to remember whether anything had changed since it was last built (a `NeedsReconstruct` flag), and taught `SimpleRootSite.RefreshDisplay()` to **skip rebuilding the view whenever that flag said "nothing changed."**

The problem is that the flag only knows about changes the root box itself can see: edits to the data it displays, a stylesheet change, or being handed a new root object. But view constructors also read state the root box *cannot* see. The clearest example is the multi-string fields in the Lexicon Edit detail pane (Lexeme Form, Citation Form, etc.): the list of writing-system rows to show, and the little "Frn" / "Frn-Aud" / "FrIPA" labels next to them, come from writing-system lists held on the managed side, not from the root box's data.

When you change which writing systems a field displays — "Show all writing systems," the per-field Configure Writing Systems dialog, or the Set Up Vernacular Writing Systems dialog — the code updates those lists and then calls `RefreshDisplay()` to redraw the field. After #724, that call looked at `NeedsReconstruct`, saw "nothing changed" (no data edit had happened), and **silently did nothing**. The field kept showing the rows and labels from its previous state: labels missing, or one stale label sitting on the wrong row, with the audio-field sound button positioned against the stale layout.

This also explains every odd detail in the bug report:

- **"I tried the Set Up Vernacular WS dialog and changing things, but that didn't help"** — that dialog's refresh goes through exactly the code path that was being skipped.
- **"Later on it seemed to correct itself"** — the moment you actually edit data in the field, a real data-change notification sets the flag, and the next refresh genuinely rebuilds the view.
- **Restarting FLEx fixed it** — a fresh start rebuilds everything from scratch.
- **"Random"** — the skip path also cleared the "refresh pending" flag without doing the refresh, so a refresh queued while a field was scrolled out of view or being reused could be dropped permanently, depending on timing.

## How this fixes it

`RefreshDisplay()` now always rebuilds the view, as it did in every release before 9.3.9. It is the deliberate "rebuild everything the drastic way" API, called rarely (master refresh, writing-system changes) and precisely when something *outside* the root box's knowledge has changed — so the root box is the wrong thing to ask about whether the rebuild is needed.

Scope notes:

- The now-unused `m_fForceNextRefreshDisplay` / `ShouldReconstructDisplay()` plumbing is removed; the `NotifyDataAccessSemanticsChanged()` / `SetRootBoxDataAccess*` API surface is kept.
- The native performance work from #724 is untouched: the layout guard (`PATH-L1`), the HFONT/color caches, and the Uniscribe shape/analysis caches all remain, so the warm-render wins are preserved. Only the managed refresh short-circuit — the one piece that made a correctness decision from incomplete information — is removed.
- The regression test that pinned the skip behavior now pins the opposite: `RefreshDisplay` must reconstruct even when the root box reports no pending changes.

## Testing

- `SimpleRootSiteTests`: 116/116 pass (includes the updated `RefreshDisplayNeedsReconstruct` tests).
- `WidgetsTests` (LabeledMultiStringView): 19/19 pass.
- `XMLViewsTests`, `DetailControlsTests` (DataTree render baselines): pass.

Manual repro for verification: in Lexicon Edit, right-click a multi-string field → "Show All Writing Systems" (or Configure Writing Systems) and confirm the rows and WS labels update immediately, without editing data first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1006)

### DIFF
--- a/Src/Common/SimpleRootSite/SimpleRootSite.cs
+++ b/Src/Common/SimpleRootSite/SimpleRootSite.cs
@@ -2533,13 +2533,8 @@ namespace SIL.FieldWorks.Common.RootSites
 				return false;
 			}
 
-			// Always reconstruct here, even if the root box reports NeedsReconstruct == false.
-			// That flag only tracks mutations the root box itself can see (PropChanged,
-			// SetRootObjects, OnStylesheetChange). View constructors also read state the
-			// root box cannot see — e.g. the writing-system lists behind the labels in
-			// LabeledMultiStringView — and RefreshDisplay is the only way such changes
-			// reach the screen. Skipping based on NeedsReconstruct left stale writing
-			// system rows/labels in the Lexicon Edit detail pane (LT-22610).
+			// Always reconstruct: view constructors read state the root box can't see
+			// (e.g. WS lists), so skipping on NeedsReconstruct left stale views (LT-22610).
 
 			// Rebuild the display... the drastic way.
 			SelectionRestorer restorer = CreateSelectionRestorer();
--- a/Src/Common/SimpleRootSite/SimpleRootSite.cs
+++ b/Src/Common/SimpleRootSite/SimpleRootSite.cs
@@ -252,7 +252,6 @@ namespace SIL.FieldWorks.Common.RootSites
 		/// <summary>True if we are waiting to do a refresh on the view (will be done when the view
 		/// becomes visible); false otherwise</summary>
 		protected bool m_fRefreshPending = false;
-		private bool m_fForceNextRefreshDisplay;
 		private RefreshPhase m_refreshPhase;
 		private bool m_fRefreshReplayRequested;
 
@@ -1054,7 +1053,6 @@ namespace SIL.FieldWorks.Common.RootSites
 			if (m_rootb?.Site == null)
 				return;
 
-			m_fForceNextRefreshDisplay = true;
 			if (!Visible || FindForm() == null)
 			{
 				m_fRefreshPending = true;
@@ -2535,17 +2533,13 @@ namespace SIL.FieldWorks.Common.RootSites
 				return false;
 			}
 
-			// PATH-L5: Skip the expensive selection save/restore and drawing
-			// suspension when the VwRootBox reports no pending changes.
-			// The root box's NeedsReconstruct flag is set by PropChanged,
-			// OnStylesheetChange, and other mutation paths. When false,
-			// Reconstruct() would be a no-op (PATH-R1), so we can avoid
-			// the managed overhead entirely.
-			if (!ShouldReconstructDisplay())
-			{
-				m_fRefreshPending = false;
-				return false;
-			}
+			// Always reconstruct here, even if the root box reports NeedsReconstruct == false.
+			// That flag only tracks mutations the root box itself can see (PropChanged,
+			// SetRootObjects, OnStylesheetChange). View constructors also read state the
+			// root box cannot see — e.g. the writing-system lists behind the labels in
+			// LabeledMultiStringView — and RefreshDisplay is the only way such changes
+			// reach the screen. Skipping based on NeedsReconstruct left stale writing
+			// system rows/labels in the Lexicon Edit detail pane (LT-22610).
 
 			// Rebuild the display... the drastic way.
 			SelectionRestorer restorer = CreateSelectionRestorer();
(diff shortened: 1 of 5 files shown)


johnml1135:
## Self-review findings and disposition

Ran a recall-biased, high-effort self-review of this branch against `main` (8 finder angles + verification). The LT-22610 fix itself is correct — the confirmed findings were all about the *blast radius* of removing the `NeedsReconstruct` skip globally rather than bugs in the fix's core logic. Cleanups landed in `1283d295c`; the rest is dispositioned below.

### Fixed in `1283d295c`

- **Weakened test coverage** — after the fix, two tests in `SimpleRootSiteTests_RefreshDisplayNeedsReconstruct.cs` asserted identical outcomes, and several still stubbed `NeedsReconstruct`, which production code no longer reads. Renamed the file to `SimpleRootSiteTests_RefreshDisplayReconstruct.cs`, merged the now-identical pair into `RefreshDisplay_AlwaysReconstructs`, and dropped the dead mock setups. 115/115 pass (was 116; the merge removed one redundant test).
- **Docs drift** — `openspec/changes/render-speedup-benchmark/refresh-dirty-flag-audit.md` still described the removed skip as shipped. Added a superseded note there and struck through the `PATH-L5` line in `tasks.md`, both pointing at this PR.

### Considered and deliberately not done: scoping the fix instead of removing the skip globally

The review flagged that removing the skip for all `RefreshDisplay()` callers costs a full native reconstruct per pane on master refresh (Undo/Redo across a save point, Send/Receive merge, style/WS change), plus a second reconstruct on reentrant refresh replay, and suggested restoring the skip everywhere except the known blind spot (`InnerLabeledMultiStringView`/`MultiStringSlice`'s managed-only WS-list state).

I looked into this seriously rather than dismissing it. `RefreshDisplay()` has exactly 5 overrides in the whole tree. Auditing all 5: `GhostStringSlice` and `SandboxBase` never call `base.RefreshDisplay()` (unaffected either way); `XmlBrowseViewBase` calls it but only touches selection/index state, not view-constructor state (safe under the old skip). But **`TitleContentsPane.RefreshDisplay()`** *also* mutates managed-only VC state (`m_vc.SetupWritingSystemsForTitle()`, which rebuilds `m_writingSystems`/`m_WsLabels`) immediately before calling `base.RefreshDisplay()` — a second blind spot the review itself didn't surface.

That's why the skip stays out: the overrides only bound *known* blind spots. LT-22610 was already an intermittent, hard-to-reproduce bug caused by exactly one such unscoped optimization; a scoped fix is only as safe as the completeness of the audit behind it, and confirming there's no *third* call site that mutates VC-only state and calls `RefreshDisplay()` directly (not through an override) would require auditing every one of the ~20+ call sites, not just the 5 overrides. That's real, valuable follow-up work, but it deserves its own ticket and test pass rather than being folded into this bug-fix PR.

### Deferred to the same follow-up ticket

- **Dead native `NeedsReconstruct` flag** (`Src/views/VwRootBox.h`) — no managed code reads it anymore. Removing it is a native C++ change requiring its own rebuild/verification, and it probably wants to happen alongside deciding whether a scoped skip is worth reintroducing at all. The follow-up starts from the two known blind spots (`InnerLabeledMultiStringView`, `TitleContentsPane`).

jasonleenaylor (approved):
:lgtm:

Make sure the final commit comment does not mention any terminology from the planning documents that isn't part of long term FieldWorks domain terminology (e.g. PATH-L5, and PATH-L1) these intermediate planning documents do not carry the same weight as the commit comment. They will get cleaned up at some future date and someone shouldn't have to research the commit contents to interpret and understand the commit message. Also since we spun off a new Jira issue to capture the remainder of the original issue report make sure that we scope the commit message to what is being done here and don't overclaim its resolution of the original issue.

> @jasonleenaylor partially reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1006#-:-Oxgk-xl8BLcArg9VeZX:b33rhtg)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #1007, opened 2026-07-13
This fixes LT-22616: Word Cat contents change when bundle selected

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22616.  The root problem was that SandboxBase and InterlinVc were making different guesses about the word cat and gloss, which confused the user.  The general principle is that when you first click on a word in the interlinear text, the sandbox should show whatever was showing in the interlinear text.  The word cat and gloss should only change after the user makes a change to the sandbox.

I also added code to guess the word gloss and category whenever the morphemes change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1007)

Devin's analysis:

The interlinear display (InterlinVc) and the editing sandbox (SandboxBase) had independent implementations for guessing word glosses and parts of speech from morph bundles. This caused inconsistencies—the display could show one guess while the sandbox applied a different one.
This PR:
Extracts the private guess methods from an inner class of InterlinVc into internal static methods at the InterlinVc class level, returning richer types (ILexSense, IPartOfSpeech) instead of just display strings.
Adds SandboxBase.GuessWordCatAndGloss which delegates to those same InterlinVc methods, ensuring consistent guesses.
Renames CopyLexEntryInfoToMonomorphemicWordGlossAndPos → CopyLexEntryInfoToWordGlossAndPosIfEmpty and adjusts CopySenseToWordGloss to accept a real HVO directly.
After morpheme updates in HandleSelectMorphComboItem, calls CopyLexEntryInfoToWordGlossAndPosIfEmpty to refresh guesses.

### DIFF
--- a/Src/LexText/Interlinear/InterlinVc.cs
+++ b/Src/LexText/Interlinear/InterlinVc.cs
@@ -2311,6 +2311,101 @@ namespace SIL.FieldWorks.IText
 			vwenv.CloseParagraph();
 		}
 
+		internal static ILexSense GuessWordSense(IWfiAnalysis wa)
+		{
+			if (wa.MeaningsOC.Count != 0)
+			{
+				return null;
+			}
+			// This is modeled after SandboxBase.SyncMonomorphemicGlossAndPos.
+			ILexSense rootSense = null;
+			foreach (IWfiMorphBundle morphBundle in wa.MorphBundlesOS)
+			{
+				// Get the sense for morphBundle.
+				ILexSense sense = morphBundle.SenseRA;
+				if (sense == null && morphBundle.MorphRA != null)
+				{
+					ILexEntry lexEntry = morphBundle.MorphRA.Owner as ILexEntry;
+					if (lexEntry != null && lexEntry.SensesOS.Count > 0)
+					{
+						sense = lexEntry.SensesOS[0];
+					}
+				}
+				if (sense?.Gloss?.BestAnalysisAlternative == null)
+					continue;
+				// Consider the sense's gloss.
+				IMoMorphSynAnalysis msa = sense.MorphoSyntaxAnalysisRA;
+				bool fStem = msa is IMoStemMsa;
+				// If we have only one morpheme, treat it as the stem from which we will copy the gloss.
+				// otherwise, use the first stem we find, if any.
+				if ((fStem && rootSense == null) || wa.MorphBundlesOS.Count == 1)
+					rootSense = sense;
+			}
+			return rootSense;
+		}
+
+		internal static ITsString GuessWordPOSName(IAnalysis analysis)
+		{
+			IPartOfSpeech lexPOS = GuessWordPOS(analysis);
+			if (lexPOS == null)
+				return null;
+			if (lexPOS.Abbreviation.BestAnalysisAlternative.Length > 0)
+				return lexPOS.Abbreviation.BestAnalysisAlternative;
+			return lexPOS.Name.BestAnalysisAlternative;
+		}
+
+		internal static IPartOfSpeech GuessWordPOS(IAnalysis analysis)
+		{
+			var wa = analysis.Analysis;
+			if (wa.CategoryRA != null)
+			{
+				return null;
+			}
+			// This is modeled after SandboxBase.SyncMonomorphemicGlossAndPos.
+			IPartOfSpeech stemPOS = null;
+			IPartOfSpeech derivedPOS = null;
+			foreach (IWfiMorphBundle morphBundle in wa.MorphBundlesOS)
+			{
+				// Get the sense for morphBundle.
+				ILexSense sense = morphBundle.SenseRA;
+				if (sense == null && morphBundle.MorphRA != null)
+				{
+					if (morphBundle.MorphRA.Owner is ILexEntry lexEntry && lexEntry.SensesOS.Count > 0)
+					{
+						sense = lexEntry.SensesOS[0];
+					}
+				}
+				if (sense == null)
+					continue;
+				// Consider the sense's part of speech.
+				IMoMorphSynAnalysis msa = sense.MorphoSyntaxAnalysisRA;
+				bool fStem = msa is IMoStemMsa;
+				if (fStem)
+				{
+					IPartOfSpeech POS = (msa as IMoStemMsa).PartOfSpeechRA;
+					if (POS != stemPOS && stemPOS != null)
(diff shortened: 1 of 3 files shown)


jasonleenaylor:
*[`Src/LexText/Interlinear/SandboxBase.cs` line 4264 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-OxSEn5sBGaqhjyBaVLR:-OxSEn5tCo4_t3T5JYvX:b-1ns5bs) ([raw file](https://github.com/sillsdev/FieldWorks/blob/66112c76bead06069c3e0105bdca05a261d20c51/Src/LexText/Interlinear/SandboxBase.cs#L4264)):*
> ```Smalltalk
> 			{
> 				m_editMonitor.DoPendingMorphemeUpdates();
> 				CopyLexEntryInfoToMonomorphemicWordGlossAndPos(true);
> ```

If a user has manually chosen a word gloss and word category will this overwrite them when they choose a different msa? I'm not sure that is desired if that is the case.

jtmaxwell3:
@jasonleenaylor CopyLexEntryInfoToMonomorphemicWordGlossAndPos is poorly named because it only does the copy if the word gloss and/or category don't already exist.  It should be something like GuessWordGlossAndPos.

jasonleenaylor:
*[`Src/LexText/Interlinear/InterlinVc.cs` line 2402 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-OxfeROtBj5LMzQ9etsI:-OxfeROtBj5LMzQ9etsJ:b-gkdr28) ([raw file](https://github.com/sillsdev/FieldWorks/blob/c8657d286ab671597d2c30d4f3c1f8a3492961fe/Src/LexText/Interlinear/InterlinVc.cs#L2402)):*
> ```Smalltalk
> 				}
> 			}
> 			if (stemPOS == null)
> ```

For clear developer understanding of this statement a comment would be good. With domain knowledge in mind I think this is probably correct, but just reading the code I would wonder "What if we have a derivedPOS and no stemPOS, shouldn't we just return the derived?"

jasonleenaylor (commented):
I had one small comment request otherwise this change seems sound to me.
We want to make sure this is really hitting the felt user need of consistency between the focus box and the other sandbox views. 
It seems like it does that, hopefully this will be an acceptable way.

> @jasonleenaylor reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-:-OyF7MKb2MSYvJgg0Bv4:b-dc1pgj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

jtmaxwell3 (commented):
> @jtmaxwell3 made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1007)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

___
*[`Src/LexText/Interlinear/InterlinVc.cs` line 2402 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-OxfeROtBj5LMzQ9etsI:-OyJhvuVDGCkVir48LYa:b-q846q4) ([raw file](https://github.com/sillsdev/FieldWorks/blob/c8657d286ab671597d2c30d4f3c1f8a3492961fe/Src/LexText/Interlinear/InterlinVc.cs#L2402)):*

For clear developer understanding of this statement a comment would be good. With domain knowledge in mind I think this is probably correct, but just reading the code I would wonder "What if we have a derivedPOS and no stemPOS, shouldn't we just return the derived?"
</blockquote></details>

I didn't write this code, I just moved it.  I don't feel comfortable adding a comment about what it is supposed to do.
___
*[`Src/LexText/Interlinear/SandboxBase.cs` line 4264 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-OxSEn5sBGaqhjyBaVLR:-OyJiGMf1L20muqwPMcF:b6iusgp) ([raw file](https://github.com/sillsdev/FieldWorks/blob/66112c76bead06069c3e0105bdca05a261d20c51/Src/LexText/Interlinear/SandboxBase.cs#L4264)):*

If a user has manually chosen a word gloss and word category will this overwrite them when they choose a different msa? I'm not sure that is desired if that is the case.
</blockquote></details>

I renamed the function to make it clear that it doesn't overwrite the word gloss and word category.

jasonleenaylor:
*[`Src/LexText/Interlinear/InterlinVc.cs` line 2364 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-OyJuSnM1W91LK2NzqP3:-OyJuSnM1W91LK2NzqP4:b-j2vvph) ([raw file](https://github.com/sillsdev/FieldWorks/blob/c8657d286ab671597d2c30d4f3c1f8a3492961fe/Src/LexText/Interlinear/InterlinVc.cs#L2364)):*
> ```Smalltalk
> 				return null;
> 			}
> 			// This is modeled after SandboxBase.SyncMonomorphemicGlossAndPos.
> ```

```
// Guess the word category from the morphemes' MSAs, matching
// SyncMonomorphemicGlossAndPos so the display and the sandbox agree:
//   - if there is more than one stem and they have different parts of
//     speech, give up (return null);
//   - if there is more than one derivational affix (DA), give up;
//   - if there is no stem, give up — a derived-to category is only
//     meaningful relative to the stem it attaches to, so a lone DA is
//     not enough to guess from;
//   - otherwise use the DA's 'to' POS if there is one, else the stem's POS.
//     (we don't insist that the DA's 'from' POS matches the stem)
```



_Code quote:_
```Smalltalk
// This is modeled after SandboxBase.SyncMonomorphemicGlossAndPos.
```

jasonleenaylor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1007#-:-OyZefbQEBKNz4zzOqje:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1007)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #1011, opened 2026-07-14
Fix LT-22614: FLEx does not restart after a Send/Receive merge

jasonleenaylor:
The .NET 4.8 retarget changed the hash WCF uses in named-pipe names, so FLEx and FLEx Bridge (net462) could no longer find each other's endpoints and the "changes received" signal was silently lost.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1011)

### DIFF
--- a/Src/Common/FwUtils/FLExBridgeHelper.cs
+++ b/Src/Common/FwUtils/FLExBridgeHelper.cs
@@ -8,7 +8,9 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Security.Cryptography;
 using System.ServiceModel;
+using System.Text;
 using System.Threading;
 using SIL.LCModel;
 using SIL.LCModel.Utils;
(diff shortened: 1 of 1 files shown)


jtmaxwell3 (approved):

Src/Common/FwUtils/FLExBridgeHelper.cs
```diff
@@ -324,6 +326,27 @@ public static bool LaunchFieldworksBridge(string projectFolder, string userName,
...
+		private static string GeneratePipeId(string projectFolder, string command)
+		{
+			using (var sha = SHA256.Create())
+			{
+				// SHA256.Create() is stable across processes and FIPS-safe (unlike GetHashCode or MD5).
+				var digest = sha.ComputeHash(Encoding.UTF8.GetBytes(projectFolder ?? string.Empty));
+				// First 8 bytes as hex: unique enough, and URI-safe where Base64 is not.
+				// (On .NET 5+ this would be Convert.ToHexString(digest.AsSpan(0, 8)).)
+				var token = BitConverter.ToString(digest, 0, 8).Replace("-", string.Empty);
+				return "SendReceive" + token + command;
```
hahn-kev, line 346: if we know what "long" is defined as then should we clamp this to not go past that length? Not knowing anything about `command` that seems like it could trigger the same issue here. Maybe we just throw at the start if command is long? To catch this issue before it goes into production in the future?

jasonleenaylor: 'command' comes from a set of constants in the repo and the longest current command gives us a 52 character margin, so we could throw but it seems like overkill. Plus we're not ever going to add any new commands as the future of collaboration is Harmony not FB.

PR #1018, opened 2026-07-21
Fix LT-22611: Allow Bulk Edit Inflection Features for affixes

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22611.  We can't just set the inflection features of the MSA since it may be shared with other lexical senses.  So we make a copy of the MSA and set its inflection features.  We then see whether there is an existing MSA that matches it.  If there is, then we use it and delete the copy.

I added code for deleting newFsTarget if it was created by FillInBlanks and it isn't used elsewhere.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1018)

### DIFF
--- a/Src/FdoUi/InflectionFeatureEditor.cs
+++ b/Src/FdoUi/InflectionFeatureEditor.cs
@@ -378,8 +378,42 @@ namespace SIL.FieldWorks.FdoUi
 						{
 							newFsTarget = FillInBlanks(fsTarget, ls);
 						}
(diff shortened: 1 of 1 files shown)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1018)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #1024, opened 2026-07-23
Fix LT-22632: Show # of changed analyses in Run Tests report

jtmaxwell3:
This implements https://jira.sil.org/browse/LT-22632.  It adds a column that reports the number of analyses that changed since the last time that a word was parsed.  This is only meaningful if Updates Word Analyses is disabled, so it hides the column if it is enabled.  This is useful if you are trying to speed up a grammar without changing any of the analyses.  Then you can
1) Parse a corpus to establish a baseline.
2) Disable Updates Word Analyses.
3) Make changes to the grammar.
4) Invoke Run tests on the corpus.
5) Verify that the number of changed analyses is zero for the corpus.
6) Verify that the corpus parses faster.
7) Go to step 3.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1024)

### DIFF
--- a/Src/LexText/ParserUI/ParserUIStrings.resx
+++ b/Src/LexText/ParserUI/ParserUIStrings.resx
@@ -205,6 +205,12 @@
   <data name="ksNumAnalysesToolTip" xml:space="preserve">
     <value>The number of analyses produced by the parser</value>
   </data>
+  <data name="ksNumChangedAnalyses" xml:space="preserve">
+    <value>Num Changed Analyses</value>
+  </data>
+  <data name="ksNumChangedAnalysesToolTip" xml:space="preserve">
+    <value>The number of analyses that have changed since the last parse</value>
+  </data>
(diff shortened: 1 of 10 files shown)


Src/LexText/ParserUI/ParserUIStrings.resx
```diff
@@ -286,6 +292,12 @@
   <data name="ksTotalAnalysesToolTip" xml:space="preserve">
     <value>The total number of analyses in the words parsed</value>
   </data>
+  <data name="ksTotalChangedAnalyses" xml:space="preserve">
+    <value>Num Changed Analyses</value>
+  </data>
+  <data name="ksTotalChangedAnalysesToolTip" xml:space="preserve">
+    <value>The total number of changed analyses since the last parse</value>
```
papeh: nit: "The total number of analyses that have changed since the last parse" would be more consistent with the other new tooltip.

papeh (approved):

PR #1025, opened 2026-07-23
Fix LT-22633

AndyBlack:
We have two places where we display information in a web browser. We allow users to invoke a Find dialog so they can more easily find what they are looking for.  There were two places where invoking the Find dilaog could cause a crash.  The reporter for LT-22633 found one of them.

The fix in GeneratedHtmlViewer.cs is for the case the reporter found.

The fix in TryAWordDlg.cs is for the other case I found. The simplest solution for it is to only allow the user to invoke the Find dialog when TryAWord is showing a parsing result page.  There is no need to let the user look for something with either the initial message page or the while trrying page.

There was a typo in the InitialDocument.htm title which is fixed (not that this is crucial...).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1025)

### DIFF
--- a/DistFiles/Language Explorer/Configuration/Words/Analyses/TraceParse/InitialDocument.htm
+++ b/DistFiles/Language Explorer/Configuration/Words/Analyses/TraceParse/InitialDocument.htm
@@ -1,6 +1,6 @@
 <html>
    <head>
-	  <title>The results of tyring a word will appear here</title>
+	  <title>The results of trying a word will appear here</title>
    </head>
    <body>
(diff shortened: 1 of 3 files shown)


thejambi (approved):
Changes look good!
:lgtm:

> @thejambi reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1025#-:-OyFRSNe5MzzGumfpkAV:b4ttmsl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on AndyBlack).

PR #1029, opened 2026-07-24
LT-22638: Show OpenType feature names and multi-valued variants

jasonleenaylor:
## Summary
Fixes [LT-22638](https://jira.sil.org/browse/LT-22638). Makes OpenType font
features usable in Font Options: features show human-readable names from the
font (or a registered-feature catalog), and character variants with multiple
named alternates are selectable instead of a single On/Off toggle — parity with
the existing Graphite experience, with no change to the dropdown menu, so both
the Writing System and Styles font dialogs inherit it.

## What changed
- **New `OpenTypeFontFeatureInfoReader` (FwUtils)** — parses GSUB/GPOS feature
  lists, `featureParams`, and the `name` table via a table-source delegate (GDI
  `GetFontData` in the app, font-file bytes in tests). Bounds-checked; malformed
  fonts degrade to tag-only records. Adapted from Paratext's `OpenTypeFeatures.Ttf`.
- **New `OpenTypeFeatureCatalog` (FwUtils)** — classifies registered features
  (hidden / default-on) and supplies English names, seeded from Paratext and
  audited against the OpenType registry (`dlig` visible, `aalt` hidden, `kern`
  default-on).
- **Rewrote the OpenType provider in `FontFeaturesButton`** — character variants
  with named options become "None + option" submenus stored as `cvNN=k`; labels
  resolve font-supplied → resx → catalog → numbered fallback; default-on features
  initialize enabled; hidden features filtered.
- **resx** — added None / numbered-set fallbacks; removed dead labels (`aalt`,
  `ccmp`) and entries orphaned by the rewrite (`ss01`–`ss05`, `ValueOff`/`ValueOn`).
- **Docs** — `Docs/opentype-font-features.md` updated.

Storage, rendering, and CSS/Word export are untouched — the renderer-neutral
`tag=value` string already carried multi-values.

## Testing
- Reader verified against real **Charis SIL 6.200** (`cv43` "Capital Eng" + 3
  options, `ss01` "Single-story a and g", GPOS `mark`/`mkmk`) and **Scheherazade
  New** (`cv70` "Damma").
- Synthetic malformed-table robustness tests.
- Provider tests: multi-value cv, binary fallback, default-on init, label
  priority, hidden filtering, `cv43=2` round-trip, resx↔catalog consistency.
- Green: FwUtilsTests 33, FwCoreDlgControlsTests 17, FwCoreDlgsTests 19. Coverage:
  catalog/info 100%, reader 86%.
- Manual acceptance passed (WS setup → Font tab → Charis → Font Features).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1029)


Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
```diff
@@ -0,0 +1,309 @@
...
+			label = LookupName(names, labelNameId);
+
+			if (numNamedParameters == 0 || numNamedParameters > MaxNamedParameters)
+				return;
+			var resolved = new List<string>(numNamedParameters);
+			for (var i = 0; i < numNamedParameters; i++)
+			{
+				var option = LookupName(names, firstParamNameId + i);
+				if (!string.IsNullOrEmpty(option))
+					resolved.Add(option);
```
johnml1135: Preserve undecodable `cvNN` option slots. Dropping one renumbers later values, so selecting "option 2" can persist the wrong glyph. Use a numbered placeholder and add a missing-option-name test.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
```diff
@@ -0,0 +1,229 @@
...
+				["pcap"] = Visible("Lowercase to Petite Capitals", OpenTypeFeatureDefaultState.Off),
+				["pkna"] = Visible("Proportional Kana", OpenTypeFeatureDefaultState.Off),
+				["pnum"] = Visible("Proportional Numbers", OpenTypeFeatureDefaultState.Off),
+				["pref"] = Hidden("Pre Base Forms"),
+				["pres"] = Hidden("Pre Base Substitutions"),
+				["pstf"] = Hidden("Post Base Forms"),
+				["psts"] = Hidden("Post Base Substitutions"),
+				["pwid"] = Visible("Proportional Width"),
+				["qwid"] = Visible("Quarter Widths", OpenTypeFeatureDefaultState.Off),
+				["rand"] = Visible("Randomize", OpenTypeFeatureDefaultState.On),
```
johnml1135: The catalog's default-on set exceeds the four tags specified by the proposal, design, tasks, and shipped docs. Align it with verified Uniscribe defaults or update the spec/docs deliberately; confirm `kern`, then pin the exact set in a test.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
```diff
@@ -0,0 +1,229 @@
...
+				["rphf"] = Hidden("Reph Form"),
+				["rtbd"] = Visible("Right Bounds", OpenTypeFeatureDefaultState.Off),
+				["rtla"] = Hidden("Right to Left Alternates"),
+				["rtlm"] = Hidden("Right to Left mirrored forms"),
+				["ruby"] = Visible("Ruby Notational Forms", OpenTypeFeatureDefaultState.Off),
+				["rvrn"] = Hidden("Required Variation Alternates"),
+				["salt"] = Visible("Stylistic Alternatives", OpenTypeFeatureDefaultState.Off),
+				["sinf"] = Visible("Scientific Inferiors", OpenTypeFeatureDefaultState.Off),
+				["smcp"] = Visible("Lowercase to Small Capitals", OpenTypeFeatureDefaultState.Off),
+				["size"] = Visible("Optical size", OpenTypeFeatureDefaultState.On),
```
johnml1135: `size` is optical-sizing metadata, not a meaningful `size=1` toggle. Hide it under the spec's "otherwise not user-configurable" rule.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		private static readonly Dictionary<string, OpenTypeFeatureCatalogEntry> s_entries =
+			new Dictionary<string, OpenTypeFeatureCatalogEntry>(StringComparer.Ordinal)
+			{
+				["aalt"] = Hidden("Access All Alternates"),
+				["abvf"] = Hidden("Above Base Forms"),
+				["abvm"] = Hidden("Above Base Mark"),
+				["abvs"] = Hidden("Above Base Substitutions"),
+				["afrc"] = Visible("Vertical Fractions", OpenTypeFeatureDefaultState.Off),
```
johnml1135, line 76: Move these user-visible English names to `FwUtilsStrings.resx`; C# fallbacks bypass Crowdin and violate `AGENTS.md`. Preserve or deliberately replace the translated keys removed from `FwCoreDlgControls.resx`.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
```diff
@@ -0,0 +1,309 @@
...
+		}
+
+		private static bool IsRicher(OpenTypeFontFeatureInfo candidate, OpenTypeFontFeatureInfo existing)
+		{
+			var candidateScore = (candidate.FontSuppliedLabel != null ? 1 : 0) + candidate.Options.Count;
+			var existingScore = (existing.FontSuppliedLabel != null ? 1 : 0) + existing.Options.Count;
+			return candidateScore > existingScore;
+		}
+
+		private static byte[] SafeGet(Func<string, byte[]> tableSource, string tag)
```
johnml1135, line 119: Keep graceful degradation, but trace parse failures. Bare catches make malformed fonts indistinguishable from fonts without features and remove the prior diagnostic path; route failures through the existing trace switch or an equivalent FwUtils source.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
```diff
@@ -1075,35 +1124,29 @@ internal static class OpenTypeFontFeatureReader
...
 			{
 				var cacheKey = FontFeatureCacheKey.FromHdc(hdc);
 				lock (s_cacheLock)
 				{
-					string[] cachedTags;
-					if (s_featureTagCache.TryGetValue(cacheKey, out cachedTags))
-						return cachedTags.ToArray();
+					OpenTypeFontFeatureInfo[] cached;
+					if (s_featureCache.TryGetValue(cacheKey, out cached))
+						return cached;
```
johnml1135, line 1164: Return a copy or immutable collection. Returning the cached array exposes shared mutable state across callers, unlike the previous `.ToArray()` behavior.

johnml1135 (commented):
Four merge issues; details are inline:

1. Preserve `cvNN` option positions when labels cannot decode.
2. Align default-on features with verified renderer behavior, spec, and docs; hide `size`.
3. Move user-visible catalog names to `.resx`.
4. Add default-on persistence tests: unset stays absent; unchecked writes `liga=0`.

**Next:** address or explicitly defer each inline finding.

johnml1135:
Three follow-ups:

1. Align visible defaults with verified Uniscribe behavior; confirm `kern` and document the policy.
2. Move FieldWorks-authored names to `.resx`; track UI-language wiring separately. **Not known:** whether bundled fonts contain reusable translations.
3. File the planned UI follow-up for a selection summary and grouped live preview. LibreOffice is a license-compatible reference, subject to normal review.

**Next:** decide which items belong in this PR versus Jira.

johnml1135:
Consolidated into the [follow-up summary](https://github.com/sillsdev/FieldWorks/pull/1029#issuecomment-5103556284).

johnml1135:
Consolidated into the [follow-up summary](https://github.com/sillsdev/FieldWorks/pull/1029#issuecomment-5103556284).

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
```diff
@@ -1006,28 +999,93 @@ public string GetFeatureTag(int featureId)
...
 			{
-				defaultValue = 0;
+				var tag = GetFeatureTag(featureId);
+				OpenTypeFontFeatureInfo info;
+				m_infoById.TryGetValue(featureId, out info);
+
+				if (info != null && info.Options.Count > 0)
+				{
+					// value 0 = "None", value i = the i-th named character-variant option.
+					var optionCount = Math.Min(info.Options.Count, Math.Max(0, maxValues - 1));
```
johnml1135, line 1035: OpenType character variants after option 31 are silently unreachable because this reuses Graphite's `kMaxValPerFeat`. Raise the OpenType limit or expose truncation explicitly; add a test for more than 31 named options.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
```diff
@@ -0,0 +1,309 @@
...
+
+			foreach (var tableTag in s_layoutTables)
+			{
+				var table = SafeGet(tableSource, tableTag);
+				if (table == null)
+					continue;
+				foreach (var info in ReadFeatureList(table, names))
+				{
+					OpenTypeFontFeatureInfo existing;
+					if (!byTag.TryGetValue(info.Tag, out existing) || IsRicher(info, existing))
```
johnml1135, line 104: Same-tag feature records are merged across scripts by "richest" metadata, so a Cyrillic writing system can receive Latin labels and values. Prefer the writing system's script, then `DFLT`; otherwise document this as an accepted limitation.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
```diff
@@ -1075,35 +1124,29 @@ internal static class OpenTypeFontFeatureReader
...
-				var tags = new SortedSet<string>(StringComparer.Ordinal);
-				foreach (var table in s_layoutTables)
-				{
-					var tableData = s_tableReader(hdc, table);
-					if (tableData != null)
-						ReadFeatureList(tableData, tags);
-				}
-				var discoveredTags = tags.ToArray();
+				var discovered = OpenTypeFontFeatureInfoReader
+					.Read(tag => s_tableReader(hdc, MakeTableTag(tag))).ToArray();
```
johnml1135: Snapshot `s_tableReader` inside `s_cacheLock` before invoking it. Tests write this mutable static under the lock but production reads it outside, allowing parallel fixtures to observe the wrong reader.

johnml1135 (commented):
Three follow-up findings; details are inline:

1. OpenType character variants after option 31 are unreachable.
2. Same-tag records can expose one script's labels and values in another writing system.
3. `s_tableReader` is read outside the lock guarding test-time replacement.

**Next:** fix or explicitly accept each limitation.

johnml1135:
Correction incorporated into the [follow-up summary](https://github.com/sillsdev/FieldWorks/pull/1029#issuecomment-5103556284).

johnml1135:
Correction incorporated into the [follow-up summary](https://github.com/sillsdev/FieldWorks/pull/1029#issuecomment-5103556284).

johnml1135:
Correction incorporated into the [follow-up summary](https://github.com/sillsdev/FieldWorks/pull/1029#issuecomment-5103556284).

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
jasonleenaylor: Fixed. Slots keep their position now and a name that won't decode stays null for the UI to number, since that index is the value we persist and the doc comment on Options already promised it.

I surveyed 192 font files here to see how this behaves in practice: Windows fonts, the per-user store, DistFiles and Downloads. 477 cvNN features declare named options across 15 fonts and not one of them has a slot the reader can't resolve, so nobody is hitting this today. Worth fixing anyway since a user can install any font, but it isn't a live corruption.

The same survey changed what I did about the all-fail case. 420 of those 477 features declare exactly one option, so if that single name fails to decode a "None / Option 1" dropdown would be worse than the checkbox we show now. I kept the fallback to a binary toggle when none of the names decode, and only preserve positions when at least one does.

Two tests: a middle slot missing keeps `{"First", null, "Third"}`, and no slot decoding leaves the feature binary.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
jasonleenaylor: rand, chws, cpsp and halt are off now, size is hidden per your other comment, and `DefaultOnFeatures_AreOnlyTheDocumentedFour` pins the set. `CommonlyDefaultOnFeatures_AreDefaultOn` checks one tag per TestCase, so it could only ever catch a tag going missing, never a fifth being added, which is how these got in.

Chasing this down turned up something more interesting than the five tags, and it changed my mind twice.

`IsDefaultOn` is the one place this provider isn't equivalent to the Graphite one. `GraphiteFontFeatureProvider.GetFeatureValues` gets `defaultValue` straight out of the font, and the same font table drives the Graphite shaper, so the checkbox and the renderer can't disagree. OpenType has nowhere to put that: a GSUB/GPOS FeatureRecord has no default-value field, and only cvNN/ssNN carry featureParams, which hold labels and not defaults. So the OpenType provider fills the same out-parameter from a static table, and the two providers share a signature that hides the difference. Under Graphite a wrong default is impossible; here it's an unverified claim about Uniscribe.

That's the actual defect. The five tags were a symptom.

I went to the OpenType registry expecting it to be the equivalent of the Graphite font table, and it isn't. It recommends cpsp and rand be on by default, and chws and halt for horizontal CJK layout in applications that don't implement CLREQ/JLREQ/KLREQ, which we don't. So on the registry's own wording the nine-tag set was closer than the four-tag set. But the registry is telling an application to *apply* a feature, which here means writing `tag=1`, and marking a tag default-on does the opposite: the string stays silent and the feature is never requested. Marking cpsp default-on doesn't honour that recommendation, it prevents it.

Marking a tag on only sets the initial checkbox for a feature the writing system doesn't mention. Marked on, cpsp, rand, chws and halt showed as enabled while the text was unaffected, and a user's first click wrote `tag=0`, so the gesture meant to enable them disabled them. Marked off, the first click writes `tag=1`, a no-op if Uniscribe applies the feature anyway, and `tag=0` is one click further.

chws and halt couldn't both have been on anyway. Each is mutually exclusive with the other and with every horizontal glyph-width feature, and the catalog has no per-script state to choose between them.

Which brings me back to your wording. "Align it with verified Uniscribe defaults" is the only phrasing that closes this loop, and it's a stronger point than I first read it as. kern is in the same position as the other three, just better-supported by shaping convention: `UniscribeSegment.cpp:792` parses the feature string and `:835` switches to `ShapePlaceRunWithOpenType` with an explicit record list once anything is set, so whether kern survives that switch is measurable and unmeasured. `ShapeText_OpenTypeFeatureToggleChangesShapingData` would answer it but skips itself, because Times New Roman gave no deterministic liga delta. That's LT-22774.

`Docs/opentype-font-features.md` now carries all of this, and the pinning test points at it, so the next person who wants to add cpsp back reads the reason first.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
jasonleenaylor: Hidden. The spec is more definite than "not meaningful": the size Feature table must not reference any lookups, and its featureParamsOffset points at five uint16 values. There are no lookups for `size=1` to trigger, so the toggle could never have done anything. Zero of the 192 fonts I surveyed declare `size` at all, so nobody has met the dead checkbox either.

One correction on the citation. `spec.md:41` reads "required for script shaping AND not user-configurable", and `size` isn't shaping-required, so that scenario doesn't reach it. The clause that does is the requirement above it at `:34`, which scopes discovery to user-configurable tags.

Worth flagging so it doesn't look like a contradiction later: the registry does say "This feature should be active by default" for `size`. That's telling the application to honour the optical-size data when the user picks a point size, not to show a ticked checkbox. Hiding it and honouring it are different questions, and the second one isn't in scope here.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
jasonleenaylor: Fixed, 57 keys added, so the resx now has one for each of the 71 visible features instead of 14. You're right on the rule and `AGENTS.md:15` states it outright: keep localization in .resx, don't hardcode translatable UI strings.

I put them in `FwCoreDlgControls.resx` rather than `FwUtilsStrings.resx`. Three reasons. The resx-first lookup already exists at `FontFeaturesButton.cs:1010`, so the keys are consumed where they land with no new plumbing. The 14 that were already there are in that file, so splitting would leave feature labels in two places. And these are UI labels, which sit oddly in a utility assembly. `crowdin.json:20` globs `Src/**/*.resx`, so either destination is a Crowdin source with no config change. Say the word if you'd still rather have them in FwUtils and I'll move them.

The 54 hidden names stay in C#. They're never shown, so they aren't translatable UI strings.

One thing I'm not claiming: nothing yet stops the next added feature from having no key and quietly falling back to the C# name. All 71 resolve today because I generated the keys from the catalog, and `ResxOpenTypeFeatureLabels_MapToVisibleFeatures` guards the other direction, that a key maps to a visible feature. The missing direction, that a visible feature has a key, is in LT-22774.

On the nine removed keys, all deliberate:

- `kstidOpenTypeFeatureValueOff` and `_ValueOn` had to go, and localizing them would have been a bug. `:763-781` lowercases those strings and matches them against "on" and "off" to decide whether a feature is binary. A translated "Ein"/"Aus" fails that match, and the checkbox becomes a submenu.
- `_aalt` and `_ccmp` had to go too. They're hidden now, and `ResxOpenTypeFeatureLabels_MapToVisibleFeatures` fails if a key maps to a hidden feature.
- `_ss01` through `_ss05` are replaced by `kstidOpenTypeFeatureStylisticSet`, "Stylistic Set {0}", so they're still translatable as one format string instead of five literals.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
jasonleenaylor: Fixed, all three, and they route through `FontFeatureSettings.DiagnosticsSwitch`, the equivalent FwUtils source you pointed at. It's internal and the reader is in the same assembly, so nothing new was needed.

The three are the table read at `:119`, name-record decoding at `:267`, and the Mac Roman encoding fallback at `:336`. Split by severity rather than all warnings: the table read and the encoding fallback are warnings, name-record decoding is info. That one runs per name record, and the comment at `:261` says non-decodable platforms are expected and fall back by design, so logging each as a warning would bury the two that mean something.

The Mac Roman one is worth a warning on its own account. If codepage 10000 is unavailable we silently fall back to ASCII, which mangles any non-ASCII Mac name rather than failing.

Degradation is unchanged. `ThrowingTableSource_IsSwallowed` and `UndecodablePlatform_FallsBackToNoLabel` both still pass, so discovery still degrades instead of throwing; it just says so now.

Expected volume, from the 192-font survey: no font produced a name slot the reader couldn't resolve, so these should be silent in practice and a message should mean something real rather than being background noise.

The truncation case in `FontFeaturesButton` traces too, but through `FontFeatures.OpenType` rather than the FwUtils switch, since `DiagnosticsSwitch` is internal to FwUtils and that code is in FwCoreDlgControls. That one belongs with your `kMaxValPerFeat` comment.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
jasonleenaylor: Fixed, and you were right about the comparison. I went and looked at `GetFeatureTags` on main before this PR: it did `return cachedTags.ToArray()` on the hit path and `return discoveredTags.ToArray()` on the miss path. So the copy wasn't missing, it was removed, which makes this a regression of an existing safeguard rather than hardening that was never there. That changed how I treated it.

The cache stores a `ReadOnlyCollection` now and both paths hand back that same wrapper. Two reasons for wrapping once at insert instead of copying on the way out: no allocation on the hit path, which is where this is called from menu building, and a caller who casts to `IList<T>` gets a `NotSupportedException` rather than a silent success. `IReadOnlyList<T>` on its own was only a compile-time promise, so a cast back to the array type could write through to the cache and change what every later caller saw for that font.

One change beyond what you asked for. The miss path used to return its own `discovered` even when another thread had won the insert race, so two callers could hold different equal instances. It now returns whatever is in the cache, so they share one. Happy to take that back out if you'd rather keep the diff literal.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
jasonleenaylor: Taking your second option, exposing it rather than raising it, and I'll explain why since you offered both.

I surveyed 192 font files: Windows fonts, the per-user store, DistFiles and Downloads. 477 cvNN features declare named options across 15 fonts, and they break down as 420 with one option, 40 with two, 16 with three, and one with four. The maximum anywhere is four, in ScheherazadeNew cv82. Nothing is above 31, and nothing is close.

The spec ties that number to something physical: numNamedParameters "should not exceed the number of alternate glyphs in lookups associated with the feature". So 32 named options means a font shipping 32 named alternate glyphs of a single character. That isn't a font anyone has made, and a test for it would have to hand-build the table, which is why I skipped the test rather than writing one that can only pass against a fixture. Your call if you'd rather have it anyway.

Two things I did instead. `GetFeatureValues` now traces a warning at `:1042` when it truncates, naming the tag and both counts, so a font that ever hits this says so before a user notices. And the comment on the constant pointed at `FmtFntDlg.h`, which isn't in the tree, and nothing in any `.h` or `.cpp` references `kMaxValPerFeat` either. The ceiling is a vestigial managed constant, not a platform limit, so the comment now says what the number is and where it came from rather than sending the next reader to a file that doesn't exist. Your "reuses Graphite's" reading was exactly right, and it's the same pattern as the default-on discussion: the OpenType path inherited a Graphite constant that had no OpenType justification.

Same class, worse failure mode, also unreachable: the reader's own `MaxNamedParameters = 1024` at `OpenTypeFontFeatureInfoReader.cs:59`. Above that it returns no options at all, so a variant selector silently becomes a binary toggle. Given the observed maximum is four I've left it, but it's the more damaging of the two if anything ever gets there.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
jasonleenaylor: Taking your second option, with the measurement written down rather than just a note.

Your description is right, and understated: `ReadFeatureList` at `:129` walks the flat feature list and never touches `ScriptList`, so the reader has no concept of script at all. It isn't preferring the wrong script, it doesn't know scripts exist. A font that labelled the same tag differently per script would show one script's strings to the other, exactly as you say.

What decided me against building it: across 192 fonts, 176 cv/ss tags appear more than once in a flat feature list, so `IsRicher` at `:106` runs constantly, and zero of those duplicates differ in derived label or options. Every one is identical, so it's always choosing between equal records and the choice never matters.

Your scenario is also the normal case here rather than a corner, which is what made me check it properly. CharisSIL and DoulosSIL both cover DFLT, cyrl and latn; GentiumPlus adds grek; ScheherazadeNew covers arab and latn. So a Cyrillic writing system using a font that also covers Latin is the ordinary situation, and it's inert because these fonts define each feature once and register it under every script they cover.

The real fix means passing the writing system's script into `Read` and keeping per-script records instead of deduplicating by tag, which changes the signature and its only caller. That's a fair amount of work for a difference no font in the corpus exhibits, so the remarks on `Read` now say what the limitation is, why it's accepted, and what would break it, and the docs carry the same. Whoever meets the first font that labels a tag per script will find the reason rather than the behaviour.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
jasonleenaylor: Done, and there's a second hole behind it that the snapshot alone doesn't close.

Snapshotting the reader inside the lock stops a caller invoking the wrong one, which is what you described. It doesn't stop this: a discovery misses the cache and leaves the lock, the reader is swapped and the cache cleared, then the discovery finishes and seeds the freshly cleared cache with old-reader data. Every later caller then gets results from a reader that is no longer installed. The snapshot doesn't help because that discovery legitimately captured the old reader; the fault is that its result outlives the swap.

So the reader and a generation counter are read together at `:1166`, the counter is bumped on both the swap and the restore in `UseTableReaderForTests`, and at `:1186` a result whose generation has moved is handed back to the caller that asked for it but not written to the cache.

Being straight about reachability: this isn't reachable today. Nothing in `Src` carries `[Parallelizable]` or `LevelOfParallelism`, so despite `Test.runsettings` setting `NumberOfTestWorkers=0`, NUnit runs fixtures sequentially within an assembly, and the `MaxCpuCount` parallelism is across assemblies, which are separate processes with separate statics. In production the field is never reassigned at all, it's set once to `ReadTable`.

I fixed it anyway where I deferred your option-31 and cross-script items, and the difference is cost rather than severity. This is four lines in one private nested class with no signature change. Those two needed a font nobody ships and a signature change respectively. It also matters a bit more now: the `ReadOnlyCollection` change makes this method the only place the cache gets populated.

jasonleenaylor:
Round two. All nine inline comments answered, seven with code changes and two documented as accepted limitations. Details are in the threads; this is the method and the parts that don't belong in any one of them.

Fixed: cvNN option positions, the default-on set plus a test pinning it, `size` hidden, 57 feature names moved to resx, parse failures traced, the feature cache no longer handing out its own array, and the table reader read under its lock with a generation guard. Deferred with reasons in the threads: the >31-option limit and script-aware deduplication.

**Font survey.** Several of these came down to what fonts actually do, so I measured rather than reasoned. 192 font files: Windows fonts, the per-user font store, `DistFiles` and `fwroot/Downloads`. I applied the reader's own decode rules, platforms 0, 1 and 3 with a non-empty string, so the numbers describe what our code sees and not what fontTools can read.

    cvNN features declaring named options   477, across 15 fonts
    option counts                          1 option x420, 2 x40, 3 x16, 4 x1
    maximum named options                  4  (ScheherazadeNew cv82)
    above 31, where the menu truncates     0
    above 1024, where the reader drops all 0
    slots the reader cannot resolve        0
    fonts declaring 'size'                 0
    cv/ss tags appearing more than once    176
      of those with differing metadata     0

    script coverage   CharisSIL, DoulosSIL   DFLT cyrl latn
                      GentiumPlus            DFLT cyrl grek latn
                      ScheherazadeNew        arab latn

**Verification.** `build.ps1` clean, 0 warnings and 0 errors. Full `test.ps1`: 4156 passed, 75 not executed, 1 failed. The failure is `DataTreeOpt_VisibilitySequenceHasNoGaps`, which exceeded its 70000ms timeout and then passed in 12 seconds on its own. `DetailControls` and its tests contain no reference to `FontFeaturesButton`, `OpenTypeFeatureCatalog` or `OpenTypeFontFeatureInfoReader`, so nothing here can reach it; the machine was carrying 15 worktrees at the time. The 75 not-executed are pre-existing skips with their own messages, including the shaping test discussed below.

**LT-22774** covers what's deferred: confirming the default-on set against measured renderer behavior, and a test that every visible feature resolves a resx label. Fix version FW 9.3, related to LT-22638. The blocker there is that `ShapeText_OpenTypeFeatureToggleChangesShapingData` skips itself for want of a font with a deterministic liga delta, so the harness needs work before it can answer anything.

`Docs/opentype-font-features.md` now carries the design rationale for the default-on set, including why this provider isn't equivalent to the Graphite one on that point, so the next person to widen the set reads the reason first.

**On your three follow-ups:** defaults are aligned with the docs and pinned, with the measurement in LT-22774. Names are in resx, and I don't know whether the bundled fonts carry reusable translations, so treat that as still open. The UI follow-up for a selection summary and grouped live preview isn't filed. Say the word and I'll file it, or take it yourself if you have the shape in mind.

Nothing is resolved from my side. Items 7 and 8 are deferrals and 2 is partly one, so they're yours to close or push back on.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
johnml1135: Verified in a worktree build: slots keep their position, a middle slot decodes to null and is numbered by the UI, and the all-fail case stays binary. Accepted, including the binary fallback rule you added. Resolving.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
johnml1135: Accepted. The set is the documented four and `DefaultOnFeatures_AreOnlyTheDocumentedFour` pins it; the renderer measurement belongs in LT-22774. The reframing is right: this flag is a policy assertion, not font data, and the docs now say so. One follow-through in the summary comment about trimming the survey numbers. Resolving.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
johnml1135: Accepted, and thanks for the clause correction. Resolving.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
johnml1135: Accepted. Two warnings and one info through `FontFeatureSettings.DiagnosticsSwitch`, degradation unchanged, both robustness tests still pass locally. Resolving.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
johnml1135: Accepted, including returning the cache winner on the miss path. Wrapping once at insert is the right trade. Resolving.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
johnml1135: Accepted as-is. Nothing ships more than four options and the trace names the tag if a font ever does. The 1024 ceiling in the reader is accepted on the same basis; no ticket needed. Resolving.

Src/FwCoreDlgs/FwCoreDlgControls/FontFeaturesButton.cs
johnml1135: Accepted. Read the generation guard at 1158-1195 in the worktree; with `ReadOnlyCollection` making this the single cache-population site, the four lines earn their place. Resolving.

Src/Common/FwUtils/OpenTypeFeatureCatalog.cs
johnml1135: Withdrawn, and the repo settles it. The convention is that a string lives in the `.resx` of the project that displays it, not the project that owns the data table. `fieldworks-localization-review/SKILL.md` states it as the "owning project" rule, every assembly here has its own strings resx (`FwUtilsStrings.resx` holds only what FwUtils itself shows), and no other assembly reads `FwCoreDlgControls.resx` strings. Your `design.md` decision 3 already recorded this trade-off, and the keys have lived in this file since LT-22324. So `FwCoreDlgControls.resx` is right; `FwUtilsStrings.resx` would have broken the convention and re-keyed the existing translations.

Two things to carry forward, neither for this PR: the 54 hidden names staying in C# is fine because they are never shown; and when an Avalonia font dialog appears it will want these 71 names too, at which point a shared font-feature strings home becomes a real question. Noted on LT-22792. Resolving.

Src/Common/FwUtils/OpenTypeFontFeatureInfoReader.cs
johnml1135: Accepted, and the spec makes it stronger than an accepted limitation. The registry lists **Script/language sensitivity: None** for both `cv01`–`cv99` and `ss01`–`ss20` ([features_ae](https://learn.microsoft.com/en-us/typography/opentype/spec/features_ae), [features_pt](https://learn.microsoft.com/en-us/typography/opentype/spec/features_pt)), so a flat, one-label-set-per-tag reader is what the spec describes for these two families, not a shortcut. The tooling agrees: feaLib keys `featureNames` and `cvParameters` by tag alone, so fontmake, ufo2ft and Glyphs pipelines cannot author per-script labels for a shared tag, and HarfBuzz's `hb_ot_layout_table_get_feature_tags` documents the duplicates as raw script/langsys repetition for the caller to collapse. I could not find a single real font or issue thread that labels a cvNN differently per script. Word, InDesign and Paratext's reader were not checked, so this is strong evidence rather than proof.

One request for the cleanup pass: swap the justification. The `Read` remark and the docs paragraph currently lean on survey counts (178/173 there, 192/176 in this thread). Cite the registry's sensitivity field instead and drop the numbers; the spec does not go stale. Resolving.

johnml1135:
Round three. All nine threads are answered and resolved: seven fixes verified in a worktree build (`build.ps1 -SkipNative` clean; FwUtilsTests 402, FwCoreDlgControlsTests 55, FwCoreDlgsTests 353 pass, 5 pre-existing skips), the two deferrals accepted, the 1024 reader ceiling accepted with them, and the resx location withdrawn: the repo's rule is that strings live in the resx of the project that displays them, so `FwCoreDlgControls.resx` is correct.

The follow-up is filed: **LT-22792** "Font Features: show the set features together and preview them live", linked Related to LT-22638 and LT-22774. It also carries the note that an Avalonia font dialog will want the same 71 names.

Four things before merge, all yours, one cleanup commit:

1. **Rebase onto `main`.** The branch predates the comment-hygiene tooling, so `build.ps1 -CommentHygiene` does not exist here yet. After the rebase it does, and CI already checks the merge result: 57 advisory violations. 46 are mechanical rewraps the script does itself (read the result, it splits without reflowing). 11 need a hand: five comments over the 200-character budget (`OpenTypeFontFeatureInfoReader.cs:197`, `FontFeaturesButton.cs:32`, `:1170`, `:1183`, the 877-character one at `OpenTypeFeatureCatalogTests.cs:61`), two absence narrations (`FontFeaturesButton.cs:34`, `OpenTypeFeatureCatalog.cs:43`), two doc pointers (`OpenTypeFeatureCatalogTests.cs:63`, `:71`).
2. **Drop the survey numbers from everything that ships.** They already disagree (192/176 in the threads, 178/173 in `Docs/opentype-font-features.md` and the `Read` remark). Keep in the doc: the label order, the multi-value rule, the hidden and default-on lists, and one short paragraph on why default-on is a policy assertion. Move the registry argument and the counts to LT-22774 and the archived `research.md`, which is dated and allowed to age. For the script-blind paragraph, cite the registry's "Script/language sensitivity: None" for cvNN/ssNN instead of the count. This also clears most of the 11 above.
3. **Sync the delta spec.** `openspec/changes/archive/2026-07-24-improve-opentype-feature-ui/specs/font-feature-settings/spec.md` adds four requirements that `openspec/specs/font-feature-settings/spec.md` does not have. The previous OpenType change was synced; this one was archived without the sync.
4. **Move LT-22638 to In Progress** so the ticket matches the PR. Resolve with Fix Version FW 9.3 after the merge.

Next: Jason rebases and pushes the cleanup commit; I re-review the diff of that commit only.

jasonleenaylor:
Rebased and pushed as `4ae348262`, one cleanup commit as asked. All four items are done; three notes where what I found differs from what the list expected.

**1. Rebase and hygiene.** The 46 mechanical rewraps are in, and the 11 are fixed by hand. Two corrections to the breakdown: the count of 11 is right, but the enumeration named nine — `OpenTypeFontFeatureInfoReaderRobustnessTests.cs:118` (231 chars) and `:137` (249) are also over budget and are not on the list. And one of the two absence narrations is a false positive: the rule matches `\bused to\b`, and `OpenTypeFeatureCatalog`'s summary said "features **used to** classify feature tags", meaning *employed to*. I reworded it to "that classifies" rather than argue with the checker.

**2. Survey numbers.** Gone from everything that ships. The doc paragraph and the `Read` remark now cite the registry's Script/language sensitivity field, and the survey is recorded in the archived `research.md` — scope, date, and the fact that it is machine-local and not reproducible from the repository, which is the real reason it does not belong in a comment.

There was a third pair, which is worse than two disagreeing: `OpenTypeFontFeatureInfoReaderRobustnessTests.cs` carried "248 of the 275", SIL fonts only, agreeing with neither 178/173 nor 192/176. All three are now in the archive together with the option-count distribution behind the ceiling and binary-fallback decisions.

**This cleared none of the 11.** The survey numbers were in the doc and the `Read` remark; the flagged comments were different ones. The lengths came down on their own merits instead.

Worth flagging one cut: the 872-character comment on `DefaultOnFeatures_AreOnlyTheDocumentedFour` is now three lines. Everything else in it — the registry divergence, the Graphite-versus-OpenType asymmetry, why a wrong assertion inverts the first click — is already in `Docs/opentype-font-features.md`, and stated more accurately there. What survives is the only test-local part: why the test asserts the whole set rather than membership one tag at a time.

**3. Delta spec.** Synced. The MODIFIED discovery requirement plus all four ADDED ones, inserted after the discovery requirement rather than appended, since the live file reads thematically.

**4.** Ticket moved.

**Verification.** Comment hygiene clean and stable across two local runs, and green in CI. `check-whitespace`, commit messages, `stray-docs`, `lychee` and `validate-refs` all pass. Every changed C# line is a comment, and every doc comment in the six changed files parses as well-formed XML.

**I did not compile this locally.** A targeted build of `FwUtils` and `FwCoreDlgControls` is not possible in a fresh worktree — `ViewsInterfaces` needs native artifacts and `CommonAssemblyInfo.cs` is generated — so the native build is the whole cost, and I left it to CI. "Build Debug and run tests" is still running as I write this; if it fails, that is mine.

Unrelated byproduct: `Worktree-CreateFromBranch.ps1` aborts under PowerShell 5.1 when git's checkout progress reaches stderr, leaving a worktree created but never colorized. That is #1131.

johnml1135 (approved):

PR #1038, opened 2026-08-04
LT-21402: Part 2 - Add DictionaryConfigured event

mark-sil:
## Summary

The Configure Dictionary dialog senders did not really mean "MasterRefresh"; they meant
"the dictionary configuration (or its styles) changed — regenerate any view that renders
generated dictionary content". Under the Mediator that intent was routed by dispatch
tricks: in the XHTML document views (Dictionary, Reversal Indexes, Classified Dictionary)
`XhtmlDocView`'s High-priority stop-when-handled intercept turned the message into a cheap
content-only regeneration, while in Lexicon Edit/Browse the window ran a full master
refresh whose only needed effect was that the dictionary preview pane regenerated as a
side effect.

Pub/Sub has no priority or stop-when-handled, so this PR names the intent,
**`DictionaryConfigured`**, as its own event instead.

## Changes

- The two senders — `DictionaryConfigurationListener.OnConfigureDictionary`
  (Tools → Configure) and `XhtmlDocView.RunConfigureDialogAt` (right-click Configure) —
  publish `DictionaryConfigured` (window-scoped) when configuration changes were saved.
- `XhtmlDocView`: the `OnMasterRefresh` handler body is renamed to `RefreshAllContent`
  and subscribed to `DictionaryConfigured` (Init/Dispose). A thin public
  `OnMasterRefresh()` wrapper remains so the Mediator still reaches the view for F5 and
  for the broadcast calls not yet converted to Pub/Sub.
- `XhtmlRecordDocView` (the Lexicon Edit dictionary preview pane) subscribes a new
  `RefreshPreviewContent` handler, so the preview still updates now that Configure
  Dictionary no longer triggers a full refresh. Closing the dialog from Lexicon Edit is
  therefore much cheaper: the preview regenerates directly instead of riding on a
  full-application refresh.
- Changes made through the dialog's **sub-dialogs** to settings rendered *outside* the
  generated dictionary content — styles, and the global homograph configuration
  (Headword Numbers) — still require the full master refresh (e.g. the Browse view's
  Headword column shows homograph numbers). To track this, the controller's
  `MasterRefreshRequired` bool is replaced by a `RefreshRequired` enum
  (`None` / `DictionaryContent` / `Master`) that only escalates; the senders publish
  `MasterRefresh` or `DictionaryConfigured` accordingly. The full master refresh is
  currently the only mechanism that reaches the affected views; a narrower invalidation
  could replace the `Master` level for these causes if one is ever added.

## Bugs fixed

[...]

PR #1039, opened 2026-08-04
LT-21402: Part 3 - Move MasterRefresh to Pub/Sub (Insert/Delete)

mark-sil:
## Summary

Converts the last two synchronous MasterRefresh senders —
`InsertEntryDlgListener.DialogInsertItemInVector` (Insert Entry) and
`RecordClerk.OnDeleteRecord` (Delete Record) — completing the synchronous half of the
LT-21402 migration (follows #1033 and #1038).

These two were deliberately excluded from Part 1: both commands update the cache directly
(UOW + PropChanged), so when an XHTML document view is active, the Mediator's
High-priority stop-when-handled intercept correctly satisfied the refresh with a fast
content-only reload; the full master refresh was only needed in the other tools.
Converting them to a plain `Publish(MasterRefresh)` would have turned that fast, correct
reload into a needless full refresh.

## Design: two-stage refresh request

Pub/Sub has no priority or stop-when-handled, so the intercept is replaced with a
two-stage request, using the same `ReturnObject` idiom as the `DialogInsertItemInVector`
conversions (see `RecordClerk.OnInsertItemInVector`):

- New event **`RefreshCurrentView`**. The sender publishes it (window-scoped) with a
  `ReturnObject`; a subscriber that can satisfy the request by regenerating its own
  content sets `ReturnValue`. If no subscriber claims the request, the sender publishes
  `MasterRefresh` instead.
- `XhtmlDocView` subscribes while it exists (Init/Dispose) — i.e. exactly while its tool
  is the active content control — so the subscription lifetime does the routing that
  colleague priority used to do.
- `XhtmlRecordDocView` (the Lexicon Edit preview pane) deliberately does **not**
  subscribe: in Lexicon Edit these commands do the full refresh today, and that is
  preserved.

Behavior is unchanged: content-only reload when a Dictionary, Reversal Indexes, or
Classified Dictionary document view is active; full master refresh everywhere else.

The `JumpToRecord` send next to the Insert Entry site stays on the Mediator; that message
is a separate conversion.

## Remaining MasterRefresh work (future PRs)

Still on the Mediator: the F5/CmdRefresh menu command (which can reuse this two-stage
pattern when converted), four deferred `BroadcastMessage` senders, and one `PostMessage`
sender — each needs per-site timing analysis before conversion.

## Testing

- Build clean; no automated-test changes (dispatch-only change; behavior covered by
  manual tests).
- Manual: Insert Entry and Delete Record from the Dictionary view (content-only reload,
  `FwXWindow.OnMasterRefresh` not invoked, jump-to-new-entry intact) and from Lexicon
  Edit (full refresh, unchanged); reversal document view delete; multi-window scoping
  (single dispatch, focus retained).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1039)

### DIFF
--- a/Src/xWorks/XhtmlDocView.cs
+++ b/Src/xWorks/XhtmlDocView.cs
@@ -82,6 +82,7 @@ namespace SIL.FieldWorks.XWorks
 				}
 			}
 			Subscriber.Subscribe(EventConstants.DictionaryConfigured, RefreshAllContent, m_propertyTable.GetWindow());
+			Subscriber.Subscribe(EventConstants.RefreshCurrentView, RefreshCurrentView, m_propertyTable.GetWindow());
 		}
 
 		protected override void Dispose(bool disposing)
@@ -93,6 +94,7 @@ namespace SIL.FieldWorks.XWorks
 			if (disposing)
 			{
 				Subscriber.Unsubscribe(EventConstants.DictionaryConfigured, RefreshAllContent);
+				Subscriber.Unsubscribe(EventConstants.RefreshCurrentView, RefreshCurrentView);
 			}
 			base.Dispose(disposing);
 		}
@@ -1321,6 +1323,27 @@ namespace SIL.FieldWorks.XWorks
 		public void OnMasterRefresh(object sender)
 		{
 			RefreshAllContent(null);
+		}
+
+		/// <summary>
+		/// First stage of a sender's two-stage refresh request: while this view is the active
+		/// content control it claims the request and satisfies it by regenerating its own
+		/// content, so the sender skips the full master refresh.
+		/// </summary>
+		private void RefreshCurrentView(object obj)
+		{
+			if (!(obj is ReturnObject retObj))
+			{
+				Debug.Assert(false, "Received unexpected object type.");
+				return;
+			}
+			// Return if already handled by another Subscriber.
+			if (retObj.ReturnValue)
+			{
+				return;
+			}
+			RefreshAllContent(null);
+			retObj.ReturnValue = true;
 		}
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


Src/xWorks/XhtmlDocView.cs
```diff
@@ -1323,6 +1325,27 @@ public void OnMasterRefresh(object sender)
...
+		/// content, so the sender skips the full master refresh.
+		/// </summary>
+		private void RefreshCurrentView(object obj)
+		{
+			if (!(obj is ReturnObject retObj))
+			{
+				Debug.Assert(false, "Received unexpected object type.");
+				return;
+			}
+			// Return if already handled by another Subscriber.
```
papeh, line 1340: Would there ever be two current views that need to handle this (e.g. multiple panes or windows)?

mark-sil: Good question — no, by design, on both counts:

Multiple windows: both ends are window-scoped (m_propertyTable.GetWindow()), so a publish from window A's clerk only reaches window A's subscribers. A second window showing the Dictionary view never sees the request (matching the old per-window Mediator behavior).

Multiple panes in one window: XhtmlDocView only exists as the main content control of the three document tools (Dictionary, Reversal Indexes, Classified Dictionary), and a window has exactly one main content control at a time (this is similar to what the Mediator relied on) — the view is disposed (and unsubscribed) on tool switch. The one view that can coexist alongside another content control, XhtmlRecordDocView (the Lexicon Edit preview pane), deliberately doesn't subscribe: in Lexicon Edit these commands should do the full refresh, and the preview updates via PropChanged. So at most one subscriber can receive any given request today.

papeh (approved):

PR #1048, opened 2026-08-07
Fix the failing LexOptionsDlg InitializeUIModeControls shift test

jasonleenaylor:
Measure the shift delta from m_labelAdvanced's observed move instead of recomputing it with a hardcoded padding literal, this will keep the test from drifting from the actual code. Add an assertion that the injected group box clears m_labelAdvanced to keep proof the shift was sufficient, and drop the label position check.

Also rename the Advanced heading from label4 to m_labelAdvanced so the layout code names the control it moves.

## Why the test was failing

PR #964 changed InitializeUIModeControls to derive its padding from the dialog font (`Font.Height / 2`) instead of a literal 8, but the test kept recomputing its expected positions with the old literal. Wherever `Font.Height / 2 != 8` the expectations are off by the difference (at 96 DPI, `13 / 2 = 6`, hence "Expected: 217 / But was: 215"), so the test could pass on one machine and fail on another.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1048)

### DIFF
--- a/Src/LexText/LexTextControls/LexOptionsDlg.Designer.cs
+++ b/Src/LexText/LexTextControls/LexOptionsDlg.Designer.cs
@@ -52,7 +52,7 @@ namespace SIL.FieldWorks.LexText.Controls
 			this.m_chDescription = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
 			this.m_tabInterface = new System.Windows.Forms.TabPage();
 			this.m_autoOpenCheckBox = new System.Windows.Forms.CheckBox();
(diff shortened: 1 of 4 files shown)


thejambi (approved):
:lgtm:

> @thejambi reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1048#-:-OzRPQGU4_OIHUNradN3:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jasonleenaylor).

PR #1051, opened 2026-08-07
LT-22576: Don't allow set phon feats on user-defined NCs

aror92:
Add method to RuleFormulaControl to check if NC is user-defined.

Update OnDisplayContextSetFeatures in RuleFormulaSlice to display the set phonological features option only if the NC uses features and is not user-defined.

_Note: currently there is no way to distinguish user-defined NCs from auto-generated NCs other than the class's name. When we make model changes, we plan to add a flag to track whether a NC is auto-generated, and then the check here can be modified._

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1051)


aror92:
I've had Devin review and have addressed all comments except the following:

> _Sibling slice still enables a features-related command without the new check_
> RegRuleFormulaSlice.OnDisplayContextSetVariables (RegRuleFormulaSlice.cs:71) still gates purely on IsFeatsNCContextCurrent. That command sets feature variables rather than feature values, so it may intentionally remain enabled for user-defined classes, but it edits the same shared PhNCFeatures object; worth confirming whether LT-22576 also applies there.

The request in LT-22576 was specifically about the "Set Phonological Features" option, not about set variables. I don't know whether users would want set variables gated similarly as well. I wouldn't make this additional change unless users request it, or unless a human reviewer thinks it is needed.

aror92:
Unit tests to be added in a separate PR. Danny is kick-starting that for me using Claude (since I don't have a subscription yet), to get all the test set-up out of the way.

thejambi (approved):
:lgtm:

> @thejambi reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1051#-:-OzSaL6YBH-_3Em4d2MK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on aror92).

thejambi (approved):
> @thejambi reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on aror92).

PR #1053, opened 2026-08-10
LT-21720: Give unpositioned dialogs centered start positions and owners

jasonleenaylor:
## Quick Summary

Partially addresses [LT-21720](https://jira.sil.org/browse/LT-21720) (remembered window locations misbehaving on multi-monitor setups).

- Revive `UploadToWebonaryDlg` position persistence: its restore/clamp/save code null-checked a `PropertyTable` auto-property that the constructor never assigned, so it has been dead since it was written. The property is now a constructor-assigned field, and the dialog is shown with the main window as owner. Per review feedback, `OnShown` now recenters the dialog on the owner's screen when the remembered position is on a different monitor, so a modal dialog is never stranded where the user isn't looking.
- Give `DictionaryConfigurationManagerDlg` a `CenterParent` start position and clamp its bounds back onto a screen in `OnShown` — it previously had no positioning code and could open off-screen, leaving the app looking hung (one of the LT-21720 reports).
- Sweep the highest-risk dialogs that had no positioning code at all: `CenterParent` on ten (dictionary-configuration child dialogs, Webonary log viewer, the Gecko-hosting MGA/diagnostics/import-marker dialogs, semantic-domains chooser, respeller, concordance dialog) and `CenterScreen` on five that can be shown before any main window exists (the four FdoUi restore dialogs and the update chooser).
- Pass an explicit owner at the six `ShowDialog()` call sites that had none, threading an owner parameter through `DictionaryConfigurationImportController.DisplayView`.

This PR does not cover the dialogs that persist and restore positions of their own (`BaseGoDlg` and its subclasses, `ReallySimpleListChooser`, `InsertEntryDlg`'s registry restore, `FwFindReplaceDlg`, the shared `Persistence` helper, and similar) — several of the dialogs named in LT-21720 are in that family, which is follow-up work.

## CI-ready checklist

  ```powershell
  git fetch origin
  git diff --check --cached
  ```
- [ ] If this is core-developer AI-assisted work, I followed `Docs/workflows/ai-pr-workflow.md` and ran `pr-preflight` or the equivalent branch-readiness review before requesting review.

WinForms' default `WindowsDefaultLocation` leaves placement to the OS cascade, which on multi-monitor setups can land a dialog on the wrong screen or entirely off the visible area. Centering on an owner — or on a screen, for dialogs that can appear before any window exists — takes the OS out of the placement decision. The swept dialogs are behavior-neutral one-liners except `DisplayView`, whose new owner parameter has exactly one caller (updated in the same commit).

Validation: `build.ps1 -SkipNative` clean; `test.ps1` on `xWorksTests` filtered to UploadToWebonary + DictionaryConfiguration, all passing after rebasing onto current main.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1053)

### DIFF
--- a/Src/Common/Controls/DetailControls/SemanticDomainReferenceLauncher.cs
+++ b/Src/Common/Controls/DetailControls/SemanticDomainReferenceLauncher.cs
@@ -63,7 +63,7 @@ namespace SIL.FieldWorks.Common.Framework.DetailControls
 			var labels = ObjectLabel.CreateObjectLabels(m_cache, m_obj.ReferenceTargetCandidates(m_flid),
 				m_displayNameProperty, displayWs);
 			chooser.Initialize(labels, sense.SemanticDomainsRC, m_propertyTable);
-			var result = chooser.ShowDialog();
+			var result = chooser.ShowDialog(FindForm());
 			if(result == DialogResult.OK)
 			{
 				UndoableUnitOfWorkHelper.DoUsingNewOrCurrentUOW(Resources.DetailControlsStrings.ksUndoSet,
--- a/Src/Common/Controls/DetailControls/SemanticDomainsChooser.cs
+++ b/Src/Common/Controls/DetailControls/SemanticDomainsChooser.cs
@@ -47,6 +47,7 @@ namespace SIL.FieldWorks.Common.Framework.DetailControls
 		public SemanticDomainsChooser()
 		{
 			InitializeComponent();
+			StartPosition = FormStartPosition.CenterParent;
 			btnCancelSearch.Init();
 			editDomainsLinkPic.Image = DetailControlsStrings.gotoLinkPic;
 		}
--- a/Src/Common/FwUtils/FwUpdateChooserDlg.cs
+++ b/Src/Common/FwUtils/FwUpdateChooserDlg.cs
@@ -14,6 +14,7 @@ namespace SIL.FieldWorks.Common.FwUtils
 		public FwUpdateChooserDlg()
 		{
 			InitializeComponent();
+			StartPosition = FormStartPosition.CenterScreen;
 		}
 
 		public FwUpdateChooserDlg(FwUpdate current, IEnumerable<FwUpdate> available) : this()
--- a/Src/FdoUi/Dialogs/CantRestoreLinkedFilesToOriginalLocation.cs
+++ b/Src/FdoUi/Dialogs/CantRestoreLinkedFilesToOriginalLocation.cs
@@ -23,6 +23,8 @@ namespace SIL.FieldWorks.FdoUi.Dialogs
 		{
 			m_helpTopicProvider = helpTopicProvider;
 			InitializeComponent();
+			// Shown without an owner, possibly before any main window exists.
+			StartPosition = FormStartPosition.CenterScreen;
 		}
 
 		/// <summary>
--- a/Src/FdoUi/Dialogs/ConflictingSaveDlg.cs
+++ b/Src/FdoUi/Dialogs/ConflictingSaveDlg.cs
@@ -19,6 +19,8 @@ namespace SIL.FieldWorks.FdoUi.Dialogs
 		public ConflictingSaveDlg()
 		{
 			InitializeComponent();
+			// Shown without an owner, possibly before any main window exists.
+			StartPosition = FormStartPosition.CenterScreen;
 			pictureBox1.BackgroundImage = SystemIcons.Warning.ToBitmap();
 			pictureBox1.Size = SystemIcons.Warning.Size;
 		}
--- a/Src/FdoUi/Dialogs/FilesToRestoreAreOlder.cs
+++ b/Src/FdoUi/Dialogs/FilesToRestoreAreOlder.cs
@@ -23,6 +23,8 @@ namespace SIL.FieldWorks.FdoUi.Dialogs
 		{
 			m_helpTopicProvider = helpTopicProvider;
 			InitializeComponent();
+			// Shown without an owner, possibly before any main window exists.
+			StartPosition = FormStartPosition.CenterScreen;
 		}
 
 		/// <summary>
--- a/Src/FdoUi/Dialogs/RestoreLinkedFilesToProjectsFolder.cs
+++ b/Src/FdoUi/Dialogs/RestoreLinkedFilesToProjectsFolder.cs
@@ -23,6 +23,8 @@ namespace SIL.FieldWorks.FdoUi.Dialogs
 		{
 			m_helpTopicProvider = helpTopicProvider;
 			InitializeComponent();
+			// Shown without an owner, possibly before any main window exists.
+			StartPosition = FormStartPosition.CenterScreen;
 		}
 
 		/// <summary>
--- a/Src/LexText/LexTextControls/InsertEntryDlg.cs
+++ b/Src/LexText/LexTextControls/InsertEntryDlg.cs
@@ -1837,7 +1837,7 @@ namespace SIL.FieldWorks.LexText.Controls
 
 				using (dlg)
 				{
-					if (dlg.ShowDialog() == DialogResult.OK)
+					if (dlg.ShowDialog(this) == DialogResult.OK)
 					{
 						Gloss = dlg.Result;
 						m_MGAGlossListBoxItems = dlg.Items;
--- a/Src/LexText/LexTextControls/LexImportWizardMarker.cs
+++ b/Src/LexText/LexTextControls/LexImportWizardMarker.cs
@@ -396,6 +396,7 @@ namespace SIL.FieldWorks.LexText.Controls
 			// Required for Windows Form Designer support
 			//
 			InitializeComponent();
+			StartPosition = FormStartPosition.CenterParent;
 			AccessibleName = GetType().Name;
 
 			InitBottomPanel();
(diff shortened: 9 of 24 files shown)


mark-sil (approved):
> @mark-sil reviewed 24 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1053)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jasonleenaylor).

thejambi (approved):
> @thejambi reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1053)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jasonleenaylor).

thejambi:
*[`Src/xWorks/UploadToWebonaryDlg.cs` line 352 at r1](https://reviewable.io/reviews/sillsdev/FieldWorks/1053#-Ozkjf2I8hCQp3j8vuks:-Ozkjf2I8hCQp3j8vukt:b-baecy5) ([raw file](https://github.com/sillsdev/FieldWorks/blob/36daa0d7eb3df606261b990c4a68d340b6396441/Src/xWorks/UploadToWebonaryDlg.cs#L352)):*
> ```Smalltalk
> 			}
> 			base.OnClosing(e);
> 		}
> ```

Based on my review of the changes, against some of the notes I had gathered about it -- this one especially should be verified, it may not handle multi-monitor setup well.

thejambi:
Some Claude feedback based on my notes when looking into the issue so far. I wonder if yours would agree with this assessment? 

---

Head: `36daa0d` (`claude/webonary-dictconfig-dialog-position`). Line numbers below are at PR head.

---

## 1. Top-level review comment (post as the review summary)

> The `CenterParent` / `CenterScreen` sweep and the owner-threading look right to me, and the
> mechanism is the correct one: `Form.CenterToParent()` reads `GWL_HWNDPARENT`, centers on the
> owner's rectangle and clamps to *that* screen's working area, so these dialogs will now land on
> the monitor the main window is on. `CenterScreen` for the ownerless FdoUi/update dialogs is the
> right call too, since it centers on the screen under the mouse rather than always the primary.
> `GetValue<IWin32Window>("window")` matches existing usage in `DictionaryConfigurationListener`
> and `XhtmlDocView`, and `DisplayView(dialog, owner)` has only the one caller in the repo.
>
> Two things I'd like to sort out before this merges:
>
> **1. The revived Webonary persistence re-introduces the wrong-monitor bug** (see inline comment on
> `UploadToWebonaryDlg.cs`). That restore block was previously dead code, because `PropertyTable`
> was never assigned. Now that it runs, it sets `StartPosition = Manual` from a remembered rect that
> is only clamped by `ScreenHelper.EnsureVisibleRect` — which guarantees the dialog is on *some*
> screen, not on the owner's screen — and that assignment overrides the owner we now pass from
> `FwXWindow`. So this is the one dialog in the PR that can still come back on the wrong monitor,
> and it's this PR that turns that path on.
>
> **2. Scope vs. title.** The title reads as a general fix, but the dialogs that actually *persist*
> and restore a position are untouched: `BaseGoDlg.SetDlgInfo` (and its ~10 subclasses — EntryGoDlg,
> LinkEntryOrSenseDlg, LinkAllomorphDlg, RecordGoDlg, WordformGoDlg, ReversalEntryGoDlg, …),
> `ReallySimpleListChooser` (which gets *no* clamp at all when constructed with a null
> `IPersistenceProvider`, e.g. from `EntrySequenceReferenceLauncher`), `InsertEntryDlg`'s own
> registry restore, `FwFindReplaceDlg`, and the shared `Persistence` helper. Those are the cases in
> the multi-monitor bug report I filed, and they're not covered here. Same pattern also lives in
> `MergeObjectDlg`, `AddNewSenseDlg`, `MasterCategoryListDlg`, `MasterListDlg`, `MsaCreatorDlg`,
> `PhonologicalFeatureChooserDlg`, `MsaInflectionFeatureListDlg`, `ExportDialog`,
> `XmlDocConfigureDlg` and `DictionaryConfigurationDlg`.
>
> I'm happy for that to stay out of this PR — but could we either narrow the title (e.g. "Give
> unpositioned dialogs a centered start position and owners") or note in the description that the
> position-persistence family is follow-up work? Otherwise the ticket looks closed when the reported
> repro still reproduces.
>
> Suggestion for the follow-up: a single shared helper (`ScreenHelper` or a new
> `DialogPositionHelper`) that all the restore sites can call, rather than per-dialog `OnShown`
> overrides — something like:
>
> ```csharp
> /// <summary>
> /// If a remembered position placed the dialog on a different monitor than its owner (e.g. the
> /// main window has since moved to another screen), recentre it on the owner's screen.
> /// </summary>
> public static void EnsureOnOwnerScreen(Form dlg)
> {
>     var owner = dlg.Owner ?? Form.ActiveForm;
>     if (owner == null || dlg.StartPosition != FormStartPosition.Manual)
>         return;
>     var ownerScreen = Screen.FromControl(owner);
>     if (ownerScreen.DeviceName == Screen.FromRectangle(dlg.DesktopBounds).DeviceName)
>         return;   // remembered position is already on the owner's monitor — leave it alone
>     var wa = ownerScreen.WorkingArea;
>     dlg.Location = new Point(
>         wa.Left + Math.Max(0, (wa.Width - dlg.Width) / 2),
>         wa.Top + Math.Max(0, (wa.Height - dlg.Height) / 2));
> }
> ```
>
> It has to run where `Owner` is populated (`OnShown`), not in `SetDlgInfo`/the constructor —
> `ShowDialog(owner)` assigns the owner after those have run.

---

## 2. Inline — `Src/xWorks/UploadToWebonaryDlg.cs`, lines 68–78 (the restore block)

> This is the one change in the PR that can still put a dialog on the wrong monitor, and it's newly
> live: with `PropertyTable` never assigned, this whole block was dead, so the dialog always got
> default owner-relative placement. Now it restores a remembered rect and sets
> `StartPosition = Manual`, which overrides the owner you just started passing in `FwXWindow`.
>
> `ScreenHelper.EnsureVisibleRect` only guarantees the rect intersects *some* monitor's working
> area. A rect entirely inside monitor A is already "visible", so it stays on monitor A even after
> the user has moved FLEx to monitor B — and because the dialog is modal, the main window is
> disabled and FLEx looks frozen on the monitor the user is actually looking at.
>
> Could we add an owner-screen correction in `OnShown` (where `Owner` is finally populated)? e.g.
>
> ```csharp
> protected override void OnShown(EventArgs e)
> {
>     base.OnShown(e);
>     // A remembered position may be on a monitor the main window has since left; recentre on the
>     // owner's screen so the dialog isn't stranded off-view behind a modal state.
>     var owner = Owner ?? Form.ActiveForm;
>     if (owner == null || StartPosition != FormStartPosition.Manual)
>         return;
>     var ownerScreen = Screen.FromControl(owner);
>     if (ownerScreen.DeviceName == Screen.FromRectangle(DesktopBounds).DeviceName)
>         return;
>     var wa = ownerScreen.WorkingArea;
>     Location = new Point(wa.Left + Math.Max(0, (wa.Width - Width) / 2),
>         wa.Top + Math.Max(0, (wa.Height - Height) / 2));
> }
> ```
>
> Alternatively, if we don't want the extra code here, adding `StartPosition = CenterParent` and
> dropping the position restore for this dialog would also be consistent with the rest of the PR.

---

## 3. Inline — `Src/xWorks/UploadToWebonaryDlg.cs`, lines 342–351 (`OnClosing`)

> Minor, but this save path is also newly live. If the user maximizes this (resizable) dialog,
> `Location`/`Size` are the maximized values — origin around `(-8, -8)` on Windows — and we restore
> them as a *normal*-state window, so it reopens slightly off the top-left corner. Worth using the
> restore bounds:
>
> ```suggestion
> 			if (m_propertyTable != null)
> 			{
> 				var bounds = WindowState == FormWindowState.Normal ? DesktopBounds : RestoreBounds;
> 				m_propertyTable.SetProperty("UploadToWebonaryDlg_Location", bounds.Location, false);
> 				m_propertyTable.SetPropertyPersistence("UploadToWebonaryDlg_Location", true);
> 				m_propertyTable.SetProperty("UploadToWebonaryDlg_Size", bounds.Size, false);
> 				m_propertyTable.SetPropertyPersistence("UploadToWebonaryDlg_Size", true);
> 			}
> ```

---

## 4. Inline — `Src/xWorks/DictionaryConfigurationManagerDlg.cs`, lines 54–60 (`OnShown`)

> Nit: after the `CenterParent` you added at line 28, this is close to a no-op — `CenterToParent`
> already clamps the dialog into the owner's screen working area, so `EnsureVisibleRect` won't have
> anything to move. And where it *would* fire, it only guarantees "on some screen", not "on the
> owner's screen", so the comment ("not displayed off-screen on multi-monitor setups") promises a bit
> more than it delivers.
>
> Fine to leave as belt-and-braces — but if we add the shared `EnsureOnOwnerScreen` helper mentioned
> in my summary, this override should call that instead so all the dialogs behave the same way.

---

## 5. Optional inline — `Src/xWorks/UploadToWebonaryDlg.cs`, line 38

> Just noting for the record: changing `protected PropertyTable PropertyTable { get; set; }` to a
> private field is a (harmless) API narrowing — I checked and nothing else in the repo subclasses
> this dialog or reads that property, so no callers break.

thejambi (approved):
New changes look good!

> @thejambi reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1053#-:-Ozlo7u8Etm0TWyoGn3X:b-ze20rf)* status: all files reviewed, 2 unresolved discussions (waiting on jasonleenaylor).

PR #1054, opened 2026-08-11
LT-21483: Part 1 - Move JumpToRecord to Pub/Sub (SendMessage)

mark-sil:
## Summary

Converts every synchronous `SendMessage("JumpToRecord")` sender to a window-scoped `Publisher.Publish` (sync-to-sync, timing-neutral) — the original six sites plus the two just added by the Avalonia Insert/Merge Entry dialogs. The Avalonia insert path's `SendMessage("MasterRefresh")` is likewise converted to the two-stage `RefreshCurrentView`/`MasterRefresh` publish (the LT-21402 pattern), still only for newly created entries.

Part 1 of the JumpToRecord conversion; the 14 deferred `Broadcast*` senders and all `OnJumpToRecord` Mediator entry points are unchanged (dual dispatch until Part 2).

## Design

- **`RecordClerk` is the single Pub/Sub subscriber:** a new `protected virtual JumpToRecordViaPublisher`, subscribed in `ActivateUI` and removed in `RemoveNotification`. Every implementation first bails out unless `IsActiveClerk`, because dependent clerks (e.g. `OccurrencesOfSelectedWordform`) subscribe without ever becoming the "ActiveClerk" — without the guard, one publish could run two clerks' jumps in unspecified order.
- **`OccurrencesOfSelectedUnit` overrides the virtual** to give `ConcordanceControl` the first chance at a published jump, so the clerk's jump-miss path (which would add texts to the interesting-texts list and can prompt about filters) cannot run in the Concordance tool. No converted sender can currently fire in that tool (LT-11998 hides the FLExText import there), so this arbitration is groundwork the Part 2 Broadcast-sender conversions will rely on.
- **`XhtmlDocView` deliberately does not subscribe:** under the Mediator its handler runs only when the clerk misses, and no converted sender can reach a clerk-miss in its views (Dictionary, Reversal Indexes, and Classified Dictionary are all Lexicon-area, while the one import that could miss is gated to Texts & Words). Part 2 designs its clerk-miss arbitration when converting Follow-Link.

## What stopped happening, and why each disappearance is safe

Nothing user-reachable — every converted sender fires in a context where the active clerk was already the only effective Mediator responder. The two corners where dispatch behavior differs on paper (FLExText import while a Dictionary/Reversal doc view is active) are unreachable: that import is only offered in the Texts & Words area.

## Validation

- `build.ps1 -SkipNative` clean; re-grep confirms zero `SendMessage("JumpToRecord"`/`SendMessage("MasterRefresh"` sites remain and the 14 `Broadcast*` sites are untouched.
- xWorksTests: 1221 passed / 2 skipped. ITextDllTests: 207 passed / 1 skipped.
- Manual GUI testing: insert/merge entry (Legacy and New UI), FLExText import across Texts & Words tools (jump to new text in Interlinear Texts; interesting-texts behavior in Analyses/Complex Concordance), Grammar category insert, Notebook record insert, and the two-window check (a jump in one window does not move the other).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1054)

### DIFF
--- a/Src/FdoUi/PartOfSpeechUi.cs
+++ b/Src/FdoUi/PartOfSpeechUi.cs
@@ -6,6 +6,8 @@ using System.Diagnostics;
 using System.Windows.Forms;
 using SIL.LCModel;
 using XCore;
+using SIL.FieldWorks.Common.FwUtils;
+using static SIL.FieldWorks.Common.FwUtils.FwUtils;
 using SIL.FieldWorks.LexText.Controls;
 
 namespace SIL.FieldWorks.FdoUi
@@ -52,9 +54,7 @@ namespace SIL.FieldWorks.FdoUi
 					case DialogResult.OK: // Fall through.
 					case DialogResult.Yes:
 						posUi = new PartOfSpeechUi(dlg.SelectedPOS);
-#pragma warning disable 618 // suppress obsolete warning
-						mediator.SendMessage("JumpToRecord", dlg.SelectedPOS.Hvo);
-#pragma warning restore 618
+						Publisher.Publish(new PublisherParameterObject(EventConstants.JumpToRecord, dlg.SelectedPOS.Hvo, propertyTable.GetWindow()));
 						break;
 				}
 			}
--- a/Src/LexText/Interlinear/ConcordanceControl.cs
+++ b/Src/LexText/Interlinear/ConcordanceControl.cs
@@ -1789,6 +1789,21 @@ namespace SIL.FieldWorks.IText
 		{
 			return "";
 		}
+
+		/// <summary>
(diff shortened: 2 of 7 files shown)


thejambi (approved):
:lgtm:

> @thejambi reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1054#-:-OzmHiF_-R3nR1PF7EtX:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mark-sil).

PR #1055, opened 2026-08-11
Clean up PCPATR Preparer a little

papeh:
* Capture newly-created ICmPossibilityLists directly when created to simplify logic and prevent race condition interceptions (none proven or even likely)
* Remove unused usings
* Make Cache and ShowMessages properties readonly

## Quick Summary
There is a problem with the cache not getting updated immediately when these lists are created. I don't expect these changes to fix this problem, but I can't imagine them doing any harm, and they remove a theoretical possibility of interference.

## CI-ready checklist

- [ ] Builds/tests pass locally (or I've run the CI-style build via Bash script or MSBuild).
- [ ] For any `Src/**` folders touched, corresponding `AGENTS.md` files are updated or explicitly confirmed still accurate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1055)


AndyBlack (approved):
While this does not fix LT-22649, these are good changes to have.  I approve them.

PR #1056, opened 2026-08-12
LT-22674: fix quadratic NFC offset-map build that made opening decomposed texts take 10+ seconds

johnml1135:
Fixes LT-22674.

Opening a text in Texts & Words took 10+ seconds when the baseline was a single long paragraph containing more than one writing system. Reproduced on Maasai-Parser with camus4-D (1 paragraph, 236 segments, 7,571 characters, two writing systems on two different font families).

## Cause

`BuildNfcOffsetMaps` in `Src/views/lib/UniscribeSegment.cpp` normalised **every prefix** of a range from scratch:

```cpp
for (int ich = 1; ich <= cchOrig; ++ich)
{
    StrUni stuPrefix(stuOrig.Chars(), ich);
    StrUtil::NormalizeStrUni(stuPrefix, UNORM_NFC);
    vichOrigToNfc[ich] = stuPrefix.Length();
}
```

Quadratic in the range length, and called once per analysed range, so the aggregate cost grows with the square of the paragraph length. For a 7,571-character paragraph that is tens of millions of characters normalised per layout pass.

The maps exist to translate between original and normalised offsets — Uniscribe shapes the normalised text while selections, cursor offsets and hit-testing live in the original text. They are only needed when normalisation actually changes the string. It fires when the text is not already NFC and the layout-pass cache is present (`UniscribeSegment.cpp:3157` before this change). Text reaching the Views layer is typically NFD, so for vernacular data that was the normal path, not an edge case.

This is a regression. `BuildNfcOffsetMaps`, `vichOrigToNfc` and `LayoutPassCache` all arrived together in 502b832c4 ("perf: Views engine render optimizations", #724, 2026-05-08). Before it there was no eager construction: translation was on demand via `OffsetInNfc`/`OffsetToOrig`, whose comment notes it "is not called with high frequency".

## Fix

Text that normalisation leaves unchanged needs no maps at all, because the mapping is the identity. So the analysis cache now stores only that text, and anything normalisation would rewrite falls back to the pre-existing on-demand translation. The eager map machinery is removed rather than rewritten.

The fix is net -21 lines. `ShapeRunCache` — the genuine ~3x improvement from the same commit — is untouched.

An earlier attempt kept caching non-NFC text and made `BuildNfcOffsetMaps` linear by splitting at ICU normalisation boundaries. It was measured and **rejected**: the decomposed proxy still took 3,468 / 2,174 ms against a 298 / 269 ms control, because a linear build still makes roughly one ICU call per character and is still invoked once per range. Fixing the constant factor could not fix the aggregate.

## Numbers

Cold render, milliseconds, 2 replicates. "cache-OFF control" is the same binary with `FW_PERF_P125_PATH2=0`, which is the configuration confirmed to resolve the problem on the real project.

| scenario | before | after | cache-OFF control |
|---|---|---|---|
| single-para-mixed-ws (decomposed, camus-shaped) | 3,468 / 2,174 | **325 / 334** | 277 / 289 |
| multi-line-wrap-single-ws (decomposed) | 832 / 835 | **168 / 409** | 175 / 180 |
| multi-ws | 487 / 458 / 452 | **148 / 167** | 143 / 300 |
| rtl-script | 116 / 121 / 122 | **78 / 77** | 78 / 93 |
| many-paragraphs | 147 / 158 | **82 / 90** | 80 / 162 |
| suite cold average | 383 / 315 | **145 / 154** | 130 / 157 |

After the fix, cache-on matches the cache-off control, i.e. the penalty is gone rather than reduced.

## Verification

- **Native tests: 301 passed, 0 failed, 0 errors.** The reported exit code -1 is a pre-existing post-completion hang that `Invoke-CppTest.ps1` already has explicit handling for; it is unrelated to this change.
- **Render transparency.** These caches must not change output. The six scenarios with pre-existing pixel differences are unchanged — complex 10, footnote-heavy 16, many-paragraphs 50, custom-heavy 5, medium 5, multi-book 5 — across four runs in both cache configurations, and no passing scenario began failing.
- **Build freshness confirmed by timestamp**, because `Views.mak` has no header-dependency tracking and a stale `Views.dll` reports success (see below).

## Why the existing tests did not catch this

Three safety nets were disconnected, all introduced or left broken by 502b832c4 itself.

1. **The benchmark scenarios could not reach the regressed path.** The suite sets `IsGraphiteEnabled = false` and uses ASCII almost throughout, and ASCII is identical under NFD and NFC, so the offset-map branch never executed. `multi-ws` and `rtl-script` reached it only incidentally, because Arabic harakat are canonically reordered. This PR adds three scenarios built from decomposed text — `nfc-composable-diacritics`, and decomposed forms of `single-para-mixed-ws` and `multi-line-wrap-single-ws` — which reproduce the defect at seconds scale.

2. **The cache unit tests had never run.** `Src/views/Test/TestViewCaches.h` was added by the same commit with 9 test methods across `TestColorStateCache`, `TestFontHandleCache` and `TestShapeRunCache`, but was never listed as a dependency of `Collection.cpp` in `testViews.mak` — and that list is the input to test collection. This PR registers it; all 9 pass.

3. **Header-only native edits silently produce stale binaries.** `Views.mak` via `Bld/_rule.mak` uses `{srcdir}.cpp{objdir}.obj:` rules with no header-dependency tracking, so editing a header leaves the previous `Views.dll` in place while the build reports success. Hit twice during this investigation. Not addressed here — it is a broader build-system defect and needs its own issue.

## Deliberately not included

- **The `FW_PERF_P125_PATH2` default is unchanged.** With the quadratic construction gone the cache is no longer harmful, so flipping a default would only obscure what this PR claims. The environment variable remains available.
- **`TextAnalysisCache` capacity and retention changes.** Measured no reproducible benefit and were aimed at a cost that turned out not to be the problem.
- **Re-enabling the cache for non-NFC text with lazily built maps.** That is the right long-term design — cache the expensive parts eagerly, build offset maps only when a translation is actually requested — but it changes the entry's contract and needs to prove it beats these numbers. It belongs with the optimisation work, not with a regression fix.

The three new scenarios have no accepted pixel baseline yet, so they report "Missing verified render baseline" until someone reviews and accepts the `.received.png` files. Their timings are still recorded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1056)


Src/Common/RootSite/RootSiteTests/RenderBenchmarkTestsBase.cs
```diff
@@ -927,6 +962,132 @@ protected void AddMultiWsSections(IScrBook book, int sectionCount,
...
+				"counted the cattlé at dusk", "kept the ôld story", "asked for a blessĩng"
+			};
+
+			section.ContentOA = stTextFactory.Create();
+			var paraBldr = new StTxtParaBldr(Cache) { ParaStyleName = ScrStyleNames.NormalParagraph };
+			paraBldr.AppendRun("1", StyleUtils.CharStyleTextProps(ScrStyleNames.ChapterNumber, m_wsEng));
+
+			for (int i = 0; i < sentenceCount; i++)
+			{
+				string sentence = $"{subjects[i % subjects.Length]} {predicates[i % predicates.Length]} {i + 1}. ";
```
papeh, line 996: I don't see why we need separate subjects and predicates if they're always combined into a single run.

Slick alternating run generator, though! 👍🏻

Src/Common/RootSite/RootSiteTests/RenderBenchmarkTestsBase.cs
```diff
@@ -927,6 +962,132 @@ protected void AddMultiWsSections(IScrBook book, int sectionCount,
...
+				"copied the ledger onto fresh parchmënt", "counted the flock past the old gatê",
+				"dyed the cloth a deep saffrõn", "shaped the jar on the slow wheël"
+			};
+
+			section.ContentOA = stTextFactory.Create();
+			var paraBldr = new StTxtParaBldr(Cache) { ParaStyleName = ScrStyleNames.NormalParagraph };
+
+			for (int i = 0; i < sentenceCount; i++)
+			{
+				string sentence = $"{subjects[i % subjects.Length]} {predicates[i % predicates.Length]} on day {i + 1}. ";
```
papeh, line 1084: same here

Src/views/lib/UniscribeSegment.cpp
```diff
@@ -3135,6 +3109,8 @@ int UniscribeSegment::CallScriptItemize(OLECHAR * prgchDefBuf, int cchBuf,
 
 	Vector<int> vichOrigToNfc;
 	Vector<int> vichNfcToOrig;
```
papeh, line 3111: These two vectors seem to be set to `{0}` in one place (around line 3150), then potentially copied into `LayoutCache.h`'s `m_vichOrigToNfc`, which is used in methods that return `cch` if the vector is empty or `0` if it is `{0}`. Can these vectors be removed?

papeh (commented):
Looks reasonable, although there is some opportunity for cleanup.

jasonleenaylor:
I am closing this one rather than asking for another round. That is not a verdict on
the diagnosis, which is correct and valuable — it is a verdict on the shape of the
branch. Details below, and what a version I would take looks like.

**The diagnosis is right, and I do not want it lost.**

`BuildNfcOffsetMaps` was `O(n)` full-string normalizations of `O(n)`-length prefixes,
invoked once per analysed range, with range length growing with paragraph length. That
is a genuine quadratic, the mechanism is visible in the deleted code, and naming
502b832c4 (#724) as the introducing commit is specific and checkable. The 3,468 ms ->
325 ms figure on the decomposed proxy is an order of magnitude and survives the noise in
the table comfortably.

There is also a stronger argument for deletion than the PR makes. Normalizing
`stuOrig[0..ich)` is not merely slow, it is *wrong* at boundaries: when `ich` splits a
surrogate pair, separates a base character from a following combining mark, or lands
inside a canonically-reordered combining sequence, the prefix normalization can yield a
length the full-string normalization never produces at that offset. The eager maps were
a latent offset-corruption bug as well as a performance bug. Please keep that in the
ticket.

**1. The change invalidates a set of guards without retiring them.**

`Store` is now reached only under `fTextEligibleForCache`, which is assigned
`fComputedTextIsNfc` (`UniscribeSegment.cpp:3133`), and it passes a hardcoded `true` for
`fTextIsNfc` (`:3299`). So `TextAnalysisEntry::m_fTextIsNfc` is provably true for every
entry that can now exist. That makes the following unreachable:

- `LayoutCache.h:29`, `:41`, `:55` — the `if (m_fTextIsNfc)` guard at the top of
  `RequestedNfcLength`, `OffsetInNfc` and `OffsetToOrig` always returns, so every
  `m_vichOrigToNfc` / `m_vichNfcToOrig` branch beneath it (`:31-35`, `:46-50`, `:59-63`)
  is dead.
- `LayoutCache.h:88-89` — both member vectors, now never populated.
- `LayoutCache.h:198`, `:248-265` — the two `Store` map parameters and the copy/clear
  behind them.
- `UniscribeSegment.cpp:3110-3111` — both locals, passed at `:3300-3301` as
  `Size() ? &v : NULL`, which now always evaluates to `NULL`.
- `UniscribeSegment.cpp:3141-3148` — the `{0}` fill for empty text. `Store` does run for
  empty text, but `m_fTextIsNfc` is `true`, so nothing can ever read those maps.
- The five-argument `OffsetInNfc` / `OffsetToOrig` overloads taking `pAnalysis`
  (`:1635`, `:1693`). `ppAnalysis` receives non-NULL only when the text was NFC, in
  which case both return at the identity guard before touching it. Pure indirection now.

`papeh` asked about exactly this on `:3111` on 20 August and has had no reply. The
problem is not tidiness. `TextAnalysisEntry` still advertises an offset-translation
contract that the cache can no longer be given the data to honour, and the next person
reading `LayoutCache.h:25-64` will reasonably conclude the cache handles non-NFC text.
A safety net that is still wired up but can no longer fire is precisely the failure mode
this PR was written to fix.

There is a second, quieter invalidation. The fallback this change re-enables —
`UniscribeSegment::OffsetToOrig` at `:1651` — re-fetches and re-normalizes a growing
prefix in a loop, and its own comment concedes it is only safe because it "is not called
with high frequency." The branch permanently routes decomposed vernacular text, the
population FieldWorks actually has, onto that path, and never re-establishes the
frequency assumption for it. Trading a measured quadratic for an unmeasured one needs
evidence, and none is offered.

**2. Existing methods the branch should have used.**

Credit first: `RobustIO.DeleteDirectoryAndContents`, `StrUtil::NormalizeStrUni` as the
single normalization entry point, `TsStringUtils.MakePropsBldr`, `StTxtParaBldr` and the
`ServiceLocator` fixture seams are all used correctly rather than hand-rolled. (And to
save anyone raising it: ICU `unorm2_isNormalized` would not help at `:3130`, because the
NFC buffer has to be produced for Uniscribe regardless.)

Missed:

- `CustomIcu.GetIcuNormalizer(FwNormalizationMode.knmNFD).Normalize(...)` — used in this
  pattern in `ConfiguredExport.cs:589`, `ValidCharactersDlg.cs:985`, `XDumper.cs:164`,
  `LiftExporter.cs:617` and elsewhere. The new fixtures encode decomposition as raw NFD
  bytes in `.cs` source (`RenderBenchmarkTestsBase.cs:979-988`, `:1034-1038`,
  `:1067-1077`). I verified they are genuinely decomposed today — 37 combining marks,
  zero precomposed characters. But that correctness lives only in the file's bytes: any
  formatter, editor or agent pipeline that re-normalizes the file to NFC silently turns
  all three scenarios back into the "cannot reach the regressed path" case, with no test
  failure and no visible diff intent. Build them at runtime from readable NFC literals,
  or at minimum assert decomposition once per fixture.
- `RealDataTestsBaseCleanupTests.InvokeRunSetupFailureCleanup`
  (`RealDataTestsBaseCleanupTests.cs:47-56`) — an existing, working pattern for
  reflection-testing a private static member of the very class this branch extends.
  `HoldsOnlyGeneratedProjectData` is a private static pure predicate over a path: the
  cheapest possible unit test, and it has none.
- The `TestViewCaches.h` suites (`:18`, `:66`, `:182`). The branch wires that file into
  `testViews.mak` and thereby proves `LayoutCache.h` classes are unit-testable in the
  native harness — then adds no `TextAnalysisCache` or `TextAnalysisEntry` suite. The
  nine tests that now run cover the colour, font-handle and shape-run caches. None cover
  the code this branch changed.

**3. The branch conflates four changes.**

- A native performance fix in `UniscribeSegment::CallScriptItemize` (43 lines).
- A rewrite of the guard in front of a recursive delete in `RealDataTestsBase`
  (commit 1, unrelated to LT-22674).
- New render-benchmark scaffolding and scenario data (161 + 15 lines).
- Three accepted pixel baselines (commit 5), which the PR body still says do not exist.

The second of those is why this is a close rather than a change request.
`DeleteProjectDirectory` calls `RobustIO.DeleteDirectoryAndContents` on a directory
derived from a user-configurable projects path. The sentinel file existed so that delete
could never fire without positive proof the fixture created the directory. This branch
replaces proof with inference — no top-level `*.fwdata` other than my own — and the
inference is weaker than it looks: it inspects only top-level `*.fwdata`, while a
FieldWorks project directory also holds `LinkedFiles` (user media),
`WritingSystemStore`, `BackupSettings`, `.fwbackup` archives and `.lift` exports. A
directory holding a user's `LinkedFiles` and no foreign `.fwdata` passes and is deleted
recursively.

Worse, `RealDataTestsBase.cs:360-364` writes the sentinel and *then* deletes. If
`RobustIO.DeleteDirectoryAndContents` returns false and `:367` throws, the fixture has
permanently stamped a directory it never verified as fixture-owned — and because
`:339` only runs the `HoldsOnlyGeneratedProjectData` check when the sentinel is
*absent*, the next run deletes that directory with no check at all. The write buys
nothing: the next statement destroys the file.

`EnsureSafeProjectDirectory` requiring a worktree-hashed exact path is real mitigation,
and it is why I am describing this as risk rather than a live defect. But it is the last
fence in front of a recursive delete, it is being loosened, it is untested, and it is
riding in a PR titled as a performance fix. Someone searching `git log` later for why a
project directory vanished will not find it here.

**Why close rather than iterate.**

Individually these are all fixable. Together they mean the branch would need the dead
machinery resolved, a native cache test suite added, the fixture NFD-asserted, the
destructive-delete change extracted, its ordering bug fixed and tested, the body
corrected, and eight `-CommentHygiene` violations cleared
(`RealDataTestsBase.cs:343` is `comment-too-long` at 403 characters against a 200
budget; `:343`, `:344`, `:347` and `RenderBenchmarkTestsBase.cs:552`, `:560`, `:568`,
`:965`, `:1009`, `:1020`, `:1053` are `comment-line-too-long` at 101-128 columns — four
of the five commits are agent-co-authored, so that check was not optional). That is not
a revision cycle; that is a different branch. Reviewing it as a delta against this one
would mean re-reviewing everything anyway.

**A more careful redo would very likely be accepted.** What I would take:

1. The `UniscribeSegment` change on its own, with the dead machinery either removed in
   the same commit or explicitly kept with a one-line reason.
2. A `TextAnalysisCache` / `TextAnalysisEntry` suite alongside the three that this
   branch showed are wirable: a non-NFC range is not stored, an NFC range is stored and
   hit, `RequestedNfcLength` on a `Covers`-matched shorter request.
3. Either evidence that the re-enabled `OffsetToOrig` fallback is genuinely low-frequency
   for decomposed paragraphs, or the `Covers`-exact-length variant that keeps the itemize
   and NFC-buffer win for non-NFC text without needing any map at all.
4. One after-measurement on a real project. The 10+ second Maasai-Parser / camus4-D case
   is measured before the fix and never after; that is the number I most want and the
   cheapest one to get.
5. The `RealDataTestsBase` work as its own PR, with the ordering bug fixed and a test.
   Consider pre-writing the sentinel before `CreateNewLangProj` instead —
   `m_projectDirectory` is already known at `:52` — which closes the interrupted-run
   window without weakening the check at all.

The benchmark scenarios are good work and I would take them in either branch. Thank you
for chasing this down to the introducing commit; the analysis is the hard part and you
have already done it.

PR #919, opened 2026-05-29
Fix intermittent access violation in RenderBenchmark tests

jasonleenaylor:
## Summary
- Add `[Apartment(ApartmentState.STA)]` to `RenderTimingSuiteTests` and `RenderVerifyTests`
- Views COM objects (`VwGraphicsWin32Class`, `_VwRootBoxClass`) require STA threading — without the attribute, NUnit runs these on an MTA thread, causing intermittent access violations during `RootBox.Layout` in CI
- The sibling `StVcTests` fixture already had this attribute correctly

## Test plan
- [ ] Verify RenderBenchmark tests pass consistently in CI (previously intermittent AV failures)
- [ ] Confirm no regression in other RootSite tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/919)

PR #1057, opened 2026-08-12
Fix buffer overread setting line-break text in LgLineBreaker

jasonleenaylor:
## Quick Summary

- Fixes the intermittent `VerifyScenario("long-prose")` / `RunBenchmark("long-prose")` CI failure — a `COMException: Access violation` from `VwRootBox.Layout` that hit unrelated PRs roughly weekly (five identical failures Jul 16–Aug 10; [example run](https://github.com/sillsdev/FieldWorks/runs/93608132221)).
- `LgLineBreaker::put_LineBreakText` now copies its text with the length-taking `UnicodeString` constructor instead of assigning the bare pointer, which scans an unterminated buffer for a null terminator.
- Adds a native regression test covering the copy length and word-break lookup.

## CI-ready checklist

  ```powershell
  git fetch origin
  git diff --check --cached
  ```
- [ ] If this is core-developer AI-assisted work, I followed `Docs/workflows/ai-pr-workflow.md` and ran `pr-preflight` or the equivalent branch-readiness review before requesting review.

**Why it failed.** `UniscribeEngine::FindBreakPoint` hands `put_LineBreakText` a text buffer that carries no null terminator. Storing that text by assigning the bare pointer made ICU scan past the end of the buffer looking for a terminator; the read only faults when the memory after the buffer happens to be unmapped, which is what made the crash rare and heap-layout dependent. It surfaced only in the `long-prose` scenario because paragraphs under 1000 characters go through a stack buffer (where the overread never faults), and that scenario has by far the largest heap-allocated paragraph text. The overread shipped with the ICU 70 upgrade (6592f1ea6, Feb 2022), so real projects with paragraphs over ~1000 characters have been exposed to it since.

**Verification.** Made the crash deterministic with full PageHeap: the scan faulted at exactly `text[cch]` inside `icuuc70!UnicodeString::doAppend` on every long-prose layout before the fix, and not at all after. All 30 render baseline tests pass unchanged (the fix does not alter break positions), and the new native `TestLgLineBreaker` test passes.

**Related history.** The STA attribute added by #919 for these same AVs remains correct and should stay: Views COM objects are apartment-threaded, so the fixtures need `[Apartment(ApartmentState.STA)]` regardless — but the AVs it chased were most likely this overread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1057)

### DIFF
--- a/Src/views/Test/TestLgLineBreaker.h
+++ b/Src/views/Test/TestLgLineBreaker.h
@@ -65,6 +65,30 @@ namespace TestViews
 			}
 		}
 
+		void testPutLineBreakTextUsesLength()
+		{
+			unitpp::assert_true("m_qlb", m_qlb.Ptr());
+			// The text sits inside a larger buffer with no NUL terminator; the characters
+			// after cch must not become part of the break text, and the copy must not
+			// scan past cch looking for a terminator.
+			OLECHAR rgch[] = { 'o','n','e',' ','t','w','o',' ','s','i','x','X','X','X','X','X' };
+			const int cch = 11; // "one two six"
+			CheckHr(m_qlb->put_LineBreakText(rgch, cch));
+
+			OLECHAR rgchOut[32];
+			int cchOut = 0;
+			CheckHr(m_qlb->GetLineBreakText(32, rgchOut, &cchOut));
+			unitpp::assert_eq("break text length", cch, cchOut);
+			unitpp::assert_true("break text content",
+				memcmp(rgch, rgchOut, cch * isizeof(OLECHAR)) == 0);
+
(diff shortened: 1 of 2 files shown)


thejambi (approved):
:lgtm:

> @thejambi reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1057#-:-Ozs0n7E1DVDrup-6ZGr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jasonleenaylor).

PR #1059, opened 2026-08-12
Chore: Bump 15 dependencies from 3 dependabot PRs

jasonleenaylor:
Combines #1035, #1036, and #1037 into one pull request, so this month's
dependency updates cost one CI cycle instead of three — then hardens the one
integration those bumps put at risk.

Each Dependabot commit is cherry-picked onto current `main` with its author and
author date preserved. The bodies on #1035 and #1036 fail the
`Check commit messages` job as Dependabot wrote them (markdown link lines exceed
the 80-column body limit), so those two were trimmed to satisfy
`.github/commit-guidelines.md`; #1037's body was already compliant and is
byte-identical to the original.

## Includes

| Source | Updates |
|--------|---------|
| #1035 | `softprops/action-gh-release` 3.0.1 -> 3.0.2, `lycheeverse/lychee-action` 2.8.0 -> 2.9.0 |
| #1036 | `actions/setup-dotnet` 5 -> 6 (**major**) |
| #1037 | 12 nuget minor/patch bumps, including `SIL.Core` 18.0.0-beta0027 -> beta0030 and `SIL.LCModel` 11.0.0-beta0173 -> beta0176 |

`actions/setup-dotnet` 5 -> 6 is the only major in the batch.
`.github/dependabot.yml` ignores nuget majors but lets github-actions majors
through ungrouped, which is why it arrived on its own.

## Beyond the bumps

Three follow-on commits correct what the batch would otherwise have broken:

- **`Microsoft.Extensions.DependencyModel` stays pinned at 9.0.16.** The
  nuget-minor group proposed 9.0.18, silently reverting the deliberate pin from
  #1000 (9.0.17 broke ICU initialization in the net48 test host). Every
  dependant declares only a floor that 9.0.18 clears, so nothing in the package
  graph guards this pin — it is now also excluded in `.github/dependabot.yml`
  so it stops being re-proposed monthly. Pin comments that described floors as
  exact requirements were corrected from the actual `.nuspec` ranges.
- **LCModel's new IoC container is internalized into the Paratext lexicon
  plugin.** SIL.LCModel beta0176 replaced StructureMap with
  `Microsoft.Extensions.DependencyInjection`, which surfaced as new external
  references of the ILRepack-merged `FwParatextLexiconPlugin.dll`. Paratext
  loads that plugin into its own process, where FieldWorks' binding redirects
  do not apply and Paratext ships no copy of these assemblies — any version
  drift broke the plugin only inside Paratext, invisibly to every FieldWorks
  test. Verified in the liblcm source that no DI type is on LCModel's public
  API and the container bootstrap uses no reflection-by-name, so the assemblies
  (plus `Microsoft.Bcl.AsyncInterfaces`, which merging them promotes) are
  internalized via `ILRepack.targets` and their packages pinned. `Build/Src`
  overrides `Microsoft.Bcl.AsyncInterfaces` to the 10.0.4 floor that
  `Microsoft.Build.Utilities.Core`'s `System.Text.Json` chain requires.

## Validation

- Commit messages and whitespace: all commits pass the
  `.github/commit-guidelines.md` rules and `git log --check`.
- `build.ps1 -BuildTests` (native seeded, `-SkipNative`): 0 warnings, 0 errors.
- `project.assets.json` confirms the bumped SIL versions actually restored
  (`SIL.Core/18.0.0-beta0030`, `SIL.LCModel/11.0.0-beta0176`) rather than being
  shadowed by local dev packages.
- Full `test.ps1` (no filter) on this dependency set: 5,739 managed tests,
  0 failures.
- Merged plugin verified post-internalization: 53 external references, zero
  `Microsoft.Extensions.*`/`Bcl` entries; `FwParatextLexiconPluginTests` 31/31.
- Not covered by any test here or in CI: the merged plugin executing inside a
  real Paratext. **A manual smoke of the lexicon plugin in Paratext is
  recommended before this leaves draft**, since the internalized DI container's
  first real execution happens there.

Note for reviewers: the three source PRs all show a red `Build Debug and run
tests`, but the failing step is `Verify Codecov upload succeeded` — Dependabot
PRs do not receive `secrets.CODECOV_TOKEN`, so they cannot go green regardless
of the bumps. Their tests passed. This PR runs from a repository-owned branch
and does not hit that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1059)


jasonleenaylor:
Pushed `cadfc264d`: restores `Microsoft.Extensions.DependencyModel` to 9.0.16.

The nuget-minor group bumped it to 9.0.18, but `Directory.Packages.props` pins it deliberately — #1000 established that pin after reproducing a `TypeInitializationException` on `Icu.NativeMethods` in the .NET Framework test host, and the comment above the line documents the retest procedure. 9.0.18 has not been retested, and the green CI above does not clear it: that procedure calls for a full `test.ps1`, while CI runs with `TestCategory!=LongRunning&TestCategory!=ByHand&TestCategory!=SmokeTest&TestCategory!=DesktopRequired`.

The other 14 bumps are unchanged. Also refreshed three pin comments that still named 9.0.17 for packages that did move to 9.0.18.

Checked and clean: no `.csproj` declares a version for any pinned package; the Build/Src `Include`/`Update` layering invariant still holds; and none of the 31 property-driven version entries were rewritten into literals.

Follow-up worth doing separately: add `Microsoft.Extensions.DependencyModel` to the `ignore:` block in `.github/dependabot.yml`, or this returns every month.

jasonleenaylor:
Pushed `75540cf66`: adds `Microsoft.Extensions.DependencyModel` to the nuget `ignore:` block, and rewrites three pin comments that described floors as though they were exact requirements.

I evaluated every pin in `Directory.Packages.props` whose comment cites an external dependency, reading the actual ranges from the `.nuspec` files under `packages/`. **They are all floors**, so no further reverts are warranted:

| Depending package | Declared requirement | Pin now | Clears? |
|---|---|---|---|
| ParatextData 9.5.0.20 | `System.Drawing.Common >= 9.0.9` | 9.0.18 | yes |
| SIL.LCModel.Core | `System.Drawing.Common >= 6.0.0` | 9.0.18 | yes |
| ProDotNetZip 1.20.0 | `System.Security.Permissions >= 8.0.0` | 9.0.18 | yes |
| ParatextData 9.5.0.20 | `System.Memory >= 4.6.3` | 4.6.3 | yes |
| SIL.Machine 3.8.2 | `System.Text.Encoding.CodePages >= 10.0.3` | 10.0.3 | yes |
| Avalonia.Skia 11.3.17 | `HarfBuzzSharp >= 8.3.1.1` | 8.3.1.1 | yes |
| ParatextData / icu.net / LCModel.Core | `DependencyModel >= 9.0.9 / >= 2.0.4` | 9.0.16 | yes |

Unbracketed `version="X"` in a nuspec is a minimum, not an exact version — exactly one bracketed range exists across all 219 cached packages. With `CentralPackageTransitivePinningEnabled` the pin must sit at or above every floor, and the failure mode is pinning too *low* (`NU1109`), which the `HarfBuzzSharp` comment documents. Pinning above a floor is safe by construction.

**The DependencyModel pin is not constraint-driven at all.** Every declared range permits 9.0.18. It is held at 9.0.16 purely by the runtime break from #1000, which no manifest can express — which is why it is the only pin that needs a dependabot exclusion, and why the package graph would never have blocked it.

Also worth noting: `ParatextData` is itself in Dependabot's scope, so a future batch could move the floors these comments quote.

.github/dependabot.yml
```diff
@@ -32,6 +32,15 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
+      # Microsoft.Extensions.DependencyModel is held at 9.0.16 by an observed
```
johnml1135: Is this the best place for this information?  I would imagine it is either redundant or should be moved to next to the actual dependency declaration.

johnml1135: AI recommendation:

```
  # Pinned at 9.0.16 due to TypeInitializationException on Icu.NativeMethods in
  # .NET Framework test host. See Directory.Packages.props for retest procedure.
```

Directory.Packages.props
```diff
@@ -24,33 +24,39 @@
...
+		<PackageVersion Include="System.Drawing.Common" Version="9.0.18" />
+		<PackageVersion Include="System.Reflection.Metadata" Version="10.0.10" />
 		<!-- System.Resources.Extensions: required for non-string resources (images, icons)
 			 in .NET Framework SDK-style projects. -->
-		<PackageVersion Include="System.Resources.Extensions" Version="9.0.17" />
-		<!-- DependencyModel: other current-line pins in this block move to 9.0.17,
-			 but this stays at 9.0.16. icu.net and SIL.LCModel.Core (liblcm) both float
-			 a 2.0.4 floor, ParatextData wants >= 9.0.9, and 9.0.17 throws
+		<PackageVersion Include="System.Resources.Extensions" Version="9.0.18" />
+		<!-- DependencyModel: other current-line pins in this block move to 9.0.18,
```
johnml1135: These comments are long - can we keep them under 200 chars?  I am making the PR for "200 chars or less".  Should that change or should these change?

johnml1135 (changes requested):

Directory.Packages.props
jasonleenaylor: 200 chars might be tight, but I made a pass on this file and most could seem to stay under it.

.github/dependabot.yml
jasonleenaylor: Done.

jasonleenaylor (commented):
> @jasonleenaylor made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1059)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on johnml1135).

PR #1064, opened 2026-08-13
LT-22702: Escape custom list range ids at the point they are written

imnasnainaec:
Fixes [LT-22702](https://jira.sil.org/browse/LT-22702).

Companion to #1063. The two diffs do not overlap.

## Quick Summary

- A custom possibility list's name is written into two XML **attributes**: the `range/@id` in
  the `.lift` header, and the `range/@id` of the same range in the `.lift-ranges` file.
- The exporter escaped that name **once, on collection**, with the element-content rules
  (`MakeSafeAndNormalizedXml`), rather than at each write site with the attribute rules --
  which is both the wrong rule and the wrong place.
- This branch holds the collected names unescaped and escapes at each write site, matching what
  the rest of the exporter already does for every other range and range-element id.

**Two defects fall out of escaping on collection:**

1. **The header disagrees with its own ranges file.** The header write site escapes again, so a
   list named `Birds & Beasts` is emitted as `Birds &amp;amp; Beasts`, while the ranges file
   kept a single escape.
2. **The ranges file can be unparseable.** The ranges write site does not escape at all, and
   element rules never escape `"`, so a list named `So-called "words"` closes the id attribute
   early and leaves the `.lift-ranges` unreadable by any conforming parser.

## The change

**Collect raw.** Four sites stop pre-escaping, so the two name maps hold what the user typed:

- `GetCustomListsAndReferencedLists`
- `GetCustomListsNotYetAdded`
- `AddPossListRefdByField`
- `MapCmPossibilityListGuidsToLiftRangeNames`

The `m_ListsGuidToRangeName` map already held standard range names raw via `MapGuidToRange`, so
this also stops it holding a mix of escaped and unescaped values.

**Escape at the write site.** `WritePossibilityListAsRange` now puts its `rangeId` through
`MakeSafeAndNormalizedAttribute`, hoisted into a local so the normalizer runs once and both
branches are visibly the same.

**Leave the two already-correct write sites alone.** The header `<range id href>` and the
custom-field spec string need no change once their input is raw; they stop double-escaping on
their own.

**Record the invariant.** It is stated on the two map declarations, since "held unescaped" is
the kind of thing a later change would otherwise re-break.

## Scope

Only custom and referenced possibility lists are affected. `WritePossibilityListAsRange` has
six callers; the other five pass `RangeNames` constants, where the helper is a no-op.

**Not addressed here.** `WriteRangeRefsForListsReferencedByFields` writes a header
`<range id href>` for every collected list, while `WritePossibilityListAsRange` returns before
writing anything for a list with no possibilities. So an empty custom list still yields a header
reference to a range that has no body in the ranges file. That is pre-existing and left alone;
the only effect here is that the dangling id now carries the corrected spelling.

<details>
<summary><b>Sweep: every other range id write already escapes and normalizes</b></summary>

I swept every `<range id=` and `<range-element id=` write in `LiftExporter.cs` to be sure this
is the whole of it.

- `WritePossibilityRangeElement` -- covers semantic domains, status, locations, people,
  anthropology, translation tags, production restrictions and custom lists. Escapes and
  normalizes both its range ids and its range-element ids.
- The four part-of-speech sub-range writers -- `-feature-value`, `-slot`, `-infl-class`,
  `-stem-name`. Same.
- Everything else is a literal or a `RangeNames` constant.

</details>

## Effects worth knowing

**This is not a normalization change.** The collected names were already NFC-normalized on the
way in, and remain so -- `MakeSafeAndNormalizedAttribute` normalizes as well as escapes. No id
changes spelling as a result of this branch; only escaping changes, and only for names holding
`&`, `<`, `>` or `"`. (#1063 is the branch that changes spelling.)

**FLEx's own importer is unaffected.** A custom list's range falls through
`ProcessRangeElement`'s switch to `default:`, which handles only the `-slot`, `-infl-class`,
`-feature-value` and `-stem-name` suffixes and otherwise does nothing but a `Debug.WriteLine`.
FLEx does not import custom-list ranges at all, so there is no re-import behavior to check.
This matters only to third-party consumers of the files, and to what lands in a shared repo.

**The `qaa-x-spec` custom-field definition also changes.** `GetCustomFieldDefinition` appends
the map value raw into `; range={0}`, and `GenerateCustomFieldSpecs` escapes the whole spec
string once at its write site. With the map now holding raw names, a custom list name containing
one of the four characters reaches that form single-escaped where it used to be double-escaped.
This is the same defect and the same correction as the header range ref, and it brings custom
lists into line with the standard list names, which `MapGuidToRange` has always stored raw. It
is called out separately because it touches the header field definitions, not just ranges.

**LIFT Send/Receive sees a whole-record delete plus add.** `range` is keyed on `@id` by both the
Chorus element strategy and by `XmlMergeService`'s record splitting, and unlike `range-element`
it has no guid to fall back on.

- For any project whose custom list name contains one of the four characters, the first export
  after upgrading rewrites that id once. Chorus models that as a delete plus an add of the whole
  `<range>` record, taking every `<range-element>` inside it along -- a large single diff.
- It is a one-time event per affected list. The new spelling is stable, unlike a normalization
  change, which an un-upgraded client would keep undoing.
- A project whose list name contains a `"` has been committing an unparseable ranges file until
  now.

## Testing

Two regression tests in `LiftExportTests`, each covering one of the two defects. Both were
confirmed to fail on `main` before the fix.

1. `LiftExportRanges_CustomListRangeIdWithQuoteIsEscapedForAnAttribute` -- a custom list named
   `So-called "words"`, asserting the ranges document parses and the id reads back as the name
   stored. Without the fix this fails with
   `XmlException: 'words' is an unexpected token. Expecting white space.`
2. `LiftExportRanges_CustomListRangeIdWithAmpersandIsEscapedExactlyOnce` -- a custom list named
   `Birds & Beasts`, asserting the ranges id reads back as the name stored and that the `.lift`
   header names the range the same way. Without the fix the header and the ranges file disagree.

Both drive the real exporter over a real custom list and custom field, through a shared
`ExportWithCustomList` helper that runs the `.lift` pass and the ranges pass on one exporter --
the ranges pass writes only the lists the `.lift` pass collected.

`.\test.ps1 -SkipNative -TestProject LexTextControlsTests -TestFilter "FullyQualifiedName~LiftExportTests"`:
**7 passed, 0 failed**, including the pre-existing `LiftExport` (which asserts the header range
ids for both a referenced and an unreferenced custom list) and
`LiftExport_MultiParagraphWithAmpersandExports`. The full suite and the native tests have not
been run locally; leaving this as a draft for CI.

### Test fixture resets

The fixture also now resets its five shared custom-field collections in `[SetUp]`, alongside the
maps it already reset there. NUnit shares one fixture instance across a class, so each of them
outlived the cache it indexes and grew by several entries per test, while `VerifyExport` and
`VerifyExportRanges` index them positionally.

<details>
<summary>Why each of the five needed resetting</summary>

`m_customListsGuids` had to be reset for the new tests to pass: guids are minted per object, so
`[0]` and `[1]` named lists in a cache that had already been disposed.

The four flid lists (`m_customFieldEntryIds`, `m_customFieldSenseIds`,
`m_customFieldAllomorphsIds`, `m_customFieldExampleSentencesIds`) were still passing, but only
because `AddCustomField` allocates from `clid * 1000 + 500` upward per metadata cache, so a fresh
cache plus an unchanged `[SetUp]` order reassigns the same flids and the stale entries happened
to hold the values the current cache would assign. Resetting them too retires that unwritten
invariant, which a reordered or added custom field would otherwise break into a confusing
failure.

</details>

## CI-ready checklist

- [x] Commit messages follow `.github/commit-guidelines.md`.
- [x] No whitespace warnings locally (`git log --check` and `git diff --check` are clean).
- [x] Builds locally, and the touched test fixture passes.
- [x] Jira ticket filed: LT-22702.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1064)

### DIFF
--- a/Src/LexText/LexTextControls/LexTextControlsTests/LiftExportTests.cs
+++ b/Src/LexText/LexTextControls/LexTextControlsTests/LiftExportTests.cs
@@ -1135,6 +1135,84 @@ namespace LexTextControlsTests
 			VerifyExportRanges(xdocRangeFile);
 		}
 
+		///--------------------------------------------------------------------------------------
+		/// <summary>
+		/// A custom list name reaches the ranges file as a range id, where it needs the escaping
+		/// rules for an attribute. It used to be escaped once as element content on the way into
+		/// the exporter's map and never again, so a name holding a quotation mark left the ranges
+		/// document unparseable.
+		/// </summary>
+		///--------------------------------------------------------------------------------------
+		[Test]
+		public void LiftExportRanges_CustomListRangeIdWithQuoteIsEscapedForAnAttribute()
+		{
+			const string ksListName = "So-called \"words\"";
+			XmlDocument xdocLift;
+			XmlDocument xdocRangeFile;
+			var customList = ExportWithCustomList(ksListName, out xdocLift, out xdocRangeFile);
+
+			var range = xdocRangeFile.SelectSingleNode(string.Format("//range[@guid='{0}']", customList.Guid));
+			Assert.That(range, Is.Not.Null, "the custom list must be written as a range");
+			Assert.That(range.Attributes["id"].Value, Is.EqualTo(ksListName),
+				"the range id must read back as the name stored");
+		}
+
+		///--------------------------------------------------------------------------------------
+		/// <summary>
+		/// The same name is written into the .lift header and into the ranges file, so the two must
+		/// agree. Escaping once on the way in and again at the header write site turned an ampersand
+		/// into &amp;amp; there while the ranges file kept a single escape.
+		/// </summary>
+		///--------------------------------------------------------------------------------------
+		[Test]
+		public void LiftExportRanges_CustomListRangeIdWithAmpersandIsEscapedExactlyOnce()
+		{
+			const string ksListName = "Birds & Beasts";
+			XmlDocument xdocLift;
+			XmlDocument xdocRangeFile;
+			var customList = ExportWithCustomList(ksListName, out xdocLift, out xdocRangeFile);
+
+			var rangesRange = xdocRangeFile.SelectSingleNode(string.Format("//range[@guid='{0}']", customList.Guid));
+			Assert.That(rangesRange, Is.Not.Null);
+			Assert.That(rangesRange.Attributes["id"].Value, Is.EqualTo(ksListName),
+				"one level of escaping, undone by the parser, must give back the name stored");
+			var headerIds = xdocLift.SelectNodes("//header/ranges/range/@id");
+			Assert.That(headerIds, Is.Not.Null);
+			Assert.That(headerIds.Cast<XmlNode>().Any(id => id.Value == ksListName),
+				"the .lift header must name the range the same way the ranges file does");
+		}
+
+		/// <summary>
+		/// Exports both files from one exporter, so that the custom list is discovered by the .lift
+		/// pass and is therefore written by the ranges pass.
+		/// </summary>
+		private ICmPossibilityList ExportWithCustomList(string listName, out XmlDocument xdocLift, out XmlDocument xdocRangeFile)
+		{
+			ICmPossibilityList customList = null;
+			NonUndoableUnitOfWorkHelper.Do(m_cache.ActionHandlerAccessor, () =>
+			{
+				customList = AddCustomList(listName);
+				MakeCustomField("CustomFieldForCustomList", LexEntryTags.kClassId,
+					WritingSystemServices.kwsAnal, CustomFieldType.ListRefAtomic, customList.Guid);
+			});
+			var exporter = new LiftExporter(m_cache);
+			xdocLift = new XmlDocument();
+			using (var w = new StringWriter())
+			{
+				// SUT: the .lift pass populates the map the ranges pass writes from.
+				exporter.ExportLift(w, LiftFolder);
+				xdocLift.LoadXml(w.ToString());
+			}
+			xdocRangeFile = new XmlDocument();
+			using (var w = new StringWriter())
+			{
+				// SUT
+				exporter.ExportLiftRanges(w);
+				xdocRangeFile.LoadXml(w.ToString());
+			}
+			return customList;
+		}
+
 		/// <summary>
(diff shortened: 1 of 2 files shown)


Src/LexText/LexTextControls/LexTextControlsTests/LiftExportTests.cs
```diff
@@ -493,6 +493,12 @@ public void CreateMockCache()
 			m_mapPartsOfSpeech.Clear();
 			m_mapAcademicDomains.Clear();
 			m_mapPublications.Clear();
+			// NUnit shares one fixture instance across the tests, so these outlive the cache they index.
```
papeh: should this comment go at the beginning of this method?

Src/LexText/LexTextControls/LiftExporter.cs
```diff
@@ -56,8 +56,14 @@ public class LiftExporter
...
 		/// <summary>
 		/// This contains the possibility lists that are custom or that are referenced by a custom field.
+		/// Range names are held unescaped, so that each is escaped by the rules of the context it is
+		/// written into.
 		/// </summary>
 		private Dictionary<Guid, string> m_CmPossListsReferencedOrCustom = new Dictionary<Guid, string>();
+		/// <summary>
+		/// Maps a possibility list to the name of its LIFT range. Range names are held unescaped, so
+		/// that each is escaped by the rules of the context it is written into.
+		/// </summary>
```
papeh, line 66: I'm not sure we need this comment. The field name is self-documenting, and the usage is documented by example.

papeh (approved):

Src/LexText/LexTextControls/LexTextControlsTests/LiftExportTests.cs
imnasnainaec: Done.

Src/LexText/LexTextControls/LiftExporter.cs
imnasnainaec: Comment trimmed.

PR #1066, opened 2026-08-14
Record Avalonia UI mode in crash reports

johnml1135:
## Quick Summary

Crash reports now record whether Avalonia ("New") UI mode is active and which
tools are opted out via `UIModeDisabledTools`, alongside the OS/.NET environment
info `ErrorReporter` already collects.

The value tracks the whole session, not just window construction:
`FwXWindow.SeedUIModeProperties` seeds it at startup, and both live-toggle paths
--- `LexOptionsDlg.m_btnOK_Click` (WinForms) and
`AvaloniaOptionsDialogLauncher.ApplyUiModeLive`/`ApplyDisabledToolsLive`
(Avalonia) --- push the same update into `ErrorReporter` that they already push
into the `PropertyTable`. An OK-click that does not change the mode records
nothing, so it cannot fabricate a value.

`LexOptionsDlg` has no per-tool disabled-tools editor, so only `AvaloniaUIMode`
is updated there; `ApplyDisabledToolsLive` exists only on the Avalonia side,
where that setting is actually edited.

Property names `AvaloniaUIMode`/`AvaloniaDisabledTools` were chosen to read
clearly in a raw crash-report dump rather than to mirror the internal
`UIMode`/`UIModeDisabledTools` `PropertyTable` keys.

Standalone branch off `main`, not stacked.

## Test plan

- `./build.ps1 -CommentHygiene` --- clean.
- `./test.ps1 -TestProject "Src/xWorks/xWorksTests" -TestFilter "FwXWindowUIModeSeedingTests"`
  --- 8 tests, all pass.
- `./test.ps1 -TestProject "Src/LexText/LexTextControls/LexTextControlsTests" -TestFilter "LexOptionsDlgTests|AvaloniaOptionsDialogLauncherTests"`
  --- 46 tests, all pass, including the assertions on
  `ErrorReport.Properties["AvaloniaUIMode"]`/`["AvaloniaDisabledTools"]` after a
  live toggle and the unchanged-selection case.
- Full repo `test.ps1` was not run; CI covers it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1066)

### DIFF
--- a/Src/LexText/LexTextControls/LexOptionsDlg.cs
+++ b/Src/LexText/LexTextControls/LexOptionsDlg.cs
@@ -19,6 +19,7 @@ using SIL.FieldWorks.Common.FwUtils;
 using SIL.LCModel.Utils;
 using SIL.LCModel;
 using SIL.PlatformUtilities;
+using SIL.Reporting;
 using SIL.Settings;
 using SIL.Utils;
 using XCore;
@@ -158,6 +159,8 @@ namespace SIL.FieldWorks.LexText.Controls
 						m_propertyTable.SetPropertyPersistence(UIModePropertyName, false);
 					}
 				}
+				// Keep the crash-report value in step with a live toggle, not just the startup seed.
+				ErrorReporter.AddProperty("AvaloniaUIMode", newUiMode);
 			}
 
 			m_sNewUserWs = m_userInterfaceChooser.NewUserWs;
--- a/Src/LexText/LexTextControls/LexOptionsDlg.cs
+++ b/Src/LexText/LexTextControls/LexOptionsDlg.cs
@@ -158,9 +158,8 @@ namespace SIL.FieldWorks.LexText.Controls
 						m_propertyTable.SetProperty(UIModePropertyName, newUiMode, true);
 						m_propertyTable.SetPropertyPersistence(UIModePropertyName, false);
 					}
+					ErrorReporter.AddProperty("AvaloniaUIMode", newUiMode);
 				}
-				// Keep the crash-report value in step with a live toggle, not just the startup seed.
-				ErrorReporter.AddProperty("AvaloniaUIMode", newUiMode);
 			}
 
 			m_sNewUserWs = m_userInterfaceChooser.NewUserWs;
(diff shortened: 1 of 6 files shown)


papeh:
As you mention, this records the values only at startup. This would be confusing if a value is changed. I think we should update the error report properties when the values change. If that is too much effort for this PR, I suggest renaming the error report properties to "...AtStartup" until the update is implemented.

I would find this PR easier to review if the description were more concise. I learned nothing by reading the second sentence, the second paragraph, and the "Where to look" section.

johnml1135:
Thanks -- both addressed in aa06e6766:

- **Staleness**: implemented the live update rather than renaming. `LexOptionsDlg.m_btnOK_Click` and the new `AvaloniaOptionsDialogLauncher.ApplyUiModeLive`/`ApplyDisabledToolsLive` helpers now call `ErrorReporter.AddProperty` alongside the existing `PropertyTable` broadcast, so the crash-report value tracks every live mode/disabled-tools change instead of only the window-construction seed. Covered by new tests in `LexOptionsDlgTests` and `AvaloniaOptionsDialogLauncherTests`.
- **Description length**: rewrote the PR body to a plain Summary + Test plan, dropped the accordions and the "Where to look" walkthrough.

Src/LexText/LexTextControls/LexOptionsDlg.cs
```diff
@@ -154,6 +155,8 @@ private void m_btnOK_Click(object sender, EventArgs e)
 					m_propertyTable.SetProperty(UIModePropertyName, newUiMode, true);
 					m_propertyTable.SetPropertyPersistence(UIModePropertyName, false);
 				}
+				// Keep the crash-report value in step with a live toggle, not just the startup seed.
```
papeh: this comment is unnecessary

papeh (approved):

Src/LexText/LexTextControls/LexOptionsDlg.cs
johnml1135: Removed. I took out the matching restatement above the `FwXWindow.SeedUIModeProperties` calls too, and trimmed the crash-report clauses I had appended to the `ApplyUiModeLive`/`ApplyDisabledToolsLive` comments -- the `ErrorReporter.AddProperty` calls say it themselves.

While rebasing onto current `main` I also caught a real bug that the merge introduced: `main` had since wrapped this block in an `m_uiModeChooser != null` check, and the rebase floated the `AddProperty` call out of the `oldUiMode != newUiMode` guard, so every OK-click would have recorded a value. Restored inside the guard; `OkClick_LeavesLegacyWhenUserDoesNotChangeSelection` covers it.

Rebased and force-pushed as 2bb175972. `./build.ps1 -CommentHygiene` clean; `test.ps1` for `FwXWindowUIModeSeedingTests` (8 tests) and `LexOptionsDlgTests|AvaloniaOptionsDialogLauncherTests` (46 tests) all pass.

PR #1069, opened 2026-08-14
LT-22524: Add substring search to Find Lexical Entry

thejambi:
## Quick Summary

Find Lexical Entry now matches typed characters anywhere in a word for queries of 3+ characters, using liblcm's new SearchType.Substring; shorter queries keep the original full-text search. The mode decision lives in a unit-tested SubstringSearchPolicy, and results retain the existing exact -> starts-with -> anywhere ordering.

Depends on https://github.com/sillsdev/liblcm/pull/395

https://jira.sil.org/browse/LT-22524

**Note: Two search indexes over the lexicon.** Find Lexical Entry now keeps two `StringSearcher` engines alive per dialog session — a full-text engine and a substring engine — selected per query by length (`SubstringSearchPolicy`). Both are built lazily-but-eagerly warmed at dialog open, and `SearchEngine.Get` caches them on the property table, so it's one extra background index build per session, not per open. The cost is two entry arrays, two `ConsumerThreads`, and two `IVwNotifyChange` registrations, so every lexicon data change now notifies both engines. This is a deliberate trade: it avoids rebuilding the index on every switch between full-text and substring mode (which a single swapped-`SearchType` engine would force on each keystroke that crosses the length threshold), at the price of a second resident index. Merge Entry keeps the same two-index model, both engines filtering out the starting entry.

## CI-ready checklist

  ```powershell
  git fetch origin
  git diff --check --cached
  ```
- [x] If this is core-developer AI-assisted work, I followed `Docs/workflows/ai-pr-workflow.md` and ran `pr-preflight` or the equivalent branch-readiness review before requesting review.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1069)

### DIFF
--- a/Build/SilVersions.props
+++ b/Build/SilVersions.props
@@ -12,7 +12,7 @@
 		=============================================================
 	-->
 	<PropertyGroup Label="SIL Ecosystem Versions">
-		<SilLcmVersion>11.0.0-beta0178</SilLcmVersion>
+		<SilLcmVersion>11.0.0-beta0180</SilLcmVersion>
 		<SilLibPalasoVersion>18.0.0-beta0030</SilLibPalasoVersion>
 		<SilLibPalasoL10nsVersion>18.0.0-beta0012</SilLibPalasoL10nsVersion>
 		<SilChorusVersion>6.0.0-beta0065</SilChorusVersion>
--- a/Src/Common/Controls/XMLViews/MatchingObjectsBrowser.cs
+++ b/Src/Common/Controls/XMLViews/MatchingObjectsBrowser.cs
@@ -205,6 +205,21 @@ namespace SIL.FieldWorks.Common.Controls
 			ResumeLayout(false);
 		}
 
+		/// <summary>
+		/// Swap the active search engine (e.g. when toggling substring-match mode). The caller
+		/// owns the lifetime of both engines; this only re-hooks the SearchCompleted event.
+		/// </summary>
+		public void SetSearchEngine(SearchEngine searchEngine)
+		{
+			CheckDisposed();
+			if (ReferenceEquals(m_searchEngine, searchEngine))
+				return;
+			if (m_searchEngine != null)
+				m_searchEngine.SearchCompleted -= m_searchEngine_SearchCompleted;
+			m_searchEngine = searchEngine;
+			m_searchEngine.SearchCompleted += m_searchEngine_SearchCompleted;
+		}
+
 		private void m_searchEngine_SearchCompleted(object sender, SearchCompletedEventArgs e)
 		{
 			UpdateResults(e.Fields.FirstOrDefault(), e.Results);
--- a/Src/LexText/LexTextControls/EntryGoDlg.cs
+++ b/Src/LexText/LexTextControls/EntryGoDlg.cs
@@ -54,7 +54,8 @@ namespace SIL.FieldWorks.LexText.Controls
 		}
 
 		/// <summary>
-		/// The substring (match-anywhere) engine. Built lazily the first time the user ticks the box.
+		/// The substring (match-anywhere) engine. Built lazily the first time a query is long enough
+		/// to use substring matching (see <see cref="SubstringSearchPolicy"/>).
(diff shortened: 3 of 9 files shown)


jasonleenaylor (changes requested):
Two things I'd like changed.

1. `UseSubstring` counts UTF-16 units, not characters. `string.Length` after `FormC` is still code units, and the comment says a base character plus a combining diacritic counts as one. That only holds when the pair has a precomposed form. Calling your own `UseSubstring` on the built branch:

```
two user-perceived characters each          FormC.Length   UseSubstring
  e+U+0301 then o   (acute composes)             2            false
  b+U+0303 then o   (tilde, no precomposed)      3            TRUE
  n+U+0330 then o   (tilde below)                3            TRUE
  t+U+032C then o   (caron below)                3            TRUE
  U+10480 then o    (non-BMP, surrogate pair)    3            TRUE
```

Four of those five two-character queries turn substring matching on, and the only one that doesn't is the one whose diacritic happens to compose. So the threshold fires a character early in exactly the orthographies this feature is most useful for, which is also where your rationale about matching most entries in a large project bites hardest.

`UseSubstring_countsComposedCharacters_notUtf16Units` doesn't catch it because its fixture is `ComposedAcuteE` repeated, and acute is one of the compositions that exists. The test passes and pins a belief that doesn't generalize. `StringInfo.LengthInTextElements` gives 1 for every single grapheme above and 2 for every two-grapheme query, so it's a one-line change that makes the code do what the comment already promises. Whichever way you go, the test needs a non-composable case.

2. The substring engine never gets the warm-up the full-text engine gets. `InitializeMatchingObjects` ends with `SearchAsync(GetFields(string.Empty, ...))` under a "start building index" comment, and that reaches whichever engine the browser holds, which at that point is the full-text one because `SearchEngineFor(string.Empty)` returns it. So the substring engine's index starts at zero, and a cold engine delivers nothing until its index is complete: `PerformSearch` builds the index before it searches, and a search the next keystroke cancels returns null with no event, so results only ever arrive from a search that finished indexing. `GetSearchableObjects` is `ILexEntryRepository.AllInstances().ToArray()`, every entry in the project. So crossing into substring matching, which is where the user expects results to get better, is where the first results have to wait on a full index build.

Simplest fix I can see: alongside the existing warm-up, call `SubstringSearchEngine.SearchAsync(GetFields(string.Empty, selectedWs.Handle))`. `SearchAsync` is public, and nothing is subscribed to that engine's `SearchCompleted` until `SetSearchEngine` hooks it, so the completion is a no-op. It has to come after `Initialize` because `GetFields` reads `IsVisibleColumn`, which is where the existing call already sits. Since `SearchEngine.Get` caches in the property table, that's one extra background build per session rather than per dialog open. A more surgical option is to warm it when the key reaches `MinQueryLength - 1`, so the build overlaps the keystroke before it's needed.

Either way, please say in the body that this keeps two indexes over the lexicon. Two `StringSearcher` instances, two entry arrays, two `ConsumerThread`s and two `IVwNotifyChange` registrations, so every data change now notifies both. I think that's the right trade against rebuilding the index on every mode flip, but it should be a decision on the record rather than a side effect.

Worth keeping as it is: `SearchEngineFor` gets the key after `FindMorphType` has stripped reserved characters, `SetSearchEngine` unhooks the old `SearchCompleted` before rehooking, and moving `CurrentEntryHvo` onto a property refreshes it per access instead of once at init.

Built and ran the branch at ff11df9f0: `build.ps1` clean, and all ten of your new tests pass. The table above is from invoking the compiled `SubstringSearchPolicy.UseSubstring`, not a standalone normalization check.

thejambi:
Changes made!

thejambi (commented):
> @thejambi reviewed 9 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1069)* status: 0 of 9 files reviewed, all discussions resolved.

jasonleenaylor (approved):
LGTM

PR #1079, opened 2026-08-18
LT-22691: Fix command routing through the hidden DataTree adapter

mark-sil:
The Avalonia detail host drives the legacy DataTree as a hidden command-routing adapter, but Mediator message-target selection dropped invisible trees and slices from the colleague chain. Every handler defined on the tree itself (JumpToTool, Delete, Insert, Move Field) and every slice-level handler (the Writing Systems options list) was unreachable, so those menu items were inert or wrongly disabled.

Keep an adapter-flagged tree and its slices in the colleague chain via IsExternalCommandAdapter, and clear the flag when the legacy view reactivates so a runtime UI flip does not widen legacy targeting.

Target the adapter's CurrentSlice by object and field name: sibling rows share one MoForm, and handlers such as GetGuidForJumpToTool resolve the target from CurrentSlice.Flid. Add
SetCurrentSliceForCommandTarget because ShowObject suspends ordinary CurrentSlice assignment until an idle callback, silently discarding the adapter's targeting, and add ClearCurrentSlice as the sanctioned no-target state. Exclude lazy placeholders via the new Slice.IsLazyPlaceholder instead of IsRealSlice, which for view-backed slices reports layout state and is false forever in a tree that is never shown.

Note: Other changes are need to complete LT-22691.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1079)

### DIFF
--- a/Src/Common/Controls/DetailControls/DataTree.cs
+++ b/Src/Common/Controls/DetailControls/DataTree.cs
@@ -769,7 +769,7 @@ namespace SIL.FieldWorks.Common.Framework.DetailControls
 				CheckDisposed();
 
 				if (value == null)
-					throw new ArgumentException("CurrentSlice on DataTree cannot be set to null. Set the underlying data member to null, if you really want it to be null.");
+					throw new ArgumentException("CurrentSlice on DataTree cannot be set to null. Call ClearCurrentSlice(), if you really want it to be null.");
 				Debug.Assert(!value.IsDisposed, "Setting CurrentSlice to a disposed slice -- not a good idea!");
 
 				// don't set the current slice until we're all setup to do so (LT-7307)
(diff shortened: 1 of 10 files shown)


jasonleenaylor (commented):
I can see how this is tackling the issue but I have a question about the direction of this commit. I was expecting that we would be implementing more Pub/Sub messaging into the Avalonia DataTree and slices to trigger the updates to the winforms controls which currently track the style, and the writing system that is under the cursor. This covers all the message handling which the DataTree handled by turning the old WinForms data tree into an invisible adapter layer. That's cute and effective, but doesn't move us toward a WinForms free future as much as I'd like. I'm willing to keep going this direction for now if the alternative was considered and ruled out for some reason.

> @jasonleenaylor reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1079#-:-P-Le2tqAHHUvAIK1UIM:bnok94c)* status: 2 of 10 files reviewed, all discussions resolved.

mark-sil (commented):
Thanks for your comment, I mis-understood the scope of what we were doing.  I was thinking the invisible adapter layer was an intentional design decision since it was already functioning (just not functioning well). I thought the Jira ticket was about fixing the problems with the current design.
I think it makes sense to Publish from the Avalonia view and Subscribe in the WinForms (or other) code.  The part that is not clear is the adapters interaction with the Mediator to support command execution. Would we still want to keep that portion of the current design? PubSub currently doesn't do anything with command execution.  

> @mark-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1079#-:-P-M-e_D2rdkhh2JvsSt:b-tcfrv7)* status: 2 of 10 files reviewed, all discussions resolved (waiting on jasonleenaylor).

jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 8 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1079#-:-P-PWd0H3pnJou87W-1D:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mark-sil).

PR #1080, opened 2026-08-18
Fix LT-22649

AndyBlack:
Both PcPatr with FLEx and TonePars with FLEx added new unowned lists (as well as some custom fields).  They never broadcast this fact, though.  This code fixes that (and does some cleanup, too).
It also fixes the crashes reported in https://jira.sil.org/browse/LT-22649.

It is basically just adding a PropertyTable instance so we can publish the change via
`FwUtils.Publisher.Publish(new PublisherParameterObject(EventConstants.MasterRefresh, null, PropTable.GetWindow()));`

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1080)

### DIFF
--- a/Src/Utilities/pcpatrflex/PcPatrFLExDll/PcPatrFLExForm.cs
+++ b/Src/Utilities/pcpatrflex/PcPatrFLExDll/PcPatrFLExForm.cs
@@ -1,25 +1,23 @@
+// Copyright (c) 2018-2026 SIL International
+// This software is licensed under the LGPL, version 2.1 or later
+// (http://www.gnu.org/licenses/lgpl-2.1.html)
+
 using Microsoft.Win32;
 using SIL.DisambiguateInFLExDB;
 using SIL.FieldWorks.Common.FwUtils;
 using SIL.LCModel;
-using SIL.LCModel.Core.Text;
-using SIL.LCModel.Core.WritingSystems;
-using SIL.LCModel.DomainServices;
 using SIL.PcPatrBrowser;
 using SIL.PrepFLExDB;
-using SIL.WritingSystems;
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Data;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Text;
-using System.Threading.Tasks;
 using System.Windows.Forms;
+using XCore;
 
 namespace SIL.PcPatrFLEx
 {
@@ -69,6 +67,7 @@ namespace SIL.PcPatrFLEx
 		public int MaxAmbiguities { get; set; }
 		public int TimeLimit { get; set; }
 		public bool RunIndividually { get; set; }
+		public PropertyTable PropTable { get; set; }
 
 		private RegistryKey regkey;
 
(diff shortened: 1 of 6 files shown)


papeh:
There may be some overlap between this and https://github.com/sillsdev/FieldWorks/pull/1075

Src/Utilities/pcpatrflex/PcPatrFLExDll/PcPatrFLExForm.cs
```diff
@@ -220,8 +219,10 @@ public void SaveRegistryInfo()
 		private void EnsureDatabaseHasBeenPrepped()
 		{
 			var preparer = new Preparer(Cache, false);
+			preparer.PropTable = PropTable;
 			preparer.AddPCPATRList();
 			preparer.AddPCPATRSenseCustomField();
+			FwUtils.Publisher.Publish(new PublisherParameterObject(EventConstants.MasterRefresh, null, PropTable.GetWindow()));
```
papeh, line 225: I'm curious whether this will still be necessary after https://github.com/sillsdev/FieldWorks/pull/1075 is merged, but this should be fine as is.

Src/Utilities/pcpatrflex/PrepFLExDBDll/PrepFLExDBTests/PreparerTests.cs
```diff
@@ -49,6 +47,9 @@ public override void FixtureSetup()
 			var settings = new LcmSettings { DisableDataMigration = true };
 			var progress = new DisambiguateInFLExDBTests.NullThreadedProgress(synchronizeInvoke);
 			MyCache = LcmCache.CreateCacheFromExistingData(ProjId, "en", logger, dirs, settings, progress);
+			Preparer = new Preparer(MyCache, false);
+			Preparer.PropTable = new XCore.PropertyTable(new XCore.Mediator());
```
papeh: As IDisposables, PropTable and Mediator should be disposed during FixtureTeardown (our builds try to alert us of missing dispose calls in tests)

Src/Utilities/pcpatrflex/PrepFLExDBDll/PrepFLExDBTests/PreparerTests.cs
papeh: Rather than calling FixtureSetup() at the beginning of each test, the fixture could have `[SetUp]` and `[TearDown]` methods that are called automatically.

Src/Utilities/pcpatrflex/PrepFLExDBDll/PrepFLExDBTests/PreparerTests.cs
```diff
@@ -135,31 +135,31 @@ public void ToneParsPreparerTest()
 			Assert.IsNotNull(MyCache);
 			Assert.AreEqual(5, MyCache.LangProject.AllPartsOfSpeech.Count);
 			Assert.AreEqual(0, MyCache.LangProject.LexDbOA.Entries.Count());
-			var preparer = new Preparer(MyCache, false);
+			Preparer.PropTable = new XCore.PropertyTable(new XCore.Mediator());
```
papeh: Why do we need to make a new property table if [SetUp] just made one?

papeh (changes requested):
Looks good overall, but tests should dispose their disposables.

AndyBlack:
> There may be some overlap between this and #1075

Ah, I was not aware of Jason doing this.  I'll happily wait until that PR is merged.  When it is, I'll see if it solves the crashes for https://jira.sil.org/browse/LT-22649 and https://jira.sil.org/browse/LT-22708. 
Also, thanks for noticing the other issues, Hasso.  I'll address them as needed later.

Src/Utilities/pcpatrflex/PcPatrFLExDll/PcPatrFLExForm.cs
AndyBlack: Jason's fix did avoid the crash of LT-22649, but it did not show the new list in the list area and also did not put the utlitity dialog on top.

papeh (approved):

PR #1084, opened 2026-08-19
Fix LT-22015: Add 'parse only words without an approved analysis' option

jtmaxwell3:
This fixes https://jira.sil.org/browse/LT-22015 again.  I misunderstood "words without an approved analysis" to be word instances without an approved analysis rather than wordforms without an approved analysis.  To fix this, I extracted code from OnClearSelectedWordParserAnalyses into GetLowercaseWordform and used it in GetUnapprovedWordforms to see whether a word instance or its lowercase version had an approved analysis.  I only consider word instances that are IWfiWordform to skip punctuation and word instances that are already approved.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1084)

### DIFF
--- a/Src/LexText/ParserUI/ParserListener.cs
+++ b/Src/LexText/ParserUI/ParserListener.cs
@@ -439,20 +439,30 @@ namespace SIL.FieldWorks.LexText.Controls
 			{
 				ClearWordformAnalyses(wf);
 				// Clear lower-case version of wf.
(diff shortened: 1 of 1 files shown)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1084)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #1085, opened 2026-08-19
Fix UpdateStatusPanelProgress

jtmaxwell3:
I noticed that the status progress panel in the lower left corner would sometimes stop updating when I was parsing a text.  This was intermittent, so I couldn't construct an acceptance test for it.  Looking at the code, it seemed like one possible explanation was that the ParserActivityString could equal ksIdle_ if the timer happened to sample the parsing activity between words (e.g. after parsing one word but before parsing the next).  I fixed this by adding QueuesAreEmpty in the conditional.  This prevents it from stopping if there are still words to parse.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1085)

### DIFF
--- a/Src/LexText/ParserUI/ParserListener.cs
+++ b/Src/LexText/ParserUI/ParserListener.cs
@@ -231,8 +231,19 @@ namespace SIL.FieldWorks.LexText.Controls
 													app.ActiveMainWindow, false);
 				}
 			}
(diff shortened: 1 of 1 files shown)


mark-sil (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1085)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jtmaxwell3).

PR #1089, opened 2026-08-19
LT-22651: Fix Dictionary letter headings only Uppercase if sorting is Custom Simple Shoebox

thejambi:
## Quick Summary

Letter headings under "Custom Sorting (Shoebox style)" showed only the uppercase
letter — `A` instead of `A a`. Digraphs were unaffected (`Ch ch`).

https://jira.sil.org/browse/LT-22651

## Fix

`GenerateLetterHeaderIfNeeded` built the lowercase half of the pair by
*normalizing* the lead character rather than lowercasing it, assuming
`GetLeadChar` had already returned a lowercase form. It doesn't: it returns the
primary representative of the writing system's sort rules, in whatever case
those rules were written.

Shoebox rules name the uppercase form first (`A a`), so the equivalence map
folds headwords onto the uppercase letter. Both halves of the heading then
matched, the pair collapsed, and only `A` rendered.

Digraphs escaped because their lookup lowercases the graph and runs *before*
the equivalence replacement — which is why only non-digraph headings
were affected.

Deriving the lowercase form through `CaseFunctions` is correct whichever case
the collation happens to use. The XHTML and Word generators share this method,
so both are covered.

## Tests

Two cases in `ConfiguredXHTMLGeneratorTests`, both using the rule set and
headwords from the project attached to the issue. Each asserts three headings:
a plain letter (`A a`), a letter sharing its first character with a digraph
(`C c`), and the digraph itself (`Ch ch`).

- **`..._ShoeboxSortRules_...`** — `SimpleRulesCollationDefinition`. Reproduces
  the reported symptom directly.
- **`..._UppercaseFirstIcuRules_...`** — `IcuRulesCollationDefinition`. Not
  redundant: that project's compiled collation (`< A << a < ... < Ch << ch`)
  names the uppercase letter as primary too, so the digraph parsing folds
  headwords onto it exactly as the simple rules do. The ICU path fails
  identically without the fix.

Both were verified failing before the fix and passing after.

## CI-ready checklist

  ```powershell
  git fetch origin
  git diff --check --cached
  ```
- [x] Builds/tests pass locally (or I've run the CI-style build via `build.ps1`/`test.ps1` or MSBuild).
- [x] If this is core-developer AI-assisted work, I followed `Docs/workflows/ai-pr-workflow.md` and ran `pr-preflight` or the equivalent branch-readiness review before requesting review.
- [ ] For any `Src/**` folders touched, corresponding `AGENTS.md` files are updated or explicitly confirmed still accurate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1089)

### DIFF
--- a/Src/xWorks/ConfiguredLcmGenerator.cs
+++ b/Src/xWorks/ConfiguredLcmGenerator.cs
@@ -133,10 +133,13 @@ namespace SIL.FieldWorks.XWorks
 			if (firstLetter != lastHeader && !string.IsNullOrEmpty(firstLetter))
 			{
 				var headerTextBuilder = new StringBuilder();
-				var upperCase =
-					new CaseFunctions(cache.ServiceLocator.WritingSystemManager.Get(wsString))
-						.ToTitle(firstLetter);
-				var lowerCase = firstLetter.Normalize();
+				var caseFunctions =
+					new CaseFunctions(cache.ServiceLocator.WritingSystemManager.Get(wsString));
+				var upperCase = caseFunctions.ToTitle(firstLetter);
+				// GetLeadChar returns whatever case the sort rules use for the
+				// primary character, and Shoebox-style simple rules are
+				// conventionally uppercase-first. LT-22651
+				var lowerCase = caseFunctions.ToLower(firstLetter).Normalize();
 				headerTextBuilder.Append(upperCase);
 				if (lowerCase != upperCase)
 				{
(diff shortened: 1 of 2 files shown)


mark-sil (approved):
> @mark-sil reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1089)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on thejambi).

PR #1093, opened 2026-08-20
Anchor dev code and data directories on the running source tree

johnml1135:
A Debug build now reads the `DistFiles` of the source tree it was built in. Before this, it read whichever tree the shared registry value `HKCU\SOFTWARE\SIL\FieldWorks\9\RootCodeDir` happened to name — on the machine that prompted this, a build in the main repo was loading parts, layouts, and configuration from an unrelated worktree under `.tmp/worktrees/`.

The reviewer's question here is "what did this break for installed FieldWorks?", and the answer is nothing: the new probe only matches a directory that has both `DistFiles` and `FieldWorks.sln` beside it, which no install has. The question worth your time is whether making the source tree outrank the registry is the right default.

**Where to look**

- `CodeDirectory` / `DataDirectory` now return before consulting the registry when the running assembly is inside a source tree — the deliberate behavior change. There is no opt-out: `FW_USE_REGISTRY_DIRS` was dropped in review as undiscoverable and unused.
- `FindDevDistFiles` walks up instead of assuming `<assembly>/../../DistFiles`, so it also works from `Output/Debug/x64` and from a project's own `bin` folder. Three layouts pinned by tests.
- The installed case is pinned by its own test: `DistFiles` present, no solution file, result `null`.
- `TidyRootDir` is extracted from `GetDirectory` so both paths normalize trailing separators identically — the ~100 `Path.Combine` callers see no difference.
- `ProjectsDirectory`'s **code** is untouched, but its behavior is not, and the earlier claim that it was untouched was wrong. `FwDirectoryFinder.cs:541` is `GetDirectory(ksProjectsDir, Path.Combine(DataDirectory, ksProjects))`, so its default derives from `DataDirectory`, which this PR changes. Where the HKCU/HKLM `ProjectsDir` value is **set**, nothing changes — which is most established dev machines. Where it is **absent** — a fresh dev box, a CI agent, a container — the projects directory moves from `%ProgramData%\SIL\FieldWorks\Projects` to `<tree>\DistFiles\Projects`.

**Deliberately not here**

- `Src/FwParatextLexiconPlugin/ParatextLexiconPluginDirectoryFinder.cs` keeps its registry-only resolution; it runs inside Paratext, never from a source tree.
- No log line or warning when an existing `RootCodeDir` override stops applying.
- `Build/mkall.targets` still carries the `setKeysInHKCU` target that writes `RootCodeDir`/`RootDataDir`, though `mkall.targets:277` says those targets appear unused by the modernized build; either way a source-tree build now ignores those values.

**Verification**

`.\build.ps1 -CommentHygiene -BuildTests` succeeded (0 warnings, 0 errors; comment-hygiene clean). `.\test.ps1 -CommentHygiene -SkipNative -TestProject Src\Common\FwUtils\FwUtilsTests\FwUtilsTests.csproj` — 407/407 passed, including 6 new cases. Not run: the full suite, native tests, installer validation. Not done: a manual two-worktree launch.

---

<details>

This started as a question, not a bug report: "my local Debug build still looks at DistFiles — could it look at `Output` instead?" The investigation said no to the literal request and yes to the problem behind it. Both halves are recorded below, because the rejected half is the one that will otherwise be re-proposed.

There were no working documents to delete; the reasoning never existed anywhere but here.

</details>

<details>

**The registry loses to the source tree, rather than the probe merely being fixed.** Fixing the anchoring alone would have changed nothing on a real dev machine: `GetDevDistFilesPath()` only ever fed `defaultDir`, and `GetDirectory` returns the registry value whenever it is non-empty. On a dev machine it is always non-empty. `Build/mkall.targets` (`setKeysInHKCU`) has historically written it, but `mkall.targets:277` carries a `REVIEW (Hasso) 2026.03` noting those targets "appear unused by the recently-modernized build process", so the live writers are more likely the MSI (`FLExInstaller/Overrides.wxi:11-12`) and `Resolve-FieldWorksDevRegistry.ps1:54-55`, plus the winapp skill's launch script. The value is a single machine-wide slot shared by every worktree, so it names whichever tree ran last. That is the actual defect; the fixed-depth probe is a second, independent one.

**`FieldWorks.sln` as the tree marker.** The probe needs something that exists in a source tree and never beside an install. The installer harvests `DistFiles\**\*` — the *contents*, into the install root — so an install has neither a `DistFiles` folder nor a solution file at that level. Requiring both makes the installed path unreachable by construction rather than by convention.

**No memoization.** Each `CodeDirectory` get now walks up doing `Directory.Exists` + `File.Exists` per level. The old path opened and read a registry key on every get, so this is not a regression. Revisit only with a measurement.

</details>

<details>

**Pointing the code directory at `Output/<Configuration>` — the literal request.** `Output` holds build artifacts only; the code/data payload exists solely in `DistFiles`. Counted in the tree at the time: `Language Explorer` 10 entries in `DistFiles` vs absent from `Output/Debug`; `Parts` 4 vs absent; `Helps` 7 vs absent; `Icu70` present vs absent; `Templates` 42 vs 1. `FlexStylesPath`, `FlexFolder`, `TemplateDirectory`, and `EditorialChecksDirectory` would all have broken.

**An overlay that probes `Output` first, then falls back to `DistFiles`.** This cannot be expressed through the current API: `CodeDirectory` returns one string that ~100 call sites `Path.Combine` onto. An overlay needs a `ResolveCodeFile(relativePath)` seam instead — a much larger change, for a duplication problem that does not currently exist (only `Templates` overlaps at all, with one entry).

**Just fixing the registry and stopping there.** That is what unblocked the reporter (`Resolve-FieldWorksDevRegistry.ps1 -Force`, run before this branch existed), and it is what every worktree switch will need again tomorrow. It treats the symptom.

</details>

<details>

**Precedence, before this change** — `GetDirectory(RegistryKey, string, string)` in `Src/Common/FwUtils/FwDirectoryFinder.cs`: `rootDir` is read from the registry, and `defaultDir` is used only `if (string.IsNullOrEmpty(rootDir))`. `GetDevDistFilesPath()` fed `defaultDir`. Hence: registry set → probe irrelevant.

**The shared slot** — `HKCU\SOFTWARE\SIL\FieldWorks\$(FWMAJOR)` holds `RootCodeDir`, `RootDataDir` and `ProjectsDir`. `Build/mkall.targets` target `setKeysInHKCU` writes them from `$(dir-fwdistfiles)`, though `mkall.targets:277` says those targets appear unused now; the MSI (`FLExInstaller/Overrides.wxi:11-12`) and `Resolve-FieldWorksDevRegistry.ps1:54-55` are the likelier live writers. Nothing in `Src/` writes those two values at runtime (searching for `SetValue("RootCodeDir"` outside tests returns no hits), so the value persists from whichever tree last built or launched.

**Existing tests keep passing for a non-trivial reason** — `FwDirectoryFinderTests` sets the registry to `UtilsAssemblyDir/../../DistFiles`, and `InitializeFwRegistryHelperAttribute` does the same. Under the new precedence those values are ignored, but the walk-up returns the same path for a test run out of `Output/Debug`, so the assertions still hold.

**New coverage** — `FindDevDistFiles_InsideSourceTree_FindsTreeDistFiles` (`Output/Debug`, `Output/Debug/x64`, `Src/Common/FwUtils/bin/Debug/net8.0`), `FindDevDistFiles_OutsideSourceTree_ReturnsNull`, and `CodeAndDataDirectory_PreferSourceTreeOverRegistry` (`RootCodeDir`, `RootDataDir`), which points the registry at a fabricated other worktree and asserts both directories still resolve to this tree.

</details>

<details>

# Code Review Summary

**Branch**: fix/dev-dir-anchoring

**Base**: origin/main

**Date**: 2026-08-28

**Files changed**: 6

## Overview

A dev build resolves its code and data directories from a machine-wide registry
slot, so a build run from one worktree can silently read another worktree's
DistFiles. The branch walks up from the running assembly to the enclosing source
tree instead, identified by `FieldWorks.sln`, and falls back to the registry only
outside a tree.

This pass responded to a ten-item CHANGES_REQUESTED review. Every item was
verified against the code before being acted on; one was confirmed by executing
the suspect function, and one review claim did not hold. A later self-audit,
prompted by the author, found that this pass's own `SourceDirectory` change had
shipped without coverage of what it changed. That is fixed in `c32048433`.

## Contract/API Changes

- `FwDirectoryFinder.FindDevDistFiles(string)` — public, added by the first
  commit on this branch.
- `FwDirectoryFinder.AssemblyDirectoryFromCodeBase(string)` — new `internal`,
  visible to `FwUtilsTests` via existing `InternalsVisibleTo`.
- `FwDirectoryFinder.FindSourceDirectory(string)` — new `internal`, same
  visibility.
- `FW_USE_REGISTRY_DIRS` — removed. It was introduced on this branch and never
  shipped, so no external contract is broken.
- `FwDirectoryFinder.SourceDirectory` — unchanged signature; resolution now
  follows the source-tree walk rather than two fixed parent hops.
- `ProjectsDirectory` — unchanged code, changed behaviour where the registry
  value is absent. Recorded in the PR body.

- [x] **The walk-up was a no-op on any path containing a space** _(fixed during
  review: derived through `AssemblyDirectoryFromCodeBase`, which unescapes)_.
  Confirmed by calling the function:
  `StripFilePrefix('file:///C:/My%20Repos/...')` returns the path with `%20`
  intact, so every `Directory.Exists` missed and resolution fell back to the
  registry with no diagnostic — the bug this branch exists to remove.

- [x] **`SourceDirectory` still used two fixed parent hops and threw** _(fixed
  during review: routed through the same walk)_. A build from
  `Output/Debug/x64` resolved code and data but failed here, contradicting the
  layout `FindDevDistFiles_InsideSourceTree_FindsTreeDistFiles` claims.
- [x] **That reroute shipped with no test of its own** _(fixed during review in
  `c32048433`: `FindSourceDirectory` seam plus three layout cases and a throws
  case)_. Found by self-audit, not by the reviewer. A pre-existing test caught an
  outright break but nothing exercised the layouts, which was the point of the
  change.
- [x] **The registry test re-encoded the fixed depth it was meant to retire, and
  its cleanup could mask a real failure** _(fixed during review: expectation
  derived the way production derives it; an absent original value is deleted
  rather than passed to `SetValue`, which throws on null)_.
- [x] **Nothing failed if the source-tree marker disappeared** _(fixed during
  review: added a test naming `ksSolutionFilename`)_. Sharper than reported:
  `CanonicalJsonTests`, `EngineIsolationAuditTests` and
  `LayoutImportCoverageTests` already walk up to `FieldWorks.sln`, so a rename
  breaks four things at once.
- [x] **The PR body claimed `ProjectsDirectory` was untouched** _(fixed during
  review: body now states that where the registry value is absent the projects
  directory moves to `<tree>\DistFiles\Projects`)_.
- [x] **`FW_USE_REGISTRY_DIRS` was undiscoverable and unused** _(fixed during
  review: removed)_.
- [ ] ~~Tests constructing a WinForms control break a standing rule~~ _(the
  facts hold but the framing does not: no such rule is written down, and
  `DataTreeTests`, `SliceTests` and `ReversalEntryViewTests` all construct
  slices without STA. Not a reason to act; the `UserControl` argument stands on
  its own and was acted on elsewhere.)_

### Minor - Consider

- [x] **Three probes for one question** _(decided per site)_. `SourceDirectory`
  routed through the walk; ICU left alone with a reason and LT-22768 (it runs on
  the `CustomIcu.InitIcuDataDir` bootstrap path, so the dependency is an
  initialisation-order question); `StringTable` left alone with a reason and
  LT-22769 (walks to the last "output" segment, feeding a different fallback
  chain).
- [x] **`TidyRootDir`'s doc comment described what it does to its callers**
  _(fixed during review: states its own contract)_.
- [x] **Four comment items narrating removed behaviour or restating names**
  _(fixed during review)_.
- [x] **Temporal framing in the winapp skill doc, and the body naming the wrong
  registry writer** _(fixed during review: `mkall.targets:277` carries a
  `REVIEW (Hasso) 2026.03` saying those targets appear unused, so the body now
  points at the MSI and `Resolve-FieldWorksDevRegistry.ps1`)_.

Run:

- `./build.ps1 -CommentHygiene -BuildTests -SkipNative` — clean, comment-hygiene
  clean.
- `./test.ps1 -CommentHygiene -NoBuild -SkipNative -TestProject FwUtilsTests` —
  413/413 (was 409 before the added layout cases).
- Consumer projects of `SourceDirectory`, which the first pass skipped:
  `XMLViewsTests` 114/114, `DetailControlsTests` 119/120, `xCoreTests` 17/17,
  `LexTextControlsTests` 353/356. The shortfalls are pre-existing ignores at the
  same counts as an unrelated branch.

Ablation sweep, one production change at a time:

| Ablation | Caught by |
| --- | --- |
| Remove the unescaping | `FindDevDistFiles_UnderAPathContainingASpace` |
| Restore the registry's precedence | `CodeAndDataDirectory_PreferSourceTreeOverRegistry("RootDataDir")` |
| `SourceDirectory` returns the wrong directory | 4 tests, including all three layout cases |

Known limits, stated rather than implied:

- `CodeDirectory`/`DataDirectory` read the running assembly, so their unescaping
  is covered through the extracted seam, not end to end. Relocating a running
  assembly is out of reach for a unit test.
- The marker test guards removal of `FieldWorks.sln`; it does not test
  production logic.
- Five of the ten-plus projects consuming `SourceDirectory` were run locally. CI
  covers the rest.
- Not run locally: native tests (no native change), installer validation (no
  installer change).

## Positive Observations

- The original "Paths not taken" section rejected two designs with file counts
  and a call-site count rather than taste.
- Worktree behaviour is correct: the walk stops at the innermost match, so an
  assembly under a nested worktree resolves that worktree's DistFiles.
- Installed-build safety holds. Independently confirmed: nothing in installer or
  patch staging runs the packaged managed binaries from inside the checkout —
  the only executed step, `Remove-StaleDlls.ps1`, reads assembly metadata via
  `AssemblyName.GetAssemblyName()` and never loads them.

## Interview Notes

The author made each scoping decision in session: fix `SourceDirectory` and file
tickets for the other two probe sites; drop `FW_USE_REGISTRY_DIRS`. No author
interview was run in this pass, because the decisions were already recorded and
the one open finding was the reviewer's own verification gap rather than the
author's.

Author challenge worth recording: after being shown a passing test suite, the
author asked whether the testing was actually sound. It was not — see the third
Important finding. The audit that followed produced the ablation sweep above.

Two errors of mine during this pass, both caught before merge:

1. The first space-path test called the walk with an already-clean path, so it
   passed with the fix reverted. Rewritten to drive the derivation.
2. A commit message body line was 81 characters and would have failed
   `Check commit messages`. Amended; that is the one force-push on this branch
   (`2b1098f37` -> `94145a678`, identical tree).

## In-Review Quality Check

`c32048433` adds the `FindSourceDirectory` seam and four tests. Build,
comment-hygiene, `FwUtilsTests` and the four consumer projects were re-run after
it; results above.

## Suggested Review Focus

- [ ] Whether `SourceDirectory` changing resolution strategy is acceptable given
      its ~30 call sites across ten-plus test projects, five of which were
      exercised locally.
- [ ] Whether leaving the ICU probe alone is the right call, or whether
      LT-22768 should block this branch rather than follow it.
- [ ] `ProjectsDirectory`'s behaviour change on machines with no registry value
      — fresh dev boxes, CI agents, containers.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1093)

### DIFF
--- a/.claude/skills/fieldworks-winapp/SKILL.md
+++ b/.claude/skills/fieldworks-winapp/SKILL.md
@@ -120,6 +120,7 @@ verifies that launch will succeed.
 - `scripts/Set-FieldWorksLegacyMode.ps1`: forces `UIMode=Legacy` — run before EVERY launch.
 - `scripts/Resolve-FieldWorksDevRegistry.ps1`: aligns the dev registry (`RootCodeDir`/`RootDataDir`) to
   this worktree before launch; auto-realigns when the other worktree is idle, else prints `RESULT=ASK_USER`.
+  Current builds anchor on their own source tree, so this now matters mainly for older builds.
 - `references/headless-rendering.md`: why FieldWorks needs a display-bound desktop; what works/doesn't for
   invisible capture (winforms-mcp HEADLESS does NOT render FieldWorks; a Virtual Display Driver was tried
   and abandoned — see the doc for why; use visible capture, or RDP for true invisibility). Starts with the
--- a/.claude/skills/fieldworks-winapp/SKILL.md
+++ b/.claude/skills/fieldworks-winapp/SKILL.md
@@ -120,7 +120,8 @@ verifies that launch will succeed.
 - `scripts/Set-FieldWorksLegacyMode.ps1`: forces `UIMode=Legacy` — run before EVERY launch.
 - `scripts/Resolve-FieldWorksDevRegistry.ps1`: aligns the dev registry (`RootCodeDir`/`RootDataDir`) to
   this worktree before launch; auto-realigns when the other worktree is idle, else prints `RESULT=ASK_USER`.
-  Current builds anchor on their own source tree, so this now matters mainly for older builds.
+  A build running from inside a source tree resolves code and data from that tree and
+  ignores these values; they govern a build launched from outside one.
 - `references/headless-rendering.md`: why FieldWorks needs a display-bound desktop; what works/doesn't for
   invisible capture (winforms-mcp HEADLESS does NOT render FieldWorks; a Virtual Display Driver was tried
   and abandoned — see the doc for why; use visible capture, or RDP for true invisibility). Starts with the
--- a/.claude/skills/fieldworks-winapp/navigation/launch-or-attach.md
+++ b/.claude/skills/fieldworks-winapp/navigation/launch-or-attach.md
@@ -28,6 +28,9 @@ returns empty and `winforms_take_screenshot` is blank even though the process is
   when the other worktree is NOT active (no FieldWorks.exe running from it) and was NOT used in the last 24h
   (registry key write time). If it prints `RESULT=ASK_USER`, the other worktree may be active — **ask the
   user** before realigning, then re-run with `-Force` if they approve.
+  Since `FwDirectoryFinder` learned to anchor on the source tree it runs from, an exe built from a current
+  worktree already reads its own `DistFiles`; keep running the script for older builds, and for the other
+  registry values (`ProjectsDir`), which are still shared across worktrees.
 
 See the script headers and `../references/mcp-setup.md`.
 
--- a/Src/Common/FwUtils/FwDirectoryFinder.cs
+++ b/Src/Common/FwUtils/FwDirectoryFinder.cs
@@ -315,9 +315,15 @@ namespace SIL.FieldWorks.Common.FwUtils
 					ResourceHelper.GetResourceString("kstidInvalidInstallation")
 				);
 			}
-			// Hundreds of callers of this method are using Path.Combine with the results.
-			// Combine only works with a root directory if it is followed by \ (e.g., c:\)
-			// so we don't want to trim the \ in this situation.
+			return TidyRootDir(rootDir);
+		}
+
+		/// <summary>
+		/// Strips the trailing separator that hundreds of callers would otherwise pass on to
+		/// Path.Combine, except on a root directory (e.g. c:\), where Combine needs it.
+		/// </summary>
+		private static string TidyRootDir(string rootDir)
+		{
 			string dir = rootDir.TrimEnd(
 				Path.DirectorySeparatorChar,
 				Path.AltDirectorySeparatorChar
@@ -325,15 +331,46 @@ namespace SIL.FieldWorks.Common.FwUtils
 			return dir.Length > 2 ? dir : dir + Path.DirectorySeparatorChar;
 		}
 
+		/// <summary>The file that marks the root of a FieldWorks source tree.</summary>
+		private const string ksSolutionFilename = "FieldWorks.sln";
+
+		/// <summary>Set this to let the registry name the directories again.</summary>
+		private const string ksUseRegistryDirsVariable = "FW_USE_REGISTRY_DIRS";
+
+		/// <summary>
+		/// Gets the DistFiles folder of the source tree that <paramref name="startDirectory"/>
+		/// lies in, or <c>null</c> if it lies outside a source tree (the installed case).
+		/// </summary>
+		/// <remarks>
+		/// Walking up to the tree root, rather than assuming a fixed depth, finds
+		/// DistFiles from Output/&lt;Configuration&gt;, from its architecture
+		/// subfolders, and from a project's own bin folder alike.
+		/// </remarks>
+		/// <param name="startDirectory">The directory to start searching upwards from.</param>
+		public static string FindDevDistFiles(string startDirectory)
+		{
+			for (
+				string dir = startDirectory;
+				!string.IsNullOrEmpty(dir);
+				dir = Path.GetDirectoryName(dir)
+			)
+			{
+				string distFiles = Path.Combine(dir, "DistFiles");
+				// The solution file is what keeps an installed FieldWorks from matching here.
+				if (Directory.Exists(distFiles)
+					&& File.Exists(Path.Combine(dir, ksSolutionFilename)))
+					return distFiles;
+			}
+			return null;
+		}
+
 		private static string GetDevDistFilesPath()
 		{
+			if (EnvironmentVariables.IsTrue(ksUseRegistryDirsVariable))
+				return null;
+
 			string assemblyDir = Path.GetDirectoryName(FileUtils.StripFilePrefix(Assembly.GetExecutingAssembly().CodeBase));
-			// Check if we are in Output/Debug or Output/Release
-			// DistFiles is at ../../DistFiles
-			string distFiles = Path.GetFullPath(Path.Combine(assemblyDir, "..", "..", "DistFiles"));
-			if (Directory.Exists(distFiles))
-				return distFiles;
-			return null;
+			return FindDevDistFiles(assemblyDir);
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -349,15 +386,16 @@ namespace SIL.FieldWorks.Common.FwUtils
 		{
 			get
 			{
+				// The tree the assembly runs from owns its DistFiles, so it beats the registry.
+				string devDistFiles = GetDevDistFilesPath();
+				if (devDistFiles != null)
+					return TidyRootDir(devDistFiles);
+
 				string defaultDir = Path.Combine(
 					Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles),
 					CompanyName,
 					$"FieldWorks {FwUtils.SuiteVersion}"
 				);
-
-				string devDistFiles = GetDevDistFilesPath();
-				if (devDistFiles != null)
-					defaultDir = devDistFiles;
 
 				return GetDirectory("RootCodeDir", defaultDir);
 			}
@@ -377,11 +415,12 @@ namespace SIL.FieldWorks.Common.FwUtils
 		{
 			get
 			{
-				string defaultDir = Path.Combine(LcmFileHelper.CommonApplicationData, CompanyName, ksFieldWorks);
-
+				// See CodeDirectory: the running tree outranks the shared registry.
 				string devDistFiles = GetDevDistFilesPath();
 				if (devDistFiles != null)
-					defaultDir = devDistFiles;
+					return TidyRootDir(devDistFiles);
+
+				string defaultDir = Path.Combine(LcmFileHelper.CommonApplicationData, CompanyName, ksFieldWorks);
 
 				return GetDirectory(
 					ksRootDataDir,
--- a/Src/Common/FwUtils/FwDirectoryFinder.cs
+++ b/Src/Common/FwUtils/FwDirectoryFinder.cs
@@ -463,6 +463,24 @@ namespace SIL.FieldWorks.Common.FwUtils
 			}
 		}
 
+		/// <summary>
+		/// The Src directory of the source tree <paramref name="startDirectory"/> lies in.
+		/// </summary>
+		/// <exception cref="ApplicationException">There is no source tree above
+		/// <paramref name="startDirectory"/>, or the tree has no Src directory.</exception>
+		internal static string FindSourceDirectory(string startDirectory)
+		{
+			string distFiles = FindDevDistFiles(startDirectory);
+			string dir = distFiles == null
+				? null
+				: Path.Combine(Path.GetDirectoryName(distFiles), "Src");
+			if (dir == null || !Directory.Exists(dir))
+				throw new ApplicationException(
+					"Could not find the Src directory.  Was expecting it at: "
+					+ (dir ?? "(no source tree above " + startDirectory + ")"));
+			return dir;
+		}
+
 		private static string m_srcdir;
 
 		/// ------------------------------------------------------------------------------------
@@ -481,16 +499,7 @@ namespace SIL.FieldWorks.Common.FwUtils
 
 				// The same walk the code and data directories use, so every layout it supports
 				// resolves here too.
-				string distFiles = FindDevDistFiles(ExeOrDllDirectory);
-				string dir = distFiles == null
-					? null
-					: Path.Combine(Path.GetDirectoryName(distFiles), "Src");
-				if (dir == null || !Directory.Exists(dir))
-					throw new ApplicationException(
-						"Could not find the Src directory.  Was expecting it at: "
-						+ (dir ?? "(no source tree above " + ExeOrDllDirectory + ")")
-					);
-				m_srcdir = dir;
+				m_srcdir = FindSourceDirectory(ExeOrDllDirectory);
 
 				return m_srcdir;
 			}
--- a/Src/Common/FwUtils/FwDirectoryFinder.cs
+++ b/Src/Common/FwUtils/FwDirectoryFinder.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2003-2018 SIL International
+﻿// Copyright (c) 2003-2018 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 //
@@ -319,8 +319,8 @@ namespace SIL.FieldWorks.Common.FwUtils
 		}
 
 		/// <summary>
-		/// Strips the trailing separator that hundreds of callers would otherwise pass on to
-		/// Path.Combine, except on a root directory (e.g. c:\), where Combine needs it.
+		/// The directory without a trailing separator, except at a drive root (e.g. c:\), where
+		/// Path.Combine requires one.
 		/// </summary>
 		private static string TidyRootDir(string rootDir)
 		{
@@ -334,19 +334,27 @@ namespace SIL.FieldWorks.Common.FwUtils
 		/// <summary>The file that marks the root of a FieldWorks source tree.</summary>
 		private const string ksSolutionFilename = "FieldWorks.sln";
 
-		/// <summary>Set this to let the registry name the directories again.</summary>
-		private const string ksUseRegistryDirsVariable = "FW_USE_REGISTRY_DIRS";
-
 		/// <summary>
 		/// Gets the DistFiles folder of the source tree that <paramref name="startDirectory"/>
 		/// lies in, or <c>null</c> if it lies outside a source tree (the installed case).
 		/// </summary>
 		/// <remarks>
-		/// Walking up to the tree root, rather than assuming a fixed depth, finds
-		/// DistFiles from Output/&lt;Configuration&gt;, from its architecture
-		/// subfolders, and from a project's own bin folder alike.
+		/// The solution file is what distinguishes a source tree from an install, so an
+		/// installed FieldWorks never matches here however deep its DistFiles sits.
 		/// </remarks>
-		/// <param name="startDirectory">The directory to start searching upwards from.</param>
+		/// <summary>
+		/// The directory holding the assembly at <paramref name="codeBase"/>, a file:// URI.
+		/// </summary>
+		/// <remarks>
+		/// Unescapes before returning: a repo path containing a space arrives here as %20, and
+		/// a start directory that still holds it matches no directory at all.
+		/// </remarks>
+		internal static string AssemblyDirectoryFromCodeBase(string codeBase)
+		{
+			var uri = new Uri(codeBase);
+			return Path.GetDirectoryName(Uri.UnescapeDataString(uri.AbsolutePath));
+		}
+
 		public static string FindDevDistFiles(string startDirectory)
 		{
 			for (
@@ -366,11 +374,8 @@ namespace SIL.FieldWorks.Common.FwUtils
 
 		private static string GetDevDistFilesPath()
 		{
-			if (EnvironmentVariables.IsTrue(ksUseRegistryDirsVariable))
-				return null;
-
-			string assemblyDir = Path.GetDirectoryName(FileUtils.StripFilePrefix(Assembly.GetExecutingAssembly().CodeBase));
-			return FindDevDistFiles(assemblyDir);
+			return FindDevDistFiles(AssemblyDirectoryFromCodeBase(
+				Assembly.GetExecutingAssembly().CodeBase));
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -462,8 +467,10 @@ namespace SIL.FieldWorks.Common.FwUtils
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Gets the src dir (for running tests)
+		/// The Src directory of the source tree this assembly is running from (for tests).
 		/// </summary>
+		/// <exception cref="ApplicationException">The assembly is not inside a source tree, or
+		/// the tree has no Src directory.</exception>
 		/// ------------------------------------------------------------------------------------
 		public static string SourceDirectory
 		{
@@ -472,15 +479,16 @@ namespace SIL.FieldWorks.Common.FwUtils
 				if (!string.IsNullOrEmpty(m_srcdir))
 					return m_srcdir;
 
-				// We'll assume the executing assembly is $FW/Output/Debug/FwUtils.dll,
-				// and the source dir is $FW/Src.
-				var dir = ExeOrDllDirectory;
-				dir = Path.GetDirectoryName(dir); // strip the parent directory name (Debug)
-				dir = Path.GetDirectoryName(dir); // strip the parent directory again (Output)
-				dir = Path.Combine(dir, "Src");
-				if (!Directory.Exists(dir))
+				// The same walk the code and data directories use, so every layout it supports
+				// resolves here too.
+				string distFiles = FindDevDistFiles(ExeOrDllDirectory);
+				string dir = distFiles == null
+					? null
+					: Path.Combine(Path.GetDirectoryName(distFiles), "Src");
+				if (dir == null || !Directory.Exists(dir))
 					throw new ApplicationException(
-						"Could not find the Src directory.  Was expecting it at: " + dir
+						"Could not find the Src directory.  Was expecting it at: "
+						+ (dir ?? "(no source tree above " + ExeOrDllDirectory + ")")
 					);
 				m_srcdir = dir;
 
--- a/Src/Common/FwUtils/FwUtils.cs
+++ b/Src/Common/FwUtils/FwUtils.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2009-2018 SIL International
+﻿// Copyright (c) 2009-2018 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
@@ -221,6 +221,8 @@ namespace SIL.FieldWorks.Common.FwUtils
 					return null;
 
 				var ver = CustomIcu.Version.ToString(CultureInfo.InvariantCulture);
+				// Not FwDirectoryFinder.FindDevDistFiles: on the InitIcuDataDir bootstrap path
+				// that dependency is an initialisation-order question. See LT-22768.
 				var distFiles = Path.GetFullPath(Path.Combine(assemblyDir, "..", "..", "DistFiles"));
 				if (!Directory.Exists(distFiles))
 					return null;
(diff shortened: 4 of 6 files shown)


jasonleenaylor (changes requested):
The diagnosis is right and the installed-build safety argument holds up — I checked it
rather than taking it on trust (see item 6). The "Paths not taken" section is the most
useful part of the body: rejecting the `Output/<Configuration>` idea with actual file
counts, and rejecting the overlay because `CodeDirectory` returns one string that ~100
sites `Path.Combine` onto, are both the right calls argued the right way. Worktree
behaviour is correct too — the walk stops at the innermost match, so an assembly under
`fw/.claude/worktrees/X/Output/Debug` resolves `X/DistFiles`, which is the point.

The through-line of what follows: this change is correct when it works and **invisible when
it does not**. Every failure mode below degrades to "silently reads the wrong tree" rather
than to an error, which is the same class of bug the PR was written to eliminate.

**1. The walk-up is a no-op on any path containing a space.**

`FwDirectoryFinder.cs:330` derives the start directory with
`FileUtils.StripFilePrefix(Assembly.GetExecutingAssembly().CodeBase)`. I loaded `SIL.Core.dll`
and called it:

    StripFilePrefix('file:///C:/My%20Repos/fw/Output/Debug/FwUtils.dll')
    => C:/My%20Repos/fw/Output/Debug/FwUtils.dll

It strips the scheme but does not unescape. Every `Directory.Exists` in the new walk then
misses, `FindDevDistFiles` returns `null`, and resolution falls back to the registry — the
exact stale-worktree bug this PR exists to fix, with no diagnostic.

This is the PR's to fix rather than inherited debt, because the same class already gets it
right fifteen lines away: `ExeOrDllDirectory` (`:417-418`) uses `Uri.UnescapeDataString`, as
does `InitializeFwRegistryHelperAttribute.cs:43-47`. The probe was promoted to authoritative
without being made correct.

Please have `GetDevDistFilesPath` call `ExeOrDllDirectory` instead of re-deriving the path a
second way, and add a test case whose fixture path contains a space — nothing currently
exercises that.

(Worth noting for item 3: `FwUtils.cs:219` uses `Assembly.GetExecutingAssembly().Location`,
which is already a plain filesystem path and has no escaping problem. `CodeBase` is
presumably why this class does what it does — shadow-copying under test runners — so
`ExeOrDllDirectory` is still the right fix here, but `Location` is worth knowing about.)

**2. Please drop `FW_USE_REGISTRY_DIRS`.**

To be clear about what I am not saying: it is genuinely distinct from
`FW_ROOT_CODE_DIR` / `FW_ROOT_DATA_DIR`, and I looked hard at whether it duplicated them. It
does not. Those pin a specific path; this defers to whatever the registry currently says, and
there is no value you can set the pair to that means "the registry". Having managed code
honour that pair would also be a CI change — `Build/Agent/Setup-FwBuildEnv.ps1:104-108`
already sets both on every agent, and managed `FwDirectoryFinder` ignores them today — which
does not belong in a dev-ergonomics PR. And honouring them would apply to installed builds
too, where a stale exported variable would silently win. So the mechanism you chose is the
narrower and safer one.

The problem is discoverability. Nobody who needs it will know it exists. It is documented
only in the source, it is not named in the skill docs this PR updates, and a developer
hitting the problem it solves has no path to finding it. Unused configuration still has to be
maintained, tested and reasoned about, and it is one more environment variable in a space
that already has two honoured by a different layer with different semantics. If the need
turns up in practice, add it then, with documentation.

**3. Decide, explicitly, whether `FindDevDistFiles` is the seam.**

The PR calls the fixed-depth probe "a second, independent defect". `FwUtils` contains three
implementations of it and this fixes one:

- `FwDirectoryFinder.cs:328` — fixed here.
- `FwUtils.cs:215-241` `TryGetDevIcuDataDir()` — still `Path.Combine(assemblyDir, "..", "..", "DistFiles")`.
- `StringTable.cs:71-113` — a third idiom, stripping at the last `output` segment.

Plus `FwDirectoryFinder.SourceDirectory` (`:429-450`), eleven lines below the new method,
still does two hard-coded `GetDirectoryName` hops and throws `ApplicationException` when they
miss.

The concrete result: after this PR a build run from `Output/Debug/x64` — a layout
`FindDevDistFiles_InsideSourceTree_FindsTreeDistFiles` explicitly claims support for —
resolves code and data correctly but still fails to find dev ICU data and still throws from
`SourceDirectory`. One file, three answers.

I am not asking you to fix all of them. I am asking for a decision per site, stated: route it
through `FindDevDistFiles`, or keep it separate and say why. You made the method `public`
for reuse, so either it is the seam or it is not.

`SourceDirectory` is the sharpest case, being in this file and contradicting the new test's
own claim. **ICU deserves extra thought rather than a default answer** — it starts from
`Location` rather than `CodeBase`, it is a best-effort `catch`-all returning `null`, and it
runs on the `CustomIcu.InitIcuDataDir()` bootstrap path, so taking a new dependency on
`FwDirectoryFinder` there is an initialisation-order question, not just a refactor. It may
well be right to leave it alone; I would rather that be a conclusion than an omission.

**4. `ProjectsDirectory` is not untouched.**

The body says it is. The code is; the behaviour is not. `FwDirectoryFinder.cs:502` is
`GetDirectory(ksProjectsDir, Path.Combine(DataDirectory, ksProjects))` — its default derives
from `DataDirectory`, which this PR changed. On any machine where the HKCU/HKLM `ProjectsDir`
value is absent — a fresh dev box, a CI agent, a container — the projects directory moves
from `%ProgramData%\SIL\FieldWorks\Projects` to `<tree>\DistFiles\Projects`.

Where the registry value is set nothing changes, which is most established dev machines, so
this is narrow. But `ProjectsDir` resolution touches installer expectations and test-fixture
scratch directories, and a reviewer reading "untouched" will not check. Please correct the
claim and state plainly what happens where the registry value is absent.

**5. Two problems in `CodeAndDataDirectory_PreferSourceTreeOverRegistry`.**

`RegistryKey.SetValue(name, null)` throws `ArgumentNullException`. If the fixture setup at
`FwDirectoryFinderTests.cs:38-39` ever stops seeding those values, this test fails with an
`ArgumentNullException` from the `finally` rather than the assertion that actually broke. Your
own preflight caught this; please guard it.

More importantly, `expectedDir` is computed as `UtilsAssemblyDir/../../DistFiles` — re-encoding
the exact fixed-depth assumption this PR removes. It passes today and misleads tomorrow, and
it sits in the file that is meant to be the evidence for the change. Derive it the way the
production code now derives it.

**6. Installed-build safety: verified, with one question.**

I checked this independently and you are right. `Build/Installer.targets:133,149` harvests
`$(fwrt)\DistFiles\**\*` and copies with `%(RecursiveDir)`, flattening the contents into the
staged app folder, so no `DistFiles` directory name survives; the only root-level repo file
installed is `License.htm` (`:134`), so no `FieldWorks.sln` either; and
`FLExInstaller/Overrides.wxi:11-12` points `RootCodeDir` at APPFOLDER. The marker conjunction
cannot occur in an install tree.

The question: **does anything in install validation or patch staging run the packaged binaries
from a path inside the checkout?** If so, the probe would prefer the tree's `DistFiles` over
the staged payload and could mask a packaging defect — a validator passing because it read the
source tree is a bad failure to have. Related: dropping `FW_USE_REGISTRY_DIRS` (item 2) removes
the only lever such a flow could have pulled, so it is worth answering this before dropping it.

**7. Make the marker's disappearance loud.**

`ksSolutionFilename = "FieldWorks.sln"` (`:334`) is the only thing distinguishing a source tree
from an install. Rename or remove the solution — plausible in an SDK-style consolidation — and
every dev build silently reverts to the registry with no diagnostic.

Please add a test that fails loudly if the marker stops existing at the repo root. That turns
"someone renamed the solution" from a confusing dev-environment regression into a red build,
which is the cheapest possible guard on the whole mechanism.

**8. `TidyRootDir`'s doc comment.**

`FwDirectoryFinder.cs:320-323` reads "Strips the trailing separator that hundreds of callers
would otherwise pass on to `Path.Combine`". It names its consumers, who change silently, and it
describes what the method does to them rather than its own contract. The pre-existing `//` said
the same thing, but this PR promoted it to a doc comment, so it is fair game.

Something closer to: returns the directory without a trailing separator, except at a drive root,
where `Path.Combine` requires one.

**9. Four smaller comment items.**

- `FindDevDistFiles` `<remarks>` (`:344-348`): "Walking up to the tree root, rather than
  assuming a fixed depth" narrates what the code no longer does; the rest enumerates three
  folder layouts, restating the `[TestCase]` list.
- `<param name="startDirectory">` (`:349`) restates the name and type only — omit it.
- `ksUseRegistryDirsVariable` (`:337`): "let the registry name the directories again" implies a
  prior state the reader cannot see. Moot if item 2 lands.
- `FindDevDistFiles_InsideSourceTree_FindsTreeDistFiles`'s doc restates the method name and
  narrates removed behaviour.

Credit where it is due: the `<summary>` null contract ("or `null` if it lies outside a source
tree") is exactly the kind of tag worth keeping, and `CreateFakeSourceTree`'s doc explains
fixture shape rather than mechanics. The comments in this file are better than most.

**10. Two documentation items.**

`.claude/skills/fieldworks-winapp/SKILL.md:123` — "Current builds anchor on their own source
tree, so this now matters mainly for older builds" — is temporal framing that will read as
wrong within a release or two. Say what is true rather than what recently changed.

And an accuracy note on the body's reasoning, which is otherwise a good record: it attributes
the shared registry slot to `Build/mkall.targets` target `setKeysInHKCU`, but `mkall.targets:277`
carries a `REVIEW (Hasso) 2026.03` comment saying those targets "appear unused by the
recently-modernized build process". The likelier live writers are the MSI
(`FLExInstaller/Overrides.wxi:11-12`) and `Resolve-FieldWorksDevRegistry.ps1:54-55`. The defect
is real either way — it would just be a shame for the record to name the wrong culprit.

---

johnml1135:
All ten items addressed in `94145a678`, and item 6's question is answered below.
Two places where I checked your claim and it changed what I did.

## 1. The walk-up on a path with a space — confirmed by running it

    StripFilePrefix('file:///C:/My%20Repos/fw/Output/Debug/FwUtils.dll')
    => C:/My%20Repos/fw/Output/Debug/FwUtils.dll

Exactly as you said. (One correction: it is `SIL.LCModel.Utils.FileUtils`, not
SIL.Core — which matters only if someone goes looking for it.)

I did not simply call `ExeOrDllDirectory`, because that would have left the fix
unguarded. The escaping happens in the *derivation*, and a test that calls
`FindDevDistFiles` with an already-clean path proves nothing about it — my first
attempt did exactly that, and reverting the fix left it green. So the conversion
is now its own testable step, `AssemblyDirectoryFromCodeBase`, and the new test
drives a `file://` URI through it. Removing the unescaping turns that test red.

## 3. Decided per site

- **`SourceDirectory`** — routed through the same walk. It was the sharpest case
 as you said: two hard-coded hops and a throw, so a build from `Output/Debug/x64`
 resolved code and data but still failed here, contradicting the layout
 `FindDevDistFiles_InsideSourceTree_FindsTreeDistFiles` claims support for.
- **ICU** (`FwUtils.cs`) — left alone, with the reason at the site and
 [LT-22768](https://jira.sil.org/browse/LT-22768). Your instinct to give this
 extra thought was right: it runs on the `CustomIcu.InitIcuDataDir` bootstrap
 path, so depending on `FwDirectoryFinder` there is an initialisation-order
 question, not a refactor. Worth noting it starts from `Location`, so it does
 *not* have the escaping bug — only the fixed depth.
- **`StringTable`** — left alone, reason at the site and
 [LT-22769](https://jira.sil.org/browse/LT-22769). It walks to the last "output"
 segment and feeds a different fallback chain, so unifying it is its own change.
 It also has a failure mode the others do not: any ancestor directory containing
 the letters "output" moves where it lands.

## 2. `FW_USE_REGISTRY_DIRS` — dropped

Your discoverability argument is the right one, and item 9's third comment
complaint went with it. Item 6 below removes your stated reason to settle that
first.

## 6. Answered: nothing runs the packaged binaries from the checkout

I traced it. The only thing executed during installer build, besides the WiX
tooling, is `Remove-StaleDlls.ps1` (`Installer.targets:189`), and it inspects
assemblies with `AssemblyName.GetAssemblyName()` — metadata only, never loading
them — so `FwDirectoryFinder` never runs. Every other `.exe` in the installer
targets is a harvested file, not an invocation. So the probe cannot mask a
packaging defect, and no validator can pass by reading the source tree.

## 4, 5, 7, 8, 9, 10

- **4** — corrected in the body, which now states what happens where the
 registry value is absent: the projects directory moves from
 `%ProgramData%\SIL\FieldWorks\Projects` to `<tree>\DistFiles\Projects`.
 Confirmed at `:541`, `GetDirectory(ksProjectsDir, Path.Combine(DataDirectory,
 ksProjects))`.
- **5** — both fixed. The expectation is now derived the way production derives
 it, and the cleanup deletes an absent value instead of calling
 `SetValue(name, null)`.
- **7** — added, and the case is stronger than stated: `CanonicalJsonTests`,
 `EngineIsolationAuditTests` and `LayoutImportCoverageTests` already walk up to
 `FieldWorks.sln` to find the repo root, so a rename breaks four things at once.
 The new test says so, and names `ksSolutionFilename` as what to update.
- **8** — rewritten to its own contract.
- **9** — all four trimmed.
- **10** — the temporal framing is gone from the skill doc, and the body no
 longer names `mkall.targets` as the registry writer. You were right about that:
 `mkall.targets:277` carries the `REVIEW (Hasso) 2026.03` saying those targets
 "appear unused by the recently-modernized build process", so the body now
 points at the MSI and `Resolve-FieldWorksDevRegistry.ps1` instead.

## Verification

**Corrected after posting.** The original version of this section claimed
`FwUtilsTests` 409/409 and listed two ablations. That overstated the coverage in
one specific way, and the author was right to question it: the `SourceDirectory`
reroute above had **no test of its own**. A pre-existing test caught an outright
break, but nothing exercised the deeper layouts, which is the entire point of
the change. `c32048433` fixes that.

`FindSourceDirectory(startDirectory)` is now an extracted seam, covered from
`Output/<Configuration>`, `Output/<Configuration>/<arch>` and a project's own
bin folder, plus the throw outside a source tree.

Current state:

- Debug build clean, comment-hygiene clean.
- `FwUtilsTests` **413/413** (`FwDirectoryFinder` 17 → 25).
- The projects that consume `SourceDirectory`, which the first pass did not run
  at all — it has ~30 call sites across ten-plus test projects:
  `XMLViewsTests` 114/114, `DetailControlsTests` 119/120, `xCoreTests` 17/17,
  `LexTextControlsTests` 353/356. The shortfalls are pre-existing ignores, at
  the same counts as an unrelated branch.

Ablation sweep, one production change at a time:

| Ablation | Caught by |
| --- | --- |
| Remove the unescaping | `FindDevDistFiles_UnderAPathContainingASpace` |
| Restore the registry's precedence | `CodeAndDataDirectory_PreferSourceTreeOverRegistry("RootDataDir")` |
| `SourceDirectory` returns the wrong directory | 4 tests, incl. all three layout cases |

Limits worth stating rather than implying:

- `CodeDirectory`/`DataDirectory` read the running assembly, so their unescaping
  is covered through the extracted seam, not end to end. Relocating a running
  assembly is beyond a unit test.
- The marker test guards `FieldWorks.sln` being removed; it does not test
  production logic.
- Five of the ten-plus `SourceDirectory` consumers were run locally; CI covers
  the rest.

_History: one force-push on this branch, `2b1098f37` → `94145a678`, identical
tree — the commit message had an 81-character body line that would have failed
`Check commit messages`. `c32048433` was a normal fast-forward._

.claude/skills/fieldworks-winapp/SKILL.md
```diff
@@ -120,6 +120,7 @@ verifies that launch will succeed.
 - `scripts/Set-FieldWorksLegacyMode.ps1`: forces `UIMode=Legacy` — run before EVERY launch.
 - `scripts/Resolve-FieldWorksDevRegistry.ps1`: aligns the dev registry (`RootCodeDir`/`RootDataDir`) to
   this worktree before launch; auto-realigns when the other worktree is idle, else prints `RESULT=ASK_USER`.
+  Current builds anchor on their own source tree, so this now matters mainly for older builds.
```
jasonleenaylor: This is item 10's temporal framing, and it is unchanged -- no commit after `2939a251a` touches this file, so "the temporal framing is gone from the skill doc" does not match the tree. The other half of item 10 did land: the body's `mkall.targets` attribution is fixed.

There is now a second problem in the same sentence. It says this script matters "mainly for older builds", while the PR body names `Resolve-FieldWorksDevRegistry.ps1:54-55` as one of the likely live registry writers. The doc and the body disagree about whether the script still matters.

While you are here: body line 17 still reads "`Build/mkall.targets` still writes `RootCodeDir`/`RootDataDir`" without the `REVIEW (Hasso)` caveat that the detailed sections now carry.

jasonleenaylor (changes requested):
Round 2. Nine of ten items are addressed, and two came back better than what I asked for.

On item 1 -- I asked you to call `ExeOrDllDirectory`, and you were right not to. Making the conversion its own seam is the stronger fix, and asserting `Does.Contain("%20")` on the fixture before using it means the test cannot quietly stop exercising the defect, which is the failure mode I was actually worried about. Item 3 is the shape I was asking for: a decision at each site with the reason next to the code and a ticket behind it. Your note that `StringTable` lands somewhere else entirely if any ancestor directory contains "output" is a better argument for leaving it alone than the one I made.

Worth saying plainly: you retracted your own verification claim before anyone challenged it, and the corrected version states its limits instead of implying coverage it does not have. That is the right instinct, and it is rarer than it should be.

One required item, inline, and one nit.

**Nit.** `FwUtils.cs` and `StringTable.cs` each gained a UTF-8 BOM along with their comment -- `2f2f20` to `efbbbf` against the merge base. `.editorconfig:12` sets `charset = utf-8`, which excludes the BOM, and 52 of the 60 C# files in that directory have none. Almost certainly the editor rather than a decision; worth stripping while you are in there.

jasonleenaylor (commented):
> @jasonleenaylor reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1093)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

.claude/skills/fieldworks-winapp/SKILL.md
johnml1135: Fixed in `b59e4fbb8` — you were right that the sentence never landed; no commit after
`2939a251a` touched the file.

The sentence now states the condition instead of the history:

> A build running from inside a source tree resolves code and data from that tree and
> ignores these values; they govern a build launched from outside one.

One narrowing on your second point. I could not write the contradiction away the way it
was framed, because the doc and the body are not quite disagreeing. The body names
`Resolve-FieldWorksDevRegistry.ps1:54-55` as a likely live *writer* of the slot, which it
is; the doc was saying the slot no longer *governs resolution* for a source-tree build,
which is also true. What was wrong was only the temporal framing — "current builds",
"now matters mainly for older builds" — so that is what I removed. I did not add a
`ProjectsDir` qualifier while I was there: the script writes `RootCodeDir` and
`RootDataDir` and nothing else, so crediting it with `ProjectsDir` would have been a new
inaccuracy in place of the old one.

Body line 17 now carries the `mkall.targets:277` caveat, matching what the detail
sections already said.

**BOM nit taken.** Both files were `2f2f20` at the merge base and `efbbbf` on the branch,
so it was mine, and `.editorconfig:12` excludes it. Stripped from both. Your file count
was off in the denominator — I get 16 of 91 `.cs` files under `Src/Common/FwUtils/` with
a BOM, not 8 of 60 — but the conclusion is the same and the two files are back to plain
UTF-8. Comment hygiene clean.

jasonleenaylor (approved):
LGTM

jasonleenaylor (approved):
> @jasonleenaylor reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1093)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on johnml1135).

PR #1095, opened 2026-08-20
LT-22691: Publish writing system under cursor

mark-sil:
The detail editor focus hook now also publishes the window-scoped WritingSystemUnderCursorChanged event, so the toolbar writing-system combo tracks the cursor while the Avalonia detail view is focused.

Notes:
- This adds the first Avalonia-to-WinForms state channel rather than replacing an adapter route.
- The view publishes an event instead of writing the property itself so the relevant code stays quarantined in the one listener that already owns that combo. The view stays ignorant of its consumers, and a replacement toolbar later subscribes to the same event without touching the view.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1095)


jasonleenaylor (approved):
:lgtm:

> @jasonleenaylor reviewed 5 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1095#-:-P-UxNuXEatTEIxcRuLw:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mark-sil).

PR #1098, opened 2026-08-21
LT-22723: Add jira-issue skill and compact style for issues and PRs

johnml1135:
Filing an LT ticket with an agent now goes through a skill instead of improvisation. `/jira-issue` interviews for who/when/where/how, searches Jira for duplicates **before** drafting, gets a three-line lede approved by the developer, caps the description at 250 words, and moves the analysis into the first comment. LT-22723 was filed by the skill itself as the worked example.

**The unknown you arrive with:** why does a writing-style skill run to 850 lines? Because Jira Data Center has no `{expand}` macro. Nothing in a description can be folded away, so length in the description is length on the screen, for every reader, permanently. Every budget in here serves that one constraint. There is no product code on this branch — Markdown plus three lines of `.gitignore`.

**Start here:** `.claude/skills/jira-issue/SKILL.md`. The nine phases are the design; the four `references/` files are detail it delegates to.

Where to look:

- **Two gates block publishing** — the duplicate table (Phase 2) and the lede approval (Phase 3). If either can be skipped quietly, the skill is decoration. Both are written as "never file without".
- **Phase 0b drops sections that do not apply.** A ticket about tooling gets no environment section and no reproduction steps. Affects Version is the one field that is set regardless. This is what stops a template being filled with N/A.
- **`compact-style.md` is shared, not copied.** `pr-pitch` reads the same file, so the two cannot drift apart.
- **`jira-bugfix`'s worktree prohibition is relaxed** from "do NOT create worktrees automatically" to "not without asking", and it gains the `LT-XXXXX-short-slug` branch convention.
- **Nothing writes to Jira unprompted.** Every write sits behind an explicit ask.

Deliberately not here: screenshots, attachments and the evidence framework, stacked behind this in the follow-up PR.

Verification: `gitlint` clean on all five commits; skill frontmatter parses; no trailing whitespace on any added line. No `build.ps1` or `test.ps1` run — nothing compiled changed. `Build\Agent\check-and-fix-whitespace.ps1` errors on a *clean* branch because it reads a log `Tee-Object` never creates, so whitespace was verified by hand instead.

Next: review, or tell me to split the `pr-pitch` and `jira-bugfix` edits out of this branch.

---

<details>

The design for this skill was written as a document in `Docs/workflows/` and then deliberately deleted before merge. It was working scaffolding, not something anyone needs in order to change this code correctly. What follows is that record, kept here because a PR body outlives a branch and costs nothing to scroll past.

</details>

<details>

**Search before drafting, not before posting.** If the ticket already exists, the work is a comment on it. Discovering that after twenty minutes of drafting wastes the drafting.

**The lede is approved before anything else is written.** Three labelled lines, rendered and shown, up to three revisions. After the third the skill asks which line is wrong rather than guessing a fourth time. Every other budget in the skill is enforceable by counting; this one is only enforceable by stopping.

**"I don't know" is recorded, not resolved.** Missing facts become a `*Not known:*` line. Nothing enters a description that the reporter did not say or that we did not verify — inferred mechanism goes to the comment, labelled inferred. Fabricated detail in a ticket becomes folklore that outlives the ticket.

**Preferences are a gitignored file, not agent memory.** `.claude/.jira-issue-prefs.json` is portable to any agent this repo supports and is per-clone, so one developer's `worktree` preference never becomes another's default.

**Branches are `LT-XXXXX-short-slug`.** The number keeps `jira-bugfix`'s existing "contains the LT number" check working; the slug is what makes a list of sixteen worktrees readable. A real session was lost hunting for "the branch for LT-22715" among branches named descriptively.

</details>

<details>

**A second skill pair mirroring `pr-preflight` / `pr-pitch`.** Two entrypoints to keep straight for a workflow that is mostly linear. The PR pair earns its split because the write-up gets re-run on existing PRs; a Jira description is rewritten far less often.

**Collapsible sections in the description.** `{expand}` is a Confluence macro, not a Jira one. Verified absent before the comment split was designed around it. Had it existed, the whole shape of this skill would be different.

**A standalone `compact-writing` skill.** A fourth skill in the chain, loadable when nobody asked for it. A reference file that two skills read costs less and cannot be invoked by accident.

**Duplicating the style rules in each skill.** Rejected because they would drift, and the drift would be invisible until someone noticed tickets and PR bodies had different rules about the same thing.

</details>

<details>
<summary><b>The i-have-adhd adaptation</b></summary>

`compact-style.md` adapts the MIT-licensed [i-have-adhd](https://github.com/ayghri/i-have-adhd) skill. That skill shapes **chat turns**; these rules shape **written artifacts**, which changes several of them:

- "Lead with the next action" becomes "lead with what the reader must decide", because a triager's next action is a triage decision.
- "Give specific time estimates" becomes **cost shape, not hours** — "needs a liblcm release and a package bump" rather than "about two days". Hours on a ticket read as a commitment nobody in the conversation is authorised to make.
- "Restate state every turn" does not apply to filing at all; it applies to later comments.
- "Make completed work visible" was dropped — that is PR territory, not ticket territory.

Two rules were added that have no chat equivalent: **say what is not known**, and **never assert what was not verified**.

</details>

<details>

**The reference failure is real, not hypothetical.** LT-22715's first rendered line is `h3. The underlying problem`; its description runs past a thousand words; and it states in its own text that "that single missing distinction produces four separate user-visible problems". `references/examples.md` carries the before-and-after in full, with the rewrite at 228 words.

**The skill was used to file its own ticket.** LT-22723 went through Phase 0 (type), Phase 0b (relevance — tooling, so no environment or repro sections), Phase 2 (three JQL passes, no duplicate), Phase 3 (lede approved by the developer before drafting), and Phase 6. Description came in at 232 words against a 250 budget; summary at 60 characters against 80.

**Jira API surfaces were read, not guessed.** `jira_create_issue` sets `{'accountId': ...}` for `assignee`, which is Cloud-only — SIL Jira is Data Center and needs `{'name': ...}`, so the skill documents the `custom_fields` fallback. Confirmed by filing LT-22723 that way.

**Unrelated defect found in passing, not fixed here:** `.claude/skills/atlassian-readonly-skills/scripts/jira_projects.py` raises `NameError: name 'Optional' is not defined` on import — a missing typing import. Out of scope for this branch.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1098)


johnml1135:
Consolidated into #1100, which now targets `main` and carries all eleven commits from this branch, #1099 and #1102. Nothing is lost -- the commit boundaries survive in history. Closing.

PR #1101, opened 2026-08-21
Reduce the Atlassian skills to Jira, and fix eight broken modules

johnml1135:
**Start here:** `.claude/skills/atlassian-skills/SKILL.md` — the Configuration and Core Workflow sections show the shape of the whole change.

The two Atlassian skills supported Jira, Confluence and Bitbucket. FieldWorks uses Jira. Confluence and Bitbucket are gone — **4,361 deletions against 322 insertions** — and the read-only modules that never worked are fixed.

**Rebased onto #1100, which changes what is left here.** #1100 landed the same import fixes for `jira_workflow.py` and `jira_projects.py`, so both files are now byte-identical to `main` and have dropped out of this diff entirely. Of the nine broken modules this branch originally repaired: two are already on `main`, three are Confluence modules it repairs and then deletes, and **three are what this PR still delivers** — `jira_agile.py`, `jira_links.py`, `jira_worklog.py`.

Where to look:

- **Nine of twelve read-only modules were unusable.** Eight raised `NameError` on import, so any read through them failed and callers reached into the *write* skill to do a read. The ninth, `jira_projects`, imported cleanly and raised an unimported `NotFoundError` only when a project was absent — a successful import is not evidence a module works. That ninth one now ships on `main` rather than here.
- **The cause was upstream's, and upstream had already fixed it.** The bug arrived in `cdd1823f6` (2025-12-20) and was fixed in `0fafb48e7` (2026-02-10); `d1a9bc66d` vendored the pre-fix state on 2026-06-11. Our files are byte-identical to upstream at `cdd1823f6` — a stale snapshot, so a re-sync would have *fixed* this, not reintroduced it. Upstream #14 is closed as already-fixed.
- **Confluence and Bitbucket removed to the roots** — not just the script modules but their docs, config blocks, credential fields, the CQL reference, and the plumbing in `_common.py`. No functional references remain; each `SKILL.md` keeps one line recording that the support was removed, and `PROVENANCE.md` carries the detail. What remained was rewritten rather than left with holes, and the read-only variant lost 237 lines of function signatures that `REFERENCE.md` already documents.
- **Both frontmatter descriptions advertised Confluence and Bitbucket** and are rewritten. An overstated description is how the wrong skill gets loaded.
- **`PROVENANCE.md` records two local modifications that were unlisted**: TLS verification is off in three places, because `jira.sil.org`'s certificate chain does not validate here, and the write variant's "no script changes" was untrue.
- **`jira_attachments` from #1100 survives the rewrite.** The rebase kept its documentation inside the rewritten Core Workflow section, and the module still imports — worth stating because this PR cuts that file from 749 lines to 390.

Deliberately not here: `_common.py` keeps `check_available_skills` even though only one service can now be unavailable — load-bearing for the credentials path, cheap to leave. `REFERENCE.md` gains no `jira_attachments` entry; `main` does not document it there either, so that belongs to whoever extends the reference, not to this removal.

Verification: **`pyflakes` reports nothing across both variants** and all 19 modules import (10 write, 9 read-only). Live through the read-only skill: `jira_get_issue('LT-22723')`, `jira_get_transitions`, and `jira_get_project_versions` for both a real and an absent project. `gitlint` clean. No `build.ps1`/`test.ps1`: Python utilities outside the build.

Next: review and merge. Independent of #1098/#1099/#1102. #1100 has merged and this branch is rebased onto it, so the two-line overlap it used to carry is gone.

---

<details>
<summary><b>Size before and after</b></summary>

| | Before | After |
| --- | --- | --- |
| `atlassian-skills` (md + py) | 5,778 | **3,497** |
| `atlassian-readonly-skills` (md + py) | 4,014 | **2,256** |
| Script modules | write 18, read-only 17 | **write 10, read-only 9** |
| Broken modules (read-only) | 9 | **0** |

`SKILL.md` specifically: write variant 749 → 390 lines, read-only 560 → 141.

</details>

<details>
<summary><b>What the licence investigation found</b></summary>

Both skills carry `license: Complete terms in LICENSE` in frontmatter, which is what stopped an earlier pass from touching them at all. Checking properly:

| Question | Answer |
| --- | --- |
| Where from? | https://github.com/langpingxue/atlassian-skills, vendored in d1a9bc66d |
| Declared licence? | **MIT**, in the upstream README |
| `LICENSE` file upstream? | **No.** GitHub's license endpoint 404s |
| Copyright line anywhere? | **No** |
| `LICENSE` file here? | No — the frontmatter reference dangles in both places |

MIT permits modification, so the earlier caution was unnecessary. The attribution it does ask for was missing and now exists as `PROVENANCE.md` in each skill. Since upstream names no copyright holder, a verbatim MIT text could not honestly be authored on their behalf; the provenance files record the declaration and its gaps instead.

**This is now a hard fork.** A re-sync is no longer a merge — upstream becomes a source to cherry-pick Jira fixes from. `PROVENANCE.md` says so.

</details>

<details>
<summary><b>Data Center gotchas captured while investigating</b></summary>

Upstream's docstrings describe Jira **Cloud**; SIL's instance is Data Center. Three that silently disagree, now in `atlassian-skills/PROVENANCE.md`:

- `assignee` wants a username, not an `accountId`. `jira_create_issue` and `jira_update_issue` both send `{"accountId": ...}`, which Data Center rejects. Pass `custom_fields={"assignee": {"name": "<username>"}}`.
- Affects Version is not exposed at all. Pass `custom_fields={"versions": [{"name": "FW 9.3"}]}`.
- `resolution` cannot be set by an update — not on the edit screen; only a transition sets it.

Each was found by hitting it. They are duplicated into `jira-issue/references/publish.md` on #1098, where someone working an LT ticket looks first.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1101)


jasonleenaylor (approved):
Verified by running it: at the merge base the read-only skill had exactly the eight `NameError` modules PROVENANCE lists, `jira_projects` and `jira_workflow` do import clean on main so their absence from the diff is real, and the three remaining — `jira_agile`, `jira_links`, `jira_worklog` — are what this delivers. All nineteen modules across both skills import clean at the head. The ninth defect being runtime-only, and saying so, is a good catch.

PR #1105, opened 2026-08-21
LT-22728: Scope local libraries to one build

johnml1135:
A local library pack can no longer wear a published package's version string, which is what LT-22728 is really about. `.\build.ps1 -LocalLibraries lcm` packs your local liblcm and builds against it in one command.

**Revised after review.** The first version of this branch moved selection into an untracked `Build/LocalLibraries.props` that every ordinary build deleted, so returning to published packages was automatic. That is gone. `SilVersions.props` is written exactly as before: tracked, visibly dirty, cleared by hand. The dirty pin is not merely an undo — it is the record that this tree depends on a library version nobody has released yet, and an auto-revert would erase that release gate rather than tidy up. `-LocalLibraries` is now a wrapper over `Manage-LocalLibraries.ps1`, not a second way to select a library.

**The unknown a reviewer starts with: why did the old code evict cache entries so aggressively?** Because a local pack reused the published version string, and NuGet resolves an already-extracted `(id, version)` before consulting a folder feed — so a local pack could shadow the published package, or be shadowed by it, with nothing to tell them apart. SIL.Machine showed it plainly: no GitVersion, so it packed as a flat `3.9.2`, identical in name to nuget.org's. The eviction was a workaround for a naming problem. Fix the name and the cleanup becomes ordinary housekeeping.

**Where to look**

- **Version derivation** (`LocalLibraries.psm1`, 102 lines) — `<core>-<branch>.<commit>`, or `.dirty` when the checkout has uncommitted changes. A published version can no longer be reused.
- **`-p:DisableGitVersionTask=true`** — GitVersion assigns `Version` inside an MSBuild target, which outranks a command-line property. Without it the stamp is accepted and then silently overwritten. The produced package is now checked against the stamp, so an override fails loudly.
- **Machine PDB paths** — the base was `bin/Debug/netstandard2.0` at the checkout root, but Machine writes per project under `src/<project>/bin/Debug`. Symbol copying had been doing nothing for the one library this ticket is about, and saying nothing about it. `PdbRelativeDir` takes a list now, and a miss names every directory searched.
- **Feed location** — packs default to `.localfeed` in the working tree; `LOCAL_NUGET_REPO` still wins when set. The feed path is recorded in `SilVersions.props`, which `Build/PackageRestore.targets:28` already imports, so a nested restore in its own process finds it. That is what let `dotnet nuget add source` go: a machine-level source outlives the checkout that wanted it.
- **`build.ps1 -LocalLibraries`** — packs the named libraries, then builds. Roughly 30 lines.
- **`powershell-compat.ps1`** — scans all of `Build`, not just `Build/Agent`, because `build.ps1` loads modules from there under Windows PowerShell 5.1. This is what caught an em dash that broke the 5.1 parse earlier in review.

**Deliberately not here**

- No automatic revert. See above — that is the release gate, and removing it was the thing review pushed back on hardest.
- No `Build/LocalLibraries.props`, no `Exec` boundary, no worktree helpers. All existed only to serve the untracked-selection design.
- Clearing a pin is manual: set the released version when the library ships, or `git checkout Build/SilVersions.props`.

**Verification**

`build.ps1 -CommentHygiene -TokenHygiene -BuildTests` clean, exit 0. `comment-hygiene`, `token-hygiene` and `powershell-compat` (5.1 and 7.0) all clean. `Build/LocalLibraries.Tests.ps1` 22 assertions pass, and the gating was checked both ways: a default `test.ps1` run does not invoke it, `-LocalLibraryTests` does. The version functions were exercised against the real `machine` checkout, stamping `3.9.2-conformance-fieldworks-w.d3b7643` — not equal to published `3.9.2`, which is the whole property. `Set-LocalFeedSource` was checked on a copy of `SilVersions.props` for valid, idempotent, tidily-indented XML. `gitlint` clean.

Not exercised: a full local pack of a real library end to end. The version derivation, the feed write and the gating are covered; `dotnet pack` itself is not run in CI.

---

<details>

The reasoning behind this branch lives here rather than in the tree, on purpose.
The investigation that produced it was a one-time diagnosis, and
`Docs/architecture/local-library-debugging.md` deliberately carries only what a
developer needs in order to *use* the workflow.

The one fact worth keeping: **NuGet's resolution order is global-packages folder
→ non-HTTP sources (folder feeds) → http-cache → HTTP.** If `(id, version)` is
already extracted in `packages/`, the folder feed is never consulted at all.
Every design choice below follows from that sentence.

</details>

<details>

**Content-addressed, not branch-addressed.** An earlier draft named packages by
branch alone. That is a *stable* string, so a second pack from a dirty tree
would hit the extracted cache and silently serve the previous build. The commit
hash makes a clean version identify its contents, which is also what makes pack
reuse correct rather than a gamble. A dirty tree cannot be identified this way
at all, so it is marked `.dirty` and always repacked.

**Untracked files count as dirty.** The rule is "whatever git would report",
which honours `.gitignore` and so excludes `bin`/`obj`. A new source file added
while prototyping changes the build without changing the commit, so treating it
as clean would serve a stale package. The cost is that a stray note beside the
source keeps the slow path, so the build now names the offending paths.

**The core version comes from the library's own GitVersion**, probed with
`dotnet msbuild -restore -t:GetVersion`, so a version bump in the library shows
up in the local package. `-restore` is required: the target ships inside the
GitVersion package, which a never-built checkout has not restored yet.
SIL.Machine has no GitVersion and falls back to the consumed version.

**Discovery before creation.** `git worktree add` refuses a branch already
checked out elsewhere. Rather than fight that, the setup flow reads
`git worktree list --porcelain` and uses whatever already holds the branch —
both the fast path and the only path git allows.

**Build before pack.** A package may include output from a target framework its
own project does not build, and packing first fails on the missing file.

</details>

<details>

**A branch name in the config filename** (`localLibs.<branch>.props`). In a git
worktree `.git` is a *file*, not a directory, so discovering the branch at
MSBuild evaluation time — the only point at which `PackageReference` items can
be declared — breaks in exactly the multi-worktree case that motivates this
work. Branch names also contain `/`, a detached HEAD has no branch name, and
renaming a branch would silently drop the config. Go, Cargo and Gradle all use a
fixed filename in a per-checkout location for the same reason.

**Publishing per-branch prereleases from the library repositories** so a
FieldWorks PR could point at an unreleased library. liblcm already does this to
GitHub Packages for same-repo PRs. Dropped deliberately: it needs a
`read:packages` token for every developer and every CI job, and the intended
policy is that only a real release passes CI.

**Nesting a dependency's worktree pinned to a live branch.** No established tool
does this. Chromium's `DEPS`, Zephyr's `west` and Android's `repo` all
materialise to a detached HEAD or a pinned revision, and `west` does so
precisely to avoid the same-branch-checked-out-twice refusal.

**Having the build create the worktrees.** Every comparable tool keeps sync as an
explicit command. A build that mutates the source tree outside its output
directory is the failure mode, which is why the build only verifies that a
checkout is where a path variable says it is, and creates nothing.

</details>

<details>

**GitVersion silently wins.** `-p:Version=` is a global property, but
GitVersion.MsBuild assigns `<Version>` inside a *target*, and a target can
override a global property. This was found only because the first end-to-end
test ran against SIL.Machine — the one library with no GitVersion, and therefore
the one case that could not exhibit the bug.

**A build can report using a local library while compiling against the published
one.** MSBuild global properties do not cross an `Exec` boundary, and
`PackageRestore.targets` restores through one, reached by the native build before
the managed traversal. The first end-to-end run printed "Using local libraries:
lcm", exited 0, and left every one of 120 `project.assets.json` files naming the
published version. Reviewer analysis predicted this before it was measured; the
earlier evidence could not have caught it, because the only library tested end to
end was SIL.Machine, whose local version equalled the pin and made the override a
no-op.

**`git rev-parse --git-common-dir` returns a path relative to the caller**, not
to the repository. A first cut used it to locate `.git/info/exclude` and
silently resolved against the wrong directory; `--path-format=absolute` is
required. The same trap applies to locating a sibling checkout from inside a
worktree, where the parent directory is `.tmp/worktrees` rather than the
repositories root.

**L10NSharp was not broken by this change.** It could not be packed at all —
`NU5026`, missing `output/Debug/net461`. The discriminating test was to pack
with the original flags, which failed identically, and the main checkout had no
such output either. Its packages include output from target frameworks its own
projects do not build, so it needs a full build first. It packs cleanly now.

**Two libraries' symbol directories named paths that are never written**, so the
PDB copy did nothing and said nothing. A miss is now reported together with the
directories that were searched.

</details>

<details>

This branch does not establish a way to reference an unreleased library from a
merged commit. Derived versions are passed to restore and MSBuild as properties
and never enter `SilVersions.props`; `.localfeed` is gitignored; the feed is
never added to `nuget.config`. A pushed branch therefore carries no reference to
a local package, and CI restores from nuget.org only. If a FieldWorks change
needs a new library API, it cannot go green until that library is released —
the intended constraint, not an oversight.

</details>

<details>

- **Per-working-tree library paths.** `LIBLCM_PATH` and its siblings are machine
  global, so two working trees cannot use two different branches of the same
  library. The fix is the pattern this repository already uses twice
  (`GlobalInclude.properties`, `LibraryDevelopment.properties`): a fixed-name,
  gitignored `Build/LocalLibraries.props` imported from `Directory.Build.props`.
  `Build/Localize.targets` still records the original move away from that
  pattern.
- **A named CI check**, so a PR waiting on a library release fails legibly
  instead of as a compile error.
- **Wider PDB coverage** for L10NSharp and SIL.Machine beyond the directories
  now configured.

</details>

<details>

**GitVersion override, libpalaso, same commit:**

| | package name |
|---|---|
| with the new flags | `SIL.Core.18.0.0-lt22728-vp.1e46149.nupkg` |
| without them | `SIL.Core.18.0.0-lt22728-vp0033.nupkg` |

**All five libraries, packed from a clean worktree:**

| Library | GitVersion | Packages | Repeat build |
|---|---|---|---|
| libpalaso | 5.11.1 | 26 | 9s → 2s |
| liblcm | 5.6.10 | 6 | verified incl. restore and revert |
| chorus | 5.10.3 | 6 | 5s → 2s |
| L10NSharp | 5.11.1 | 4 | 6s → 2s |
| SIL.Machine | none | 2 | 7s → 0s |

**Full chain, liblcm.** A probe class committed to a scratch branch produced
`11.0.0-lt22728-e2e.08cb359`. The marker was present in the packed `net462`,
`net8.0` and `netstandard2.0` assemblies; `dotnet restore` resolved it in 8s;
the published `11.0.0-beta0178` and the local version coexisted in `packages/`,
with `.nupkg.metadata` recording the filesystem feed as the source; an ordinary
build then cleared 9 local cache entries and 18 feed packages and left
`beta0178` intact.

**Symbol fix, SIL.Machine.** `Output/Debug` went from 0 to 2 `SIL.Machine` PDBs,
where the previously configured path reported the directory missing.

**Test assertions were mutation-tested** rather than assumed: flipping
`.localfeed`, the reuse message, and the L10NSharp symbol directory each made
the suite fail, and pass again once restored.

**End to end through `build.ps1`.** With `-LocalLibraries lcm`, 114
`project.assets.json` files named `11.0.0-feature-grammar-json-exp.dirty`; the six
still naming the published version were stale from four days earlier and untouched
by the run. An ordinary build then removed the overrides, emptied the feed, and
returned all 114 to `11.0.0-beta0178`.

**Not run:** the full managed test suite. The diff is PowerShell, markdown,
`.gitignore` and `nuget.config`, with no compiled code.
</details>

<details>

# Code Review Summary

**Branch**: LT-22728-local-library-selection

**Base**: origin/main

**Date**: 2026-08-26

**Files changed**: 10

## Overview

Local library packs reused the published version string. NuGet resolves an
already-extracted `(id, version)` from the packages folder before it consults a
folder feed, so a local pack could shadow the published package or be shadowed
by it with nothing to distinguish them. SIL.Machine showed it plainly: with no
GitVersion it packed as a flat `3.9.2`, byte-different from but identically
named to the package on nuget.org.

The branch derives each pack's version from the checkout as
`<core>-<branch>.<commit>`, keeps the local feed inside the working tree, builds
each library before packing it, and adds a setup command that makes a library
branch available as a worktree without disturbing uncommitted work.

## Contract/API Changes

No public API change. Build-surface changes only:

- `build.ps1 -LocalLibraries` no longer requires `LOCAL_NUGET_REPO`; the feed
  defaults to `.localfeed` in the working tree and the variable still overrides.
- `Build/LocalLibraries.psm1` exports twelve new functions.
- `Manage-LocalLibraries.ps1` now stamps `-p:Version` and sets
  `-p:DisableGitVersionTask=true` while packing.

None.

- [x] **`Docs/architecture/dependencies.md` quick start still set
  `LOCAL_NUGET_REPO`** _(fixed during review: removed the line and reworded the
  surrounding text; it contradicted the updated local-library-debugging.md)_
- [x] **L10NSharp could not be packed at all (`NU5026`, missing net461/net48
  output)** _(fixed during review: each library is now built before it is
  packed, because a package may include output from a target framework its own
  project does not build)_

### Minor - Consider

- [x] **`PdbRelativeDir` named directories two libraries do not write**, so the
  symbol copy did nothing and said nothing _(fixed during review: l10nsharp now
  uses `output/Debug/net48`, machine takes one directory per pack project, and
  a miss now reports the directories searched. Proven on machine: symbols in
  Output/Debug went from 0 to 2 where the old path reported nothing)_
- [x] **`test.ps1` ran `LocalLibraries.Tests.ps1` unconditionally**, ignoring
  `-TestProject` and `-TestFilter` _(fixed during review: it now runs only when
  neither is given)_
- [x] **`nuget.config` `<clear />` dropped inherited user-level sources with no
  hint** _(fixed during review: the comment now says a private feed belongs
  here rather than in user-level configuration)_

Run and passing:

- `Build/LocalLibraries.Tests.ps1` - passes; 55 assertions. Mutation-tested
  twice (`.localfeed` to `.wrongfeed`, and the reuse message) to confirm the new
  assertions actually fire rather than passing vacuously.
- `Build/Agent/comment-hygiene.ps1 -BaseRef origin/main` - clean.
- `gitlint --ignore body-is-missing --commits origin/main..HEAD` - clean.
- All five libraries packed from a clean worktree: every produced package
  carries the derived version, and a second run reuses it.
  - liblcm 6 packages, libpalaso 26, chorus 6, L10NSharp 4, SIL.Machine 2.
  - Reuse timings: libpalaso 9s to 2s, chorus 5s to 2s, L10NSharp 6s to 2s,
    SIL.Machine 7s to 0s.
- GitVersion override proven real on libpalaso: with the new flags
  `SIL.Core.18.0.0-lt22728-vp.1e46149.nupkg`; without them
  `SIL.Core.18.0.0-lt22728-vp0033.nupkg`.
- Full chain on liblcm: a committed change reached the restored assembly in
  net462, net8.0 and netstandard2.0; the published `11.0.0-beta0178` and the
  local version coexisted in the cache; an ordinary build then cleared 9 local
  cache entries and 18 feed packages and left `beta0178` intact.

Not run:

- `./build.ps1` full compile of FieldWorks against a local library. Pack and
  restore are verified; the subsequent compile is not.
- `./test.ps1` full managed suite. The diff is PowerShell, markdown, gitignore
  and nuget.config only, with no compiled code.
- `./Build/Agent/Setup-InstallerBuild.ps1 -ValidateOnly` - no installer or WiX
  files changed.

## Positive Observations

- The aggressive cache eviction already on the branch turns out to be the
  sanctioned mitigation: NuGet documents no restore flag that bypasses the
  extraction cache for a version already present.
- Scoping the feed to the working tree made that eviction safe, and let the
  post-pack clear be deleted; `Manage-LocalLibraries.ps1` is net shorter.
- Every git call in the new code is read-only except one `worktree add`, plus a
  `fetch`. A test asserts no `switch`, `checkout`, `reset` or `clean` appears.

## Interview Notes

- Author's boundary: nothing but a real NuGet release may pass CI. Confirmed
  satisfied - derived versions travel as MSBuild properties and never enter
  `SilVersions.props`, `.localfeed` is gitignored, and the feed is never added
  to `nuget.config`, so a pushed branch carries no reference to a local package.
- `-p:DisableGitVersionTask=true` also disables GitVersion's `DefineConstants`
  and assembly-info stamping, so a local package's assembly metadata differs
  from a CI-built one. Author accepted: local versions cannot be mistaken for
  real ones and dirty detection says when to rebuild. The narrower
  `-p:UpdateVersionProperties=false` was offered and not taken.
- Author asked for the L10NSharp failure to be investigated rather than
  documented, on the grounds that it was probably already solvable. That was
  correct: it needed a build before the pack, not a toolchain change.
- Untracked files count as dirty by author's explicit choice, to keep the rule
  to what git itself would report rather than classifying file types.
- Author asked for every minor finding in the local library area to be fixed
  rather than deferred, so all three were addressed here.

## In-Review Quality Check

- `Docs/architecture/dependencies.md` corrected; markdown fences verified
  balanced after a bad `sed` deleted the wrong line mid-edit.
- Build-before-pack added, then L10NSharp re-verified from a fresh worktree.
- Comment hygiene re-run after each edit; one over-budget comment shortened.
- All five library repositories audited against snapshots taken before any
  work: branch, HEAD, worktree list, branch list, status and `.git/info/exclude`
  all match, including three repositories' pre-existing uncommitted files.

## Suggested Review Focus

- [ ] Whether packing should stay build-then-pack for every library, or only
      where a package spans target frameworks.
- [ ] The dirty-tree rule: any untracked file forces a repack, which is safe but
      means a stray note beside the source keeps the slow path.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1105)


jasonleenaylor (changes requested):
The diagnosis in LT-22728 is correct: an extracted package in `packages/` survives the
deletion of its source `.nupkg`, because `nuget.config` points `globalPackagesFolder` at the
repository's `packages` directory and NuGet does not invalidate an extracted ID/version when
the feed copy disappears. That is a real trap and worth closing.

My concern is scope. Before going through the individual findings, I want to put the
proportionality question first, because it may make most of them moot.

## The reported bug is really two, with very different severity

**Mode 1 — the version pin was never reverted.** This is what the ticket's repro actually
does: pack Machine, build, delete the `.nupkg`, build again. `Manage-LocalLibraries.ps1` wrote
the local version into `Build/SilVersions.props`, so the second build resolves the local
package because the pin still names it. This affects all five libraries — and it is
**visible**: `SilVersions.props` is tracked, `git status` shows it dirty, and the script prints
"To revert: git checkout Build/SilVersions.props" immediately after packing. The only genuinely
surprising part is that deleting the `.nupkg` produces a silent success rather than a restore
failure, because the stale extraction covers for it.

**Mode 2 — the local pack produced the same version string as the published pin.** Then even
after a correct revert, the extracted local build masquerades as the published package. No
dirty file, no signal, nothing to notice. This is the dangerous one, and in practice it is
**specific to SIL.Machine**: `Build/SilVersions.props:20` pins `SilMachineVersion` to `3.9.2`,
and Machine's local pack produces `3.9.2`. For palaso, lcm, chorus and l10nsharp a local
checkout produces a different version, so once the pin is reverted the cached local package is
inert.

## A smaller fix for the dangerous half

Make a local pack produce a version that cannot collide with a published pin — a `-local`
suffix, a build stamp, anything distinguishable. Then:

- A local package can never masquerade as the published one, for any library, including Machine.
- Mode 2 disappears entirely.
- Mode 1 remains, which is acceptable: it is already visible in `git status` and already has a
  printed revert instruction. If you want to harden it further, the cheap addition is a warning
  at pack time when the packed version equals the pin — the single case where a stale cache
  entry can quietly change a later build.

That closes the silent failure without changing how anyone works.

## What the current approach costs

As written, this PR changes the local-library workflow for all five libraries in service of a
problem whose silent form belongs to one:

- `Manage-LocalLibraries.ps1 -Palaso` (and the rest) now throws, redirecting to
  `build.ps1 -LocalLibraries palaso`.
- The persistent user-level NuGet source is removed, and `nuget.config:31` adds `<clear />`,
  which discards **every inherited user- and machine-level package source for every restore in
  this repository** — including for developers who never run a local-library build.
- Every build, whether or not `-LocalLibraries` was passed, sweeps both the package cache and
  `$env:LOCAL_NUGET_REPO`, a user-owned folder outside the repository that may be shared with
  other SIL projects and is shared across worktrees.
- Selection moves from a visible edit to a tracked file to an invisible per-invocation MSBuild
  property.

The persistent local source in particular is the intended path for painless local library
development, not a defect to be eliminated. It is also inert by construction: with exact
version pins in `Directory.Packages.props`, a local feed containing `SIL.LibPalaso 15.0.1-local`
is never consulted while the pin says `15.0.0`. It becomes live only when a version override
selects it — which is exactly the per-invocation mechanism you built. Please drop `<clear />`
and keep the persistent source.

Worth knowing, since it is directly upstream of all of this: `Build/NuGet.targets` used to
carry a `CleanNuGet` target that did `ForceDelete Files="$(fwrt)/packages/"` — a standalone,
manually invoked wipe of the package cache. It was removed in `5711bf6be` ("Modernize .NET
tooling and enable AI workflows", #678) when that file was deleted, and nothing replaced it.
The gap this PR is filling with automatic cleanup was previously filled by an explicit one.

## Cleanup should be deliberate and visible, not automatic

Local library development is uncommon, deliberate, developer-driven work. The person doing it
knows when they have finished. That argues for a well-known post-activity cleanup rather than a
sweep on every build — and it dissolves most of the machinery in this PR:

- Cleanup can afford to be blunt, because the developer asked for it. `packages/` is repo-local
  and disposable; restore repopulates it. Prefix matching over the library families is
  sufficient — which is what `Update-VersionAndClearCache` already does today via
  `CachePrefixes`, and what this PR replaces with `.nupkg.metadata` inspection.
- The metadata inspection is what creates the failure modes below. Dropping it removes them
  rather than requiring them to be fixed:
  - `LocalLibraries.psm1:145` reads `$metadata.source` outside the `try/catch` at `:138-144`,
    under `Set-StrictMode -Version Latest`. A version-1 `.nupkg.metadata` (older NuGet) has only
    `version` and `contentHash`, so a long-lived `packages/` folder containing one aborts the
    build with a property-not-found error. The fixture only ever writes `version = 2`.
  - A version directory with no `.nupkg.metadata` at all is skipped (`:140-142`), so a partially
    extracted package survives the cleanup that exists to remove it.
  - The cleanup at `build.ps1:591-592` sits outside every guard. `-LocalLibraries` is correctly
    refused with `-SkipRestore` (`:597-598`), but the reverse order is not: run
    `build.ps1 -LocalLibraries machine`, then `build.ps1 -SkipRestore`, and the second run
    deletes the package and then skips the restore that would replace it.
- **Raise the visibility where the developer already is:** print the cleanup command at the end
  of a `-LocalLibraries` build. The person who just did the uncommon thing is told, in that
  moment, how to undo it. That is the part worth engineering.

## If the current approach is kept, these still need fixing

**The version override does not reach the nested restore.**
`Build/PackageRestore.targets:100-119` `Exec`s a fresh `dotnet restore` carrying only
`/p:Configuration` and `/p:Platform`. The native build reaches it every run
(`mkall.targets:73` `DebugProcs` -> `CopyDlls` -> `downloadDlls` -> `RestorePackages`), and
native runs first. MSBuild global properties do not cross an `Exec` boundary, so that restore
resolves versions from `SilVersions.props` — the published values — with no local feed, and
rewrites `project.assets.json` before the managed traversal compiles.

I have not run this, so I am not asserting it as fact. What I am confident about is that the
existing evidence cannot settle it: the verification is a local `machine` 3.9.2 build, and
`SilVersions.props:20` pins `SilMachineVersion` to `3.9.2`, so the override was a no-op and the
nested restore was harmless by construction. Machine is simultaneously the only library that
was tested and the only one that could not have exposed this.

Please run it end to end with **palaso or lcm**, where the local version differs from the pin.
Either the binary is built against the local package and I am wrong, or it is not and the gap
is real. Both outcomes are worth having before merge. If it needs closing, a generated,
gitignored `Build/LocalLibraries.props` imported by `Directory.Packages.props` would reach every
MSBuild process including nested `Exec`s — and matches the `LibraryDevelopment.properties`
precedent.

**The test harness never fails properly.** `Build/LocalLibraries.Tests.ps1:2` sets
`$ErrorActionPreference = 'Stop'`, so the first `Write-Error` at `:129` is terminating: only one
failure ever prints and `exit 1` at `:131` is unreachable. `Build/Agent/CommentHygiene.Tests.ps1`
— the file this was modelled on — uses `Write-Host -ForegroundColor Red` in the loop and then
`exit 1`. The structure was copied; the detail that makes it work was not.

**The new PowerShell escapes the repository's compatibility check.** `LocalLibraries.psm1` and
`LocalLibraries.Tests.ps1` sit in `Build/`, not `Build/Agent/`, so `powershell-compat.ps1:52-55`
does not scan them and `.github/workflows/CI.yml:34-50` does not run them under both PowerShell
5.1 and 7 — even though `build.ps1` and `test.ps1` load the module under 5.1 in CI.

**Six of the 25 assertions are invalidated by the changes above**, so the test rework is larger
than it looks. Two are incorrect: `:102` (`$managerText -notmatch 'dotnet nuget add source'`)
pins the absence of the persistent source, and `:119` pins the presence of `<clear />`. Four are
obsolete: `:63` and `:65` test the filesystem-versus-HTTP `source` discrimination, and `:110`
and `:114` test `build.ps1` wiring that goes away.

The eleven behavioural assertions survive a design change; the twelve source-text regexes do
not — several would fail on a rename that preserved behaviour exactly, and two now make
correcting a design decision look like breaking tests. **Given how much of it the adjustments
invalidate, what does the remainder need a 133-line file and a bespoke harness for?** That is a
genuine question, not a rhetorical one — if the surviving behaviour justifies it, keep it.

**Smaller things:**

- `build.ps1:614-617` invokes `Manage-LocalLibraries.ps1` with `&` and then tests
  `$LASTEXITCODE -ne 0`. That script signals failure by `throw`, never by an exit code, so the
  check reads whatever native process ran last — or, if `$LASTEXITCODE` is unset,
  `$null -ne 0` is true and the build throws "Local library packing failed." spuriously. The
  real failure path is the exception, which is already handled.
- Pack order is now load-bearing and documented nowhere. It was
  `# Pack order: libpalaso first (other libraries may depend on it)` with an explicit list; it
  is now `$PackOrder = @($LibraryConfig.Keys)` (`Manage-LocalLibraries.ps1:112`), taking its
  order from the declaration order of an `[ordered]` hashtable in a different file. Alphabetising
  the catalogue — the obvious future tidy-up — silently breaks dependent packs, and no test
  covers ordering. Restore the comment at both sites.
- The PR body says selected libraries "evict same-version published cache entries". The
  function's own synopsis (`LocalLibraries.psm1:85-88`) correctly says "every cached version",
  and the test at `:83-86` evicts `3.9.3` while packing `3.9.2`. The code is right; the body
  promises less than it does.
- `build.ps1:683-686` (`FwBuildTasks` restore/build) and `:564-566` (native freshness refresh)
  run without the `/p:Sil*Version` properties or the local feed — same class of gap as the
  nested restore, lower risk.
- The five library names now live in four places: `LocalLibraries.psm1:4-42`, the `-Library`
  ValidateSet, the `-LocalLibraries` ValidateSet at `build.ps1:202-203`, and the docs table.
  The test asserts the catalogue has five entries but not that the ValidateSets match it.

## Comments

The three `.SYNOPSIS` blocks in `LocalLibraries.psm1` (`:85-88`, `:106-109`, `:114-117`) are
one accurate sentence each, stating only their own contract — those are right, and so are the
rewritten `.EXAMPLE` blocks in `Manage-LocalLibraries.ps1:20-27`.

Four things to fix:

- `Manage-LocalLibraries.ps1:69-70` — `.PARAMETER VersionOutputPath`, "JSON output consumed by
  build.ps1 for invocation-scoped version overrides." Describe the file the parameter names, not
  who reads it; callers change silently.
- `nuget.config:33-36` — the rewritten comment keeps "See
  Docs/architecture/local-library-debugging.md for the full workflow." Editing the block was the
  moment to drop the `.md` pointer.
- Two genuine WHY comments were lost in the move. `# Pack only the projects FieldWorks uses
  (avoids native CMake deps)` explained something the code cannot show and arrived bare at
  `LocalLibraries.psm1:38-41`; the pack-order rationale is gone from both files.
- The one genuinely non-obvious thing is uncommented: the whole design turns on
  `.nupkg.metadata`'s `source` field being a safe local-versus-published discriminator
  (`:56-66`, `:145`), and nothing says why. If the metadata approach survives, that comment is
  the one worth writing.

---

johnml1135:
Thank you — this review found a bug that would have shipped, and the analysis was right on every point I could check. Head is now `049b69273`; your review was against `260593ec3`, so most of what follows is new.

## You were right about the nested restore, and it was worse than "not settled"

I ran it end to end with `lcm`, as you asked. The result on the old code:

```
Packing as: 11.0.0-feature-grammar-json-exp.dirty
Using local libraries: lcm
BUILD EXIT=0
```

…and **all 120** `project.assets.json` files named `11.0.0-beta0178`. Zero named the local package. The build reported success and compiled against nuget.org's package.

Your reasoning was exactly the mechanism: `PackageRestore.targets` restores through `Exec`, global properties do not cross that boundary, and the native build reaches it first. You were also right about why the earlier evidence could not have caught it — SIL.Machine's local pack equalled the pin, so the override was a no-op, and it was simultaneously the only library tested end to end and the only one that could not expose the problem.

Fixed the way you suggested: a generated, gitignored `Build/LocalLibraries.props`, imported by `Build/SilVersions.props` after its defaults. Because it is read from disk, it reaches every MSBuild process including nested `Exec`s. Same build now:

| | before | after |
|---|---|---|
| projects on the local version | **0** | **114** |
| projects on the published version | 120 | 6, all stale from four days earlier, untouched by the run |

And the reverse: an ordinary build removes the overrides, empties the feed, and returns all 114 to `11.0.0-beta0178` without redownloading.

## The naming fix

Adopted, and made content-addressed rather than a fixed suffix: `<core>-<branch>.<commit>`, with the core read from the library's own GitVersion. A stable suffix would still collide with itself — a second pack from a dirty tree would hit the extracted cache and silently serve the previous build — so a dirty tree is marked `.dirty` and always repacked, and the build now names the paths responsible. Your "warn when the packed version equals the pin" is now structurally impossible rather than warned about.

One thing worth flagging, since it bit me: `-p:Version=` alone is not enough. GitVersion.MsBuild assigns `Version` inside a target, which outranks a command-line global property, so the stamp was being silently overridden on the four libraries that use it. Proven on libpalaso, same commit: `SIL.Core.18.0.0-lt22728-vp.1e46149.nupkg` with `-p:DisableGitVersionTask=true`, `SIL.Core.18.0.0-lt22728-vp0033.nupkg` without it.

## Everything else you raised

- **`<clear />`** — dropped, and inherited sources are left alone. Your argument holds and is stronger now: the feed is supplied per-build, so no persistent user-level source is needed either way.
- **Sweeping `$env:LOCAL_NUGET_REPO`** — the feed now defaults to `.localfeed` inside the working tree. That was also what made the automatic cleanup untrustworthy; scoped to one tree, it can only affect the tree that ran it.
- **Cleanup under `-SkipRestore`** — guarded. It no longer removes packages nothing will put back.
- **Version-1 `.nupkg.metadata`** — `source` is now read via `PSObject.Properties` inside the `try`. Tested with a real v1 file under StrictMode: preserved, no throw.
- **A version directory with no metadata** — now removed rather than skipped. It is a partial extraction, and restore replaces it.
- **The test harness never failing properly** — correct, and it had been hiding failures from me all along. `Write-Host -ForegroundColor Red` in the loop, then `exit 1`. Two injected breaks now surface three failures and a count.
- **`$LASTEXITCODE` after `&`** — check removed; the exception was already the real path.
- **The compatibility gate not covering `Build/`** — `powershell-compat.ps1` now scans `Build` as well as `Build/Agent`. Clean under 5.1 and 7.
- **Pack order** — rationale restored at the catalogue and at `$PackOrder`.
- **Comments** — `VersionOutputPath` describes the file; the `.md` pointer is gone from `nuget.config`; the "only the projects FieldWorks uses (avoids native CMake deps)" WHY is back.
- **Print the cleanup path at the end of a local build** — added. You were right that this is the part worth engineering.
- **Five names in four places** — a test now asserts both ValidateSets match the catalogue exactly.

## On the test file

Fair question, and the honest answer is that the source-text regexes are the weaker half — several would fail on a rename that preserved behaviour. I kept them where the thing being asserted *is* a property of the source (that a `$LASTEXITCODE` check is absent, that `SilVersions.props` imports the overrides after its defaults), and leaned on behavioural assertions elsewhere: it now runs the real functions against real fixtures for version derivation, feed resolution, cache discrimination, v1 metadata, partial extractions, and props generation and removal. 70 assertions, of which 25 are source-text. If you would still rather see it smaller, I would rather cut the regexes than the fixtures.

## Scope

I have not shrunk the branch, and I want to be straight about why rather than quietly disagreeing. The nested-restore fix required the generated props file; the props file is what the deferred per-worktree paths needed anyway; and the worktree-local feed is what makes the cleanup you objected to defensible instead of merely convenient. What is left beyond your minimal fix is the setup command, and I would drop that if you want it separate — it is self-contained.

The one thing I did not do is make cleanup fully manual. With the feed inside the working tree and the `-SkipRestore` guard in place, the automatic sweep can no longer reach anything a developer owns outside the repository. If you still want it behind an explicit command, say so and I will move it.

Build/Setup-LocalLibraries.ps1
```diff
@@ -0,0 +1,91 @@
...
+		-Branch $branch -Create
+	$verb = if ($worktree.Created) { 'created' } else { 'already present' }
+	Write-Host "  Branch:   $branch ($verb)" -ForegroundColor Gray
+	Write-Host "  Worktree: $($worktree.Path)" -ForegroundColor Green
+
+	$upstream = Get-FieldWorksLibraryUpstreamState -WorktreePath $worktree.Path
+	if ($upstream.HasUpstream -and ($upstream.Behind -gt 0 -or $upstream.Ahead -gt 0)) {
+		# Reported, not merged: a fetch cannot disturb the working tree but a
+		# merge can, and this worktree may not be the one the caller is editing.
+		Write-Host ("  Upstream: behind {0}, ahead {1} — reconcile it yourself" -f
```
jasonleenaylor: This line has an em dash in it, and that is what is breaking the build.

The file has no byte order mark, so Windows PowerShell 5.1 reads it as ANSI instead of UTF-8. The em dash is three bytes; they arrive as three separate characters and one of them ends the string early. That is where the five parse errors come from -- only the first is real and the rest are knock-on.

I changed it to `--` locally and re-ran `.uild.ps1 -CommentHygiene`: `powershell-compat: clean (5.1 and 7.0)`, and the build then carried on through the native C++ and managed projects. It is the only non-ASCII byte in any PowerShell file in this PR.

There is precedent you will recognise: `6c7de0e93` on #978 is "style: use ASCII punctuation in new prompt/doc/test prose". Same class of problem.

Worth saying that your own change is what caught this. Extending `powershell-compat.ps1` to scan `Build/` is what makes this file visible to the gate -- without it CI would have passed and this would have broken someone's build later.

jasonleenaylor (changes requested):
The build is failing, so I have held the rest of my round-2 comments until it is green. There is one cause and it is a single character -- see the inline note.

**On the setup command: yes please, but leave it out rather than splitting it.** You offered to move it to its own PR. I would rather it did not land at all for now. It is the file that broke the build, and separately I do not want the amount of PowerShell in this repo to keep growing. FieldWorks already carries about 18,000 lines across 75 files. This PR adds 1,238, which is close to a 7 percent increase from one change. Most of that I can justify -- the generated props file and the tests are things I asked for. The setup command is not in that group.

While you are looking at size, `LocalLibraries.psm1` is 547 lines on its own. I am not asking you to split it today, but if there is an obvious seam, it is worth taking.

One thing about the verification. You wrote that powershell-compat was "clean under 5.1 and 7". It is not clean under 5.1 -- that is the failure. My guess is you ran it under 7, where it does pass. Worth running both when the claim says both.

Build/Setup-LocalLibraries.ps1
johnml1135: Gone in `dea00ad52`, and the em dash went with it — you were right about the byte, the
mechanism and the fix.

I confirmed it before deleting: `e2 80 94` at offset 2603, line 70. Every other `.ps1`
and `.psm1` on this branch is zero non-ASCII bytes, so "the only non-ASCII byte in any
PowerShell file in this PR" was literally true.

**It came to 281 lines, not 91.** Deleting the wrapper orphaned four module functions
nothing else calls — `ConvertTo-FieldWorksWorktreeName`, `Resolve-FieldWorksLibraryRepo`,
`Resolve-FieldWorksLibraryWorktree` and `Get-FieldWorksLibraryUpstreamState` — so they
went too. That is your other question answered by the same cut: `LocalLibraries.psm1` is
**547 lines to 405**, and the PR's PowerShell addition drops from 1,238 to about 1,010.
The seam you asked me to look for turned out to be the thing you had already asked me to
remove.

Untouched: the version derivation, the generated props file, the cache handling. Those
are what scopes a local library to one build, and the worktree convenience separated
cleanly because it was never load-bearing for them.

**On the verification claim — you guessed right, and I have stopped inferring.** I had
run it under 7 only. Both checks have now been run under each version separately:

    powershell-compat  5.1.26100.9168  clean (5.1 and 7.0)   exit 0
    powershell-compat  pwsh 7          clean (5.1 and 7.0)   exit 0
    LocalLibraries.Tests.ps1  5.1      passed                exit 0
    LocalLibraries.Tests.ps1  pwsh 7   passed                exit 0

Comment hygiene clean. The body's verification paragraph now carries the retraction
rather than the claim.

One thing I found while editing the body: it contained a literal backspace byte (`0x08`)
where `.\build.ps1` should be, which is why it rendered as `.\uild.ps1` — including in
the passage you quoted back to me. Stripped.

The assertion count is 62 now, not 70. Ready for the rest of round 2 whenever you are.

build.ps1
```diff
@@ -575,6 +585,58 @@ try {
...
+				-LocalRepository $localFeed
+
+			# An earlier selection goes first, so a build naming no library
+			# resolves published versions even in a restore it does not launch.
+			Remove-FieldWorksLocalLibraryProps -Path $localPropsPath
+		}
+
+		$localVersionProperties = [ordered]@{}
+		$localRestoreSourceArg = $null
+		if ($LocalLibraries.Count -gt 0) {
```
jasonleenaylor, line 613: This is the root the rest of the branch hangs from, and I want to put the scope question to you before the detail.

LT-22728's mode 2 dies the moment a local pack stops producing the published version string. In the module that is four functions — `Get-FieldWorksLibrarySourceState`, `ConvertTo-FieldWorksVersionLabel`, `Get-FieldWorksLibraryCoreVersion`, `Get-FieldWorksLocalPackVersion`, about 70 lines — plus `-p:Version` and `-p:DisableGitVersionTask` on the pack call. Dropped into the old flow, which already detected a version and wrote it to `SilVersions.props`, that is the whole repair.

Everything else follows from the decision to drive this from the main build, and the dependencies run one way. `Build/LocalLibraries.props` exists only because this entry point refuses to edit the tracked file — keep the old write and the `Exec` boundary never matters. The 405-line module exists only because `build.ps1` needs the catalogue — keep it in `Manage-LocalLibraries.ps1` and there is no second consumer. Then the `.gitignore` rule, the `SilVersions.props` import and most of the 316-line test file follow from those.

Local library work is deliberate, occasional and developer-driven. It already has a driver in `Manage-LocalLibraries.ps1`, and that script keeps working. Making the main build a second driver puts an uncommon operation into the entry point every developer and every CI job runs.

The argument I can see for it, which the body never actually makes: pack mode no longer touches `SilVersions.props`, so mode 1 — the pin left unreverted, which is what the ticket's repro performs — dies structurally rather than being merely visible. That is a real gain. So the question is not "why did you do this" but "is mode 1 worth this much machinery", and I would like your answer in the description, because right now it presents the entry point as a given rather than arguing for it.

test.ps1
```diff
@@ -125,6 +125,15 @@ if ($LASTEXITCODE -ne 0) {
 	exit $LASTEXITCODE
 }
 
+# Only on a full run: a caller who named a project or filter asked for that.
```
jasonleenaylor: This runs `Build/LocalLibraries.Tests.ps1` — 316 lines, 62 assertions, bespoke harness — on every invocation without `-TestProject` or `-TestFilter`, and exits the whole run non-zero when it fails. So an occasional, deliberate, developer-driven workflow now gates every full test run, for every developer and every CI job, including everyone who never touches a local library.

I asked a milder version of this in round 1, when the file was 133 lines and 25 assertions. It has grown 140% since, which is honest — the assertions follow the surface area. That is rather the point: the test cost is a function of the workflow's size.

If the entry point stays, could this be scoped like `CommentHygiene.Tests.ps1` is — run when `Build/` changed, or behind an explicit switch — rather than gating unrelated runs? If the entry point goes, most of the file goes with it and this is moot.

Build/Manage-LocalLibraries.ps1
```diff
@@ -245,109 +224,108 @@ function Invoke-PackLibrary {
...
+		$script:packedVersions[$cfg.VersionProperty] = $packVersion
+		Copy-LibrarySymbols -LibName $LibName -SourceDir $SourceDir
+		return
+	}
+
+	# Only the library being repacked loses its local packages.
+	Clear-FieldWorksLocalLibraries -PackagesDirectory (Join-Path $repoRoot 'packages') `
+		-LocalRepository $LocalRepo -Libraries @($LibName)
+
+	# Build first: a package may include output from a framework its own project
```
jasonleenaylor: The build and the pack run over the identical `$packTargets` with the same properties, and `dotnet pack` builds unless told not to. For `machine` that is unambiguous: `PackProjects` names two csproj files and both loops are handed exactly those.

For the four libraries whose target is the source directory the comment's scenario is at least conceivable — packing a solution runs `Pack` per project, so output from a non-packable project that a package pulls in by path rather than by `ProjectReference` could be missing where a directory-wide build would have produced it. But that is me constructing the case; nothing here names the library it happened to, and the step is applied to all five regardless.

The cost is small, since MSBuild is incremental. Which library needed this, and does it still, now that the build and pack target lists are the same? If there is a real case it is worth naming in the comment and probably worth applying only to that library.

Build/LocalLibraries.psm1
```diff
@@ -0,0 +1,405 @@
...
+	palaso = @{
+		VersionProperty = 'SilLibPalasoVersion'
+		PdbRelativeDir  = 'output/Debug/net462'
+		CachePrefixes   = @(
+			'sil.core', 'sil.windows', 'sil.dblbundle', 'sil.writingsystems',
+			'sil.dictionary', 'sil.lift', 'sil.lexicon', 'sil.archiving',
+			'sil.media', 'sil.scripture', 'sil.testutilities'
+		)
+		EnvVar          = 'LIBPALASO_PATH'
+		RepoDirectory   = 'libpalaso'
```
jasonleenaylor: `RepoDirectory` is declared for all five libraries and read by nothing. The only other occurrence on the branch is `LocalLibraries.Tests.ps1:210`, a test asserting its value.

Its consumers were the four worktree helpers — `Resolve-FieldWorksLibraryRepo`, `Resolve-FieldWorksLibraryWorktree`, `ConvertTo-FieldWorksWorktreeName`, `Get-FieldWorksLibraryUpstreamState` — that went with `Setup-LocalLibraries.ps1` in `dea00ad52`. The functions went; the data they read stayed, and so did the assertion pinning it. `VersionProject` one line below is the contrast: live, read at `psm1:165-168`.

Worth removing along with the assertion. A test that asserts a value nothing consumes will persuade the next reader the field matters.

jasonleenaylor (changes requested):
The fix is small and correct, and I am not disputing it. Content-addressing the packed version kills mode 2 for every library, and `-p:DisableGitVersionTask` is a real discovery — GitVersion assigning `Version` inside a target would have made the stamp silently useless. The machine PDB path fix is a genuine bug I had not spotted myself: the base pointed at `bin/Debug/netstandard2.0` while Machine writes to `src/<project>/bin/Debug/...`, so that copy had been doing nothing for the one library this ticket is about, and doing it silently. The diagnostic naming where it looked is the right answer to that.

My difficulty is proportionality. Four functions and two pack properties — about 70 lines — repair the bug. The branch is +1078/-306, and the remainder is a new way to drive local library builds from the main build. That is a workflow change rather than a bug fix, and it is what I am asking you to justify rather than assume.

A version I would take without further argument:

1. Version derivation as you have written it, `.dirty` and all.
2. `-p:Version` and `-p:DisableGitVersionTask=true` on the pack call.
3. Pack mode keeps writing `SilVersions.props`, as it always has — mode 1 stays visible and revertible.
4. The machine PDB path fix and its diagnostic.
5. `powershell-compat.ps1` scanning `Build/`, which is what caught the em dash.
6. Unit tests for the four derivation functions.

That leaves out the entry point, the 405-line module, the generated props file, the feed relocation and the automatic cleanup — and on my reading it is still a complete fix for mode 2. Local library work is deliberate and occasional; it already has a driver, and that driver keeps working.

If you think mode 1 is worth the rest, make that argument in the description. It is a defensible one and the description does not currently make it. Four questions inline.

build.ps1
johnml1135: You were right, and the fix was to stop arguing with the premise. `SilVersions.props` is written again, exactly as before, and `Build/LocalLibraries.props` is gone along with the `Exec` boundary, the props generation and most of the module. The module is 102 lines now, not 405.

`-LocalLibraries` survives, but only as a wrapper: it runs `Manage-LocalLibraries.ps1` and then builds. The pin, its visibility and its manual reset are untouched, so there is no second way to select a library.

One thing you did not ask for, because it only became possible once the tracked write came back. The feed path now rides in `SilVersions.props` too, as `RestoreAdditionalProjectSources`. `Build/PackageRestore.targets:28` already imports that file, so a nested restore in its own process sees the feed without a machine-level NuGet source being registered. That removes `dotnet nuget add source`, which outlived the checkout that wanted it, and lets the feed default to `.localfeed` in the working tree instead of `LOCAL_NUGET_REPO` (still honoured when set).

On your 70-line figure: it was right and my earlier reply had it wrong. I had counted the package clearing as new, but `Update-VersionAndClearCache` has always cleared `packages/<prefix>*`, so mode 1 needed nothing new. What is left is the four version functions, two pack properties, and the machine PDB path.

Worth recording for whoever revisits this: the dirty pin is not only an undo. It is the reminder that this tree depends on a library version nobody has released, so an auto-revert would erase a release gate rather than tidy up. That is why clearing it stays manual, and it is now stated in `local-library-debugging.md` and in the `-LocalLibraries` help.

test.ps1
johnml1135: Behind a switch, as you suggested, and your comparison was the right one: `CommentHygiene` at `test.ps1:114` is the only precedent in that file, and it is a switch rather than change detection. `-LocalLibraryTests` follows it and is off by default.

The file also shrank rather than moving: 316 lines and 62 assertions down to 121 and 22, because most of it tested the entry-point machinery that is now gone. What remains covers the version stamp itself, which is the part LT-22728 depends on.

Verified both ways, since a guard that does not guard is worse than no guard:

```
test.ps1 ... -ListTests                      -> harness does not run
test.ps1 ... -LocalLibraryTests -ListTests   -> [PASS] Local library version tests
```

Build/Manage-LocalLibraries.ps1
johnml1135: Gone. It was added by this branch, and reverting the pack flow to the tracked-write version removed it.

I went looking for the case the comment claimed before deleting it, because if one existed it wanted naming rather than removing. I could not find one:

- `machine` is the library this ticket is about and the only one with `PackProjects`. Both packed projects are `<TargetFramework>netstandard2.0</TargetFramework>` — single-target, so there is no second framework to miss.
- The other four resolve `$packTargets` to the source directory, and each has a solution at its root (`Palaso.sln`, `L10NSharp.sln`, `LCM.sln`, `Chorus.sln`). Pack on a solution builds the projects anyway.
- Searching all five for a package pulling output by path from another project's `bin`, `output` or `artifacts` returned nothing.

So your reading was right: same targets, same properties, and `dotnet pack` builds unless told not to.

Build/LocalLibraries.psm1
johnml1135: Gone, with the assertion. Reverting the pack flow moved the library configuration back into `Manage-LocalLibraries.ps1`, whose version never had a `RepoDirectory`, so the field and its test went together. `grep -rc RepoDirectory Build/` is zero.

Your reasoning was exactly right, including why it mattered: the four worktree helpers left in `dea00ad52` and the data they read stayed behind, with a test pinning it. `VersionProject` is the live counterpart and is read at `Manage-LocalLibraries.ps1` when deriving the core version.

johnml1135:
Ran the workflow end to end against real checkouts, packing `machine` (explicit `PackProjects`) and `l10nsharp` (packs via its solution) so both code paths were exercised. It found a defect, so the numbers below are worth more than the earlier unit-test-only claim.

**What held.** The pack produced `SIL.Machine.3.9.2-conformance-fieldworks-w.d3b7643` and `L10NSharp.10.0.0-master.ae81ffb` — neither the bare published `3.9.2` nor `10.0.0-beta0014`. Restore then resolved them: `packages\sil.machine\3.9.2-conformance-fieldworks-w.d3b7643\` and `packages\l10nsharp\10.0.0-master.ae81ffb\` both appeared. An uncommitted edit in one checkout stamped `.dirty` instead of a commit. `git checkout Build/SilVersions.props` followed by a plain build restored `packages\sil.machine\3.9.2\` and `packages\l10nsharp\10.0.0-beta0014\`. `LOCAL_NUGET_REPO` still overrides `.localfeed`.

**The defect: l10nsharp had the same wrong PDB directory as machine.** Configured `output/Debug/net462`; l10nsharp builds `net461`, `net48`, `net8.0`, `net8.0-windows` and `netstandard2.0`, so there was no such directory and its symbols had never been copied — silently, exactly like machine. It now names `net48`, which is what FieldWorks targets. I checked the other three rather than assume: `libpalaso` and `chorus` (`output/Debug/net462`) and `liblcm` (`artifacts/Debug/net462`) all exist and are correct.

**Two smaller things the run exposed.**

The build claimed a local package was in play whenever one sat in the feed, even after the pin had been reverted and nothing was wired in — a developer would read that as still building locally. It now reports only when `SilVersions.props` actually names the feed, and says the pin is what keeps the file dirty.

`XmlWriter` drops the trailing newline, so every pack left `\ No newline at end of file` on a tracked file. That predates this branch but now fires on each pack through the feed write, and it is noise on the one diff that is meant to be a signal. Restored, and idempotent across repeated packs.

Not exercised: a pack of all five libraries at once, and anything on a machine without these checkouts.

johnml1135:
Exercised all five libraries at once against real checkouts, which is what the earlier two-library run left open. `.\build.ps1 -LocalLibraries palaso,lcm,chorus,machine,l10nsharp` then built FieldWorks: **Build succeeded, 0 Warning(s), 0 Error(s)**.

| library | stamped as | packages | PDBs |
| --- | --- | --- | --- |
| palaso | `18.0.0-fix-stale-localrepoupdat.0750fb9` | 20 | 27 |
| l10nsharp | `10.0.0-master.ae81ffb` | 4 | 8 |
| lcm | `11.0.0-feature-grammar-json-exp.dirty` | 6 | 14 |
| chorus | `6.0.0-master.dirty` | 6 | 6 |
| machine | `3.9.2-conformance-fieldworks-w.d3b7643` | 2 | 2 |

Three things fell out of the checkouts happening to be in interesting states. palaso's branch is 31 characters, so the label truncated to 24 without leaving a trailing dash. lcm and chorus both had uncommitted work, so both stamped `.dirty` rather than a commit, and said so. l10nsharp copied 8 PDBs where it had copied none before the `net48` correction.

Restore resolved every one: `packages\sil.core\18.0.0-fix-stale-localrepoupdat.0750fb9\`, `sil.lcmodel\...dirty\`, `sil.chorus.libchorus\...dirty\`, `sil.machine\...d3b7643\`, `l10nsharp\...ae81ffb\`. `packages\sil.core\` held only the stamped version at that point, so the published one had been cleared as intended. `git checkout Build/SilVersions.props` plus a plain build brought back `18.0.0-beta0030`, `11.0.0-beta0178`, `6.0.0-beta0065`, `3.9.2` and `10.0.0-beta0014`, and that build also stayed silent about local packages even though the feed files were still on disk — the reporting fix.

**One further defect, found while the pack ran and now fixed.** The commit identifier had no prefix, and SemVer rejects a numeric pre-release identifier with a leading zero. `NuGetVersion.TryParse` confirms it: `3.9.2-main.0123456` returns false while `3.9.2-main.0750fb9` returns true, the latter surviving only because it contains letters. A short sha that is all digits and starts with one is roughly a commit in 270, and it would have failed the pack with a parse error. The identifier now carries git describe's `g` prefix, which exists for this reason; `g0123456`, `g1234567` and `g0000000` all parse. There is a regression assertion for it.

Worth knowing rather than fixing here: `packages\` keeps one directory per local pack indefinitely. That is harmless now, since a stamped version cannot be mistaken for a published one — it was precisely the collision at an identical version that caused this ticket — but the folder does accumulate across months of local work.

jasonleenaylor (approved):

PR #1111, opened 2026-08-26
Use project .fwlayout files for Avalonia persistence

johnml1135:
Hide a field in the Avalonia detail view, move one, or configure its writing
systems, and the change now lands in the project's `.fwlayout` file — the same
file the legacy Lexicon Edit view reads. Do it in either view and the other
one shows it, and it survives moving the project to another machine.

Until now the Avalonia view kept those settings in a `.viewoverride.json` store
of its own, so the two views drifted apart and a moved project lost its
Avalonia customizations. That store is deleted here (20 files, 3,313 lines);
nothing in `Src` references it any more. Existing `.viewoverride.json` files
are ignored, not migrated — the format never shipped in a release.

The question worth your time is not "does it write the file" — tests cover
that — but **whether a layout that WinForms would render now renders the
same**. Composition, fallback, and command targeting were all rebuilt against
the legacy `XCore.Inventory`, and the contract they must meet is written down
in `Docs/architecture/avalonia-fwlayout-parity.md`.

### Where to look

- **`InventoryViewDefinitionSource.cs`** — the first Avalonia code that writes
  into a live project `.fwlayout`. Mirrors `DataTree.EnsureCustomFields`.
- **`LayoutResolutionWalk.cs`** — WinForms fallback order, including the quirk
  that the `default` search starts at the concrete class's *base*.
- **`RecordEditView.Avalonia.cs`** — largest rewrite; holds the override
  retirement and the fail-closed command targeting.
- **`XmlLayoutImporter.cs`** — where matching WinForms makes the new view
  render *less* than the previous Avalonia code did.
- **`DetailControls/{DataTree,Slice}.cs`** — the only WinForms production
  files touched; additive virtuals, no existing path changes.

### Deliberately not here

- Two behaviors were carved onto their own branches: `avalonia-viewdef-compile-cache`
  (compile-cache rewrite — unmeasured, may be dropped) and
  `avalonia-ws-alternatives-with-data` (fix for LT-22777).
- Three more have review branches without code changes:
  `avalonia-importer-winforms-strictness`, `avalonia-layout-choice-fail-closed`,
  `avalonia-custom-field-placeholder-persistence`.
- Two accepted divergences are recorded in the parity doc; the register is
  otherwise empty.

### Verification

xWorks 1,657/1,659 and FwAvalonia 637/638 pass locally (remainder skipped);
comment hygiene, token hygiene and gitlint clean. No manual FLEx run against a
real project — the parity claims rest on tests and on reading the WinForms
sources.

---

<details>

The design contract for this work is durable and lives in the tree, at
`Docs/architecture/avalonia-fwlayout-parity.md`. It carries the four-field
layout identity, the WinForms behaviors being matched, the acceptance
criteria, and the `Divergences` register — the list of places Avalonia is
allowed to differ. If you are changing detail-view layout behavior, read it
first: anything not in that register that differs from WinForms is a defect.

What is *not* in the tree, and lives only here, is why the scope is what it
is: what was tried, what was cut, and what was left in on purpose.

</details>

<details>

```
project .fwlayout + shipped XML
        ↓  XCore.Inventory              (shared with WinForms, authoritative)
        ↓  InventoryViewDefinitionSource (clones to an immutable snapshot)
        ↓  ViewDefinitionCompiler        (content-fingerprint cache)
        ↓  DetailComposer                (walks, threads layout identity)
        ↓  Avalonia DataTree             (rows)

Avalonia command → existing xWorks writer → project .fwlayout → recompose
```

`Inventory` stays the only thing that loads, merges, and persists layout XML.
FwAvalonia never depends on XCore and never holds a mutable inventory node;
the snapshot is the seam.

Layout identity is the four attributes `Inventory` itself keys on — `class`,
`type`, `name`, `choiceGuid` — plus a caller path locating the exact composed
occurrence. A command targets an already-resolved layout; it never re-runs
fallback.

</details>

<details>

**Both shipped parts directories are loaded, hand-authored first.**
`Inventory` loads `DistFiles/Parts` and then
`Language Explorer/Configuration/Parts`, letting the later files replace
same-id entries. `LayoutSourceLoader` is first-wins, so the same precedence
is expressed by listing the hand-authored directory first. This matters more
than it looks: `CmObject-Detail-HeavySummary`, the `autoCustom` part, and
every generated `default` layout live **only** in `DistFiles/Parts`. Reading
one directory silently empties every sense subtree.

**Omit what WinForms omits.** An unresolved part ref, or part content
`DataTree.ProcessSubpartNode` does not recognize, produces a diagnostic and no
row — because that is what the legacy view does. Rendering a visible
"unsupported" placeholder there would itself be a divergence. Constructs
WinForms *does* render but Avalonia cannot yet support still show as
unsupported rows.

**Fail closed on command targets.** A persistent command resolves to one
exact hidden WinForms slice by object, field, class, layout, and caller path.
Zero matches or several: the command is disabled and logged. It never falls
back to a nearby occurrence to make itself succeed.

**One walk, not two.** The fallback algorithm is generic over the live
inventory node and the shipped-file element, so the two resolution paths
cannot drift.

</details>

<details>

**Migrating `.viewoverride.json`.** Rejected. The format is pre-alpha, hidden
behind a flag, and never shipped in a release. Conversion code would have to
be maintained and eventually deleted anyway; the files are simply inert now.

**Keeping both stores in sync.** Rejected as the thing that caused the bug.

**A committed canonical-JSON snapshot as the load path.** Removed. A snapshot
baked at build time cannot reflect a project file a user just edited, which is
the entire point of this work.

**Bounding the compile cache here.** Moved to `avalonia-viewdef-compile-cache`.
The capacity bound, `Lazy`-based deduplication, memoized parts hash, and
instrumentation counters were precautionary: nothing measured said the previous
cache was a problem, and the counters were test-only instrumentation shipping
in production code. The per-compose memo *did* stay — without it a single
compose rebuilt a snapshot per sense.

**Keeping unselected writing systems that hold data.** Moved to
`avalonia-ws-alternatives-with-data`. It is a display rule in its own right,
not layout parity, and LT-22777 already reports it as a user-facing bug.

</details>

<details>

- It does not claim Avalonia implements every WinForms editor or layout
  construct. Gaps stay visible as unsupported rows with named TODOs; they are
  defects to fix, not approved divergences.
- It does not make `.fwlayout` the permanent format. The parity doc's
  `Retirement` section describes replacing it once no supported WinForms path
  depends on it — that migration is unwritten and this branch is not its
  precedent.
- It does not settle whether the three review branches
  (`avalonia-importer-winforms-strictness`,
  `avalonia-layout-choice-fail-closed`,
  `avalonia-custom-field-placeholder-persistence`) describe behavior that
  should stay. They exist so each can be judged alone; all three are WinForms
  behavior being matched, and `DataTree.EnsureCustomFields` has written the
  custom-field placeholder ref to project files for years.

</details>

<details>

**Fallback order** — `LayoutResolutionWalk.Resolve` was compared line by line
against `DataTree.GetTemplateForObjLayout`: requested name up the class chain;
at `CmObject` the name resets to `default` and the class resets to the concrete
class, but the walk *advances to that class's base before the next lookup*, so
the concrete class's own default is never checked. Covered by
`InventoryViewDefinitionSourceTests`.

**The override store is gone** — `git grep` for `ViewOverride`,
`viewoverride`, and `ViewDefinitionOverride` across `Src` returns nothing.

**Custom-field placeholder** — the ref is set on the live `XmlNode` the
`Inventory` handed out, then persisted through `PersistOverrideElement`,
matching `EnsureCustomFields`. Generated `ref="Custom"` siblings are not
persisted; WinForms regenerates them per load.

**Notebook record types** — a missing `RnGenericRec` choice layout is cloned
from the no-choice layout, tagged, added to the inventory, and written to the
project file, asserted against a real `.fwlayout` write.

**Show-all** — the transient reveal ends when another slice becomes current
and cannot survive a record or Type change; both paths have tests, including
the case where no menu can be shown at all.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1111)


mark-sil:
What is the reason PR's 1097 and 1108 could not have gone in before this PR? Would it have significantly changed this PR if those were in first? Please explain why the expectation is that those PR's rebase on this PR.

mark-sil (commented):
What is the reason PR's 1097 and 1108 could not have gone in before this PR? Would it have significantly changed this PR if those were in first? Please explain why the expectation is that those PR's rebase on this PR.

> @mark-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1111#-:-P02Xrh762809eVW2IBC:b-okfsog)* status: 0 of 36 files reviewed, all discussions resolved.

johnml1135:
We will just have a migration from one to the other.

PR #1114, opened 2026-08-31
Fix for LT-22771

AndyBlack:
When producing the lexical input file for the XAmple parser, circumfixes that contained an interfix did not produce the correct entry information.  The allomorph field (\a) was missing and the type field (\eType) did not contain the key term "nterfix".
This fixes that by correctly checking for interfixes and passing in the correct value to the DoInflAffix named template.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1114)

### DIFF
--- a/Src/LexText/ParserCore/ParserCoreTests/M3ToXAmpleTransformerTestsDataFiles/InterfixingCircumfixeslex.txt
+++ b/Src/LexText/ParserCore/ParserCoreTests/M3ToXAmpleTransformerTestsDataFiles/InterfixingCircumfixeslex.txt
@@ -1,0 +1,77 @@
+
+\lx 5146
+\g 3m.IMPV
+\c W/W
+\o -2 -2 
+\wc 3468
+\a y {6451}
+\eType infix nterfix
+\loc prefix root /#[10409]_  | /#[V]_ 
+		
+\mcc 5146 +/ _ ... 5146 +/ 5146 ... _
+
+
+\lx 5146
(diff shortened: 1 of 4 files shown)


Src/LexText/ParserCore/ParserCoreTests/M3ToXAmpleTransformerTestsDataFiles/InterfixingCircumfixeslex.txt
```diff
@@ -0,0 +1,77 @@
+
+\lx 5146
+\g 3m.IMPV
+\c W/W
+\o -2 -2 
+\wc 3468
+\a y {6451}
+\eType infix nterfix
```
papeh, line 8: *I*nterfix? (throughout this file)

papeh (approved):

Src/LexText/ParserCore/ParserCoreTests/M3ToXAmpleTransformerTestsDataFiles/InterfixingCircumfixeslex.txt
AndyBlack: Yes, it is correct to use nterfix.  See https://github.com/sillsdev/CarlaLegacy/blob/master/pc-parse/doc/dictfile.txi line 992.  (X)Ample assumes anything starting with an 'i' is an infix,

PR #1119, opened 2026-09-02
LT-22691: Add the transient Show-all writing-system reveal

mark-sil:
"Show all right now" on a detail row's Writing Systems menu did nothing. This makes it work: the click reveals every writing system the field can show, for as long as you stay on the record.

The reveal is transient by design — it is view state, not configuration. It survives clicking between fields and editing within the record, and it ends when you navigate to another record or persist a new selection through the writing-system toggles or the Configure dialog. Nothing is written to `.viewoverride.json`, so a peek can never silently pin a field's writing systems.

One reveal covers the whole part, matching what the legacy view does: revealing Gloss on sense 1 reveals it on every sense.

### Why record-scoped expiry

The legacy slice's comments say the reveal ends when the slice loses currency, and its `SetCurrentState` does call `ReloadWssToDisplayForPart` to that end — but that method only resets when `WritingSystemsSelectedForDisplay` is null, and its getter never returns null. The revert has been dead code for years. Testing the shipped behaviour confirmed it: a reveal survives moving between fields and ends only when the rows are rebuilt. This implementation matches the measured behaviour rather than the comment.

Refresh (F5) is a deliberate divergence: the reveal survives it. Legacy rebuilds its slices there, but its F5 handling is itself unreliable in this area (it can leave a block of blank space where a row was), so it is not a parity target. Clearing on every recompose would also kill the reveal whenever an in-record edit completes.

### Also in this branch

`RefreshAvaloniaDetail` now resolves the record to show through the same root walk `ShowRecord` uses. Without it, a `showDescendantInRoot` tool (Notebook) recomposed the subrecord instead of the root on every refresh — a pre-existing defect that also made the reveal clear itself the instant it was set.

The override write path clears the reveal and recomposes only after the save succeeds. `MutateOverrideAndRefresh` is split into a `TryMutateOverride` that reports success without recomposing. Previously a failed save ended the reveal invisibly, so the row collapsed at the next unrelated refresh with nothing to explain it.

### Validation

`build.ps1 -CommentHygiene` green; 23/23 targeted tests pass across the three affected fixtures with no skips (4 new tests). Manually tested in Sena 3: reveal, part scope, record-navigation expiry, replacement by a persisted selection, and confirmation that nothing reaches the override file.

<details>
<summary>Found while testing, not fixed here</summary>

**LT-22777 — writing systems holding data are hidden.** The legacy view never hides an alternative that contains data, regardless of the per-field selection or the project-level checkbox (`SkipEmptyWritingSystem`), and its option list includes active-but-unticked writing systems. The Avalonia composer has neither rule: it resolves only current writing systems and intersects them with the per-field restriction. Two consequences — restricting a field can hide data, and data in an unticked writing system is unreachable. This also bounds the feature in this PR: on a field with no restriction, "Show all right now" has nothing to add, because the composer never resolved the wider set. Filed separately because the fix is a composer-level change to the writing-system model, not to this command.

**Move Field is flaky.** Move Up/Down works sometimes and not others. A real cause was identified — a row at the top level of its own layout gets a null `ParentStableId`, which enablement ignores but `ApplyMoveField` bails on, so the item enables and silently does nothing, even though the applier already reorders roots under an empty parent key. A one-line fix and a test were written and then backed out of this branch, because intermittency means that cause is incomplete. To be investigated on its own.

**Deferred with reasons.** The reveal is left in place on the copy-bail paths (unreadable or stale adapter slice): nothing was persisted there, so keeping the view as the user left it is more honest than collapsing the row. A tidy-up of eight stale "gear-menu" comments is left for a separate commit to keep this diff on topic.

</details>

<details>

# Code Review Summary

**Branch**: LT-22691g

**Base**: origin/main (merge-base 8f46a80fc)

**Date**: 2026-09-02

**Files changed**: 5

## Overview

Makes the Avalonia detail view's "Show all right now" writing-system command work.
It was deliberately inert since the c4b menu work (#1108): the menu item existed but
nothing consulted it. The command now marks the clicked row's part in a host-held set
of template StableIds, and the composer skips that part's per-field
`visibleWritingSystems` restriction, so every row sharing the part shows its full
writing-system set until the shown record changes.

Expiry semantics were chosen against measured legacy behavior rather than the legacy
code comments: the legacy slice's documented "expires when the slice loses currency"
path is dead code (`ReloadWssToDisplayForPart` only resets when
`WritingSystemsSelectedForDisplay` is null, which its getter never returns), so legacy
reveals actually survive until the slices are rebuilt. The reveal is therefore
record-scoped, never persisted.

## Contract/API Changes

`DetailComposer.Compose` (both overloads) gains an optional trailing
`ISet<string> showAllWritingSystemsFields` parameter; all existing callers are
source-compatible. `MutateOverrideAndRefresh` is split, adding a private
`TryMutateOverride` that saves without recomposing and reports success.

None.

- [x] **Reveal was row-scoped where the legacy reveal covers the whole part**
  _(fixed during review: keyed the set on the template StableId, which is the same
  equivalence class as the legacy part-ref; added a test proving one reveal covers
  every sense's Gloss row)_
- [x] **A persist cleared only the clicked row's reveal while the override write
  applies to every row sharing the template** _(fixed during review: same template-id
  rekey)_
- [x] **`RefreshAvaloniaDetail` recomposed `Clerk.CurrentObject` without the
  `showDescendantInRoot` root walk, so the reveal self-cleared on click in Notebook**
  _(fixed during review: added `ResolveShownRecord`, shared with `ShowRecord`; this
  also fixes a pre-existing subrecord-recompose defect)_
- [x] **A failed override save cleared the reveal without recomposing, so the row
  collapsed at the next unrelated refresh** _(fixed during review: `TryMutateOverride`
  split; the reveal is cleared and the view recomposed only after a successful save)_

### Minor - Consider

- [x] **Reveal membership was computed at the call site and threaded as a positional
  bool** _(fixed during review: the check moved inside `ResolveTextRowWritingSystems`)_
- [x] **Repeated invoke-Show-all sequence and a duplicated row locator in the host
  fixture** _(fixed during review: `InvokeShowAllRightNow` and `FindLexemeFormRow`
  helpers)_
- [ ] ~~Lone `<param>` tag on an eight-parameter method breaks the comment standard's
  all-or-nothing rule~~ _(author does not accept the all-or-nothing rule; only the new
  parameter is documented)_
- [ ] ~~The reveal survives F5 / master refresh where legacy rebuilds slices~~
  _(author decision: legacy's F5 handling is itself defective here, leaving blank space
  where a removed row was, so it is not a parity target; clearing on every recompose
  would also kill the reveal when an in-record edit completes)_
- [ ] ~~Reveal state can drift on copy-bail paths (unreadable or stale adapter slice)~~
  _(nothing was persisted on those paths, so keeping the reveal leaves the view as the
  user left it; clearing it would imply a configuration change that did not happen)_

- `./build.ps1 -CommentHygiene` - green (a `-Clean` run was needed once to regenerate
  MIDL headers after PR #1065 added `IVwPattern2`; unrelated to this branch).
- `./test.ps1 -CommentHygiene -TestProject xWorksTests -TestFilter
  "FullyQualifiedName~FieldTypeComposerTests|FullyQualifiedName~DetailObjectCommandExecutionTests|FullyQualifiedName~DetailComposerOverrideTests"`
  - 23/23 passed, 0 skipped.
- Author manual testing in Sena 3 against the checklist below: reveal, part scope,
  record-navigation expiry, persistence replacement, never-persisted.
- Not run: full suite; native tests (no native change); installer validation (none).

## Positive Observations

- The composer bypass is placed where restrictions are applied, so a restriction from
  the shipped layout is honored the same way as one from the project override.
- The reveal is never written to `.viewoverride.json`, verified by a test that asserts
  the stored override still holds the restriction after a reveal.

## Interview Notes

- Expiry semantics: the author tested legacy directly and established that a reveal is
  NOT reverted by moving between fields, contradicting both the legacy doc comments and
  the initial analysis. Record-navigation expiry was chosen to match measured behavior.
- The author established that legacy always displays a writing system that holds data
  regardless of the per-field selection or the project-level checkbox
  (`SkipEmptyWritingSystem`). The Avalonia composer has no such rule and resolves only
  current writing systems. Filed as LT-22777; out of scope here.
- Move Field was found to be flaky during manual testing. A confirmed root-level cause
  was identified and a one-line fix plus test were written, then backed out at the
  author's request because the intermittency proves the cause is incomplete. To be
  investigated separately.
- Author does not understand: nothing recorded. The author reviewed the diff directly
  and drove the design decisions on expiry and scope.

## In-Review Quality Check

All fixes above were made before the commit; build and targeted tests were re-run green
after each. Comment hygiene is clean, and comments comparing behavior to legacy were
removed at the author's request so they do not rot when legacy is retired.

## Suggested Review Focus

- [ ] Template-StableId keying: is the template id the right equivalence class for
      "the same part" in every layout, not just the lexicon ones tested?
- [ ] `ResolveShownRecord` now runs on every Avalonia refresh, changing behavior for
      `showDescendantInRoot` tools (Notebook) that are not yet Avalonia-enabled.
- [ ] The reveal set lives on the host and is passed into every compose; confirm no
      other host composes these rows and would need the same wiring.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1119)

### DIFF
--- a/Src/xWorks/Avalonia/Hosting/RecordEditView.Avalonia.cs
+++ b/Src/xWorks/Avalonia/Hosting/RecordEditView.Avalonia.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2003-2017 SIL International
+﻿// Copyright (c) 2003-2017 SIL International
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
@@ -73,6 +73,13 @@ namespace SIL.FieldWorks.XWorks
 		// projects/windows for the app lifetime.
 		private readonly Dictionary<string, bool> m_expansionStates = new Dictionary<string, bool>();
 
+		// The transient "Show all right now" reveal: template StableIds of parts whose rows
+		// show every writing-system option. Cleared when the shown record changes; never
+		// persisted.
+		private readonly HashSet<string> m_showAllWsFields = new HashSet<string>(StringComparer.Ordinal);
+		// The record m_showAllWsFields belongs to; a different record expires the reveal.
+		private int m_showAllWsRecordHvo;
+
 		private bool ShouldUseAvaloniaLexiconEdit
 		{
 			get { return m_activeUIFramework == UIFramework.Avalonia; }
@@ -142,6 +149,10 @@ namespace SIL.FieldWorks.XWorks
 			SettleDetailEdits();
 			m_detailEditContext.Clear();
 			m_avaloniaEntryForm?.Dispose();
+			// Transient view state dies with the view: rebuilding the host must not
+			// resurrect a stale reveal.
+			m_showAllWsFields.Clear();
+			m_showAllWsRecordHvo = 0;
 			// Null the host + refresh controller after disposing them. The recreation guards
 			// (EnsureAvaloniaEntryFormInitialized / EnsureAvaloniaRefreshController) key on `== null`, so a
 			// runtime flip New->Legacy->New rebuilds a fresh entry form instead of re-showing a disposed one.
@@ -297,6 +308,12 @@ namespace SIL.FieldWorks.XWorks
 			// cancel-on-displace remains the safety net.
 			SettleDetailEdits();
 
+			// Record-navigation expiry: the transient reveal belongs to ONE record -- it
+			// survives focus changes within the record and dies when the record changes.
+			if (obj == null || obj.Hvo != m_showAllWsRecordHvo)
+				m_showAllWsFields.Clear();
+			m_showAllWsRecordHvo = obj?.Hvo ?? 0;
+
 			// Adapter hygiene: the hidden command-routing DataTree must never answer mediator
 			// commands for a PREVIOUS record -- reset it whenever the shown record changes; the
 			// next
@@ -333,14 +350,16 @@ namespace SIL.FieldWorks.XWorks
 			{
 				composed = lexEntry != null
 					? DetailComposer.Compose(lexEntry, Cache, showHidden,
-						overrides: ResolveViewOverride)
+						overrides: ResolveViewOverride,
+						showAllWritingSystemsFields: m_showAllWsFields)
 					// Non-entry roots compose against the tool's configured layout
 					// (m_layoutName, default "Normal"); a type-selected layout (m_layoutChoiceField, e.g.
 					// Notebook RnGenericRec keyed on "Type") resolves to the right variant inside Compose.
 					: DetailComposer.Compose(obj, Cache,
 						string.IsNullOrEmpty(m_layoutName) ? "Normal" : m_layoutName, showHidden,
 						overrides: ResolveViewOverride,
-						layoutChoiceField: m_layoutChoiceField);
+						layoutChoiceField: m_layoutChoiceField,
+						showAllWritingSystemsFields: m_showAllWsFields);
 				if (composed != null)
 				{
 					detail = composed.Model;
@@ -584,6 +603,10 @@ namespace SIL.FieldWorks.XWorks
 			// even when locating fails.
 			var registry = new OverrideCommandRegistry();
 			registry.Add(IsWritingSystemVisibilityChoice, (c, d) => WritingSystemItem(c, d, field));
+			// Show all right now never dispatches or persists: it only marks the row for the
+			// host's transient reveal.
+			registry.Add("CmdDataTree-WritingSystemMenu-ShowAllRightNow",
+				(c, d) => ShowAllWritingSystemsItem(d, field));
 
 			var templateId = ViewDefinitionOverrideEditor.StripRuntimeSuffix(field.StableId);
 			// Locate the clicked node in the field's OWN compiled model (with any current override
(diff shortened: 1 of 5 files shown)


Src/xWorks/Avalonia/Hosting/RecordEditView.Avalonia.cs
```diff
@@ -781,22 +824,31 @@ private void ApplyMoveField(DetailField field, ViewNodeLocation location, bool u
...
 				if (store == null)
-					return;
+					return false;
 
 				var existing = store.TryGet(field.ClassName, field.LayoutName)
 					?? new ViewDefinitionOverride(field.ClassName, field.LayoutName, "detail", null, null);
 				var merged = ViewDefinitionOverrideEditor.MergeOperation(existing, op);
 				store.Save(merged);
-				RefreshAvaloniaDetail();
+				return true;
```
papeh, line 846: Throws from RefreshAvaloniaDetail are no longer caught, but the other three callers aren't catching, and Devin thinks it's safe.

papeh (approved):

PR #1122, opened 2026-09-03
Attempt at fixing LT-22708

AndyBlack:
Change-Id: I24735bad6cd9e8823a0a028151afdc64722d5a8e
Using XAmple with additional properties can cause a crash per what is in https://jira.sil.org/browse/LT-22708.
This fix has the XAmple parser send a refresh message just after the new list is added.  It does cause the Try a Word window to flash and then disappear while the main window also flashes but returns.

I do not know what is causing the Try a Word window to disappear although I suspect it may have to do with the parser running in a separate thread...

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1122)

### DIFF
--- a/Src/FwParatextLexiconPlugin/FdoLexicon.cs
+++ b/Src/FwParatextLexiconPlugin/FdoLexicon.cs
@@ -2,6 +2,18 @@
 // This software is licensed under the LGPL, version 2.1 or later
 // (http://www.gnu.org/licenses/lgpl-2.1.html)
 
+using Paratext.LexicalContracts;
+using SIL.FieldWorks.WordWorks.Parser;
+using SIL.LCModel;
+using SIL.LCModel.Core.KernelInterfaces;
+using SIL.LCModel.Core.Text;
+using SIL.LCModel.DomainImpl;
+using SIL.LCModel.DomainServices;
+using SIL.LCModel.Infrastructure;
+using SIL.LCModel.Utils;
+using SIL.Machine.Morphology;
+using SIL.ObjectModel;
+using SIL.PlatformUtilities;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -11,18 +23,7 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Web;
-using Paratext.LexicalContracts;
-using SIL.LCModel.Core.Text;
-using SIL.LCModel.Core.KernelInterfaces;
-using SIL.LCModel;
-using SIL.LCModel.DomainImpl;
-using SIL.LCModel.DomainServices;
-using SIL.LCModel.Infrastructure;
-using SIL.LCModel.Utils;
-using SIL.FieldWorks.WordWorks.Parser;
-using SIL.Machine.Morphology;
-using SIL.ObjectModel;
-using SIL.PlatformUtilities;
+using XCore;
 using WordAnalysis = Paratext.LexicalContracts.WordAnalysis;
 
 namespace SIL.FieldWorks.ParatextLexiconPlugin
@@ -39,7 +40,8 @@ namespace SIL.FieldWorks.ParatextLexiconPlugin
 		private readonly int m_defaultVernWs;
 		private PoorMansStemmer<string, char> m_stemmer;
 		private readonly string m_projectId;
-
+		Mediator Mediator { get; set; }
+		PropertyTable PropertyTable { get; set; }
 		internal FdoLexicon(string scrTextName, string projectId, LcmCache cache, int defaultVernWs)
 		{
 			m_scrTextName = scrTextName;
@@ -79,6 +81,10 @@ namespace SIL.FieldWorks.ParatextLexiconPlugin
 			{
 				m_parser.Dispose();
 				m_parser = null;
+				Mediator.Dispose();
+				Mediator = null;
+				PropertyTable.Dispose();
+				PropertyTable = null;
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


Src/LexText/ParserCore/XAmplePropertiesPreparer.cs
```diff
@@ -57,13 +58,33 @@ where fd.IsCustomField
 					select fd).ToList();
 		}
 
+		public Boolean ListsAlreadyAdded()
+		{
+			var possListRepository = Cache.ServiceLocator.GetInstance<ICmPossibilityListRepository>();
+			var customList = possListRepository.AllInstances().FirstOrDefault(list => list.Name.BestAnalysisAlternative.Text == customListName);
```
papeh: `return possListRepository.AllInstances().Any(....` would look cleaner

Src/LexText/ParserCore/XAmplePropertiesPreparer.cs
```diff
@@ -57,13 +58,33 @@ where fd.IsCustomField
...
+			if (customList != null)
+			{
+				return true;
+			}
+			return false;
+		}
+
+		public Boolean XAmplePropertiesCustomFieldAlreadyAdded(string fieldName, int fieldClassId)
```
papeh, line 66: Did you mean to filter by `fieldClassId`, too?

Src/FwParatextLexiconPlugin/FdoLexicon.cs
```diff
@@ -646,7 +652,11 @@ private void InstantiateParser()
 			switch (m_cache.LanguageProject.MorphologicalDataOA.ActiveParser)
 			{
 				case "XAmple":
-					m_parser = new XAmpleParser(m_cache, parserDataDir);
+					// LT-22708 As of 2026.08.26, this is only called within FLEx from FdoLexiconTests.
```
papeh: how expensive would it be to keep this test code out of production code?

Src/LexText/ParserCore/XAmpleParser.cs
```diff
@@ -58,7 +62,11 @@ private void InitXAmpleAddonDataInfo()
 					{
 						xampleAddonFileRoot = doc.Root;
 						var preparer = new XAmplePropertiesPreparer(m_cache, xampleAddonFileRoot, false);
-						preparer.AddListsAndFields();
+						if (!preparer.ListsAlreadyAdded())
+						{
+							preparer.AddListsAndFields();
+							FwUtils.Publisher.Publish(new PublisherParameterObject(EventConstants.ReloadAreaTools, "lists", m_propTable.GetWindow()));
```
papeh, line 68: [Devin](https://app.devin.ai/review/sillsdev/FieldWorks/pull/1122#bug-BUG_pr-review-job-699428d345ad4738b2678942ccfe48b7_0003) says that this closes the Try a Word dialog by replacing the active main window.

AndyBlack: Ah, that makes sense.  I think we should just live with this since using this XAmple properties approach is relatively rare and it only happens the very first time.

Src/LexText/ParserCore/XAmplePropertiesPreparer.cs
AndyBlack: Thanks!

Src/FwParatextLexiconPlugin/FdoLexicon.cs
AndyBlack: @jasonleenaylor Could you reply to this, please?  I suspect that this code is used between Paratext and FLEx...

Src/LexText/ParserCore/XAmplePropertiesPreparer.cs
AndyBlack: Yes, I did.  Thanks.

Src/FwParatextLexiconPlugin/FdoLexicon.cs
jasonleenaylor: I agree that this needs a second look, it might be better to keep parser as null in this file rather than instantiating it, I don't think the parser is being exercicised by the paratext plugin code.

jasonleenaylor: Concrete reason to prefer that: `DisposeManagedResources` guards on `m_parser != null` but then disposes `Mediator` and `PropertyTable`, which are only created in the `XAmple` branch of `InstantiateParser`. An HC project therefore throws a `NullReferenceException` on dispose. Not instantiating the parser here avoids it entirely.

papeh (changes requested):

Src/FwParatextLexiconPlugin/FdoLexicon.cs
AndyBlack: OK.  I made the m_parser always be null with a comment that this is for now.

Src/FwParatextLexiconPlugin/FdoLexicon.cs
```diff
@@ -642,18 +642,19 @@ private ILexEntry GetMatchingEntryFromParser(string wordForm)
 
 		private void InstantiateParser()
```
papeh: It looks like this method can be removed

Src/FwParatextLexiconPlugin/FwParatextLexiconPluginTests/FdoLexiconTests.cs
```diff
@@ -556,10 +556,11 @@ public void FindClosestMatchingLexeme()
 			Assert.That(matchingLexeme.LexicalForm == "a", Is.True);
 
 			// Found by parser
-			lexeme = m_lexicon.CreateLexeme(LexemeType.Prefix, "pre");
-			m_lexicon.AddLexeme(lexeme);
-			matchingLexeme = m_lexicon.FindClosestMatchingLexeme("prea");
-			Assert.That(matchingLexeme.LexicalForm == "a", Is.True);
+			// LT-22708 2026.09.11 Jason Naylor says to keep m_parser as null for now.
```
papeh: commented code can be removed to keep the code clean (it will still be available in git history)

papeh (changes requested):

Src/FwParatextLexiconPlugin/FdoLexicon.cs
AndyBlack: OK

Src/FwParatextLexiconPlugin/FwParatextLexiconPluginTests/FdoLexiconTests.cs
AndyBlack: OK

papeh (approved):

PR #1123, opened 2026-09-03
Fix LT-22760

AndyBlack:
Change-Id: Ia24a6d7fc28fa18d6d8983e9f2fa44aedb89216a

The fix is to use the BuildCreateNewOpenCancelDialog whenever an operations file cannot be found.  Before we'd also check on the filename to see if it was null or empty before using this dialog.  There's no need for that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1123)

### DIFF
--- a/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs
+++ b/Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs
@@ -862,44 +862,32 @@ namespace SIL.AllomorphGenerator
 		{
 			if (!File.Exists(OperationsFile))
 			{
-				if (String.IsNullOrEmpty(OperationsFile))
+				// probably first time it is run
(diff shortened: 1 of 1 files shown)


Src/Utilities/AlloVarGen/AllomorphGeneratorDll/AlloGenFormBase.cs
```diff
@@ -862,44 +862,32 @@ protected void LoadMigrateGetOperations()
...
-					MessageBox.Show(
-						AllomorphGeneratorDll_Strings.ksOperationsFileNotFound,
-						AllomorphGeneratorDll_Strings.ksLoadError,
-						MessageBoxButtons.OK,
-						MessageBoxIcon.Error
-					);
+					// Assume it was canceled, so quit
+					this.Dispose();
 				}
-				return;
```
papeh, line 902: unintentional whitespace change

papeh (approved):

PR #1127, opened 2026-09-08
Run as many tests as possible in CI

papeh:
to catch test failures before merging PRs. Previously, these tests had been saved for CD builds to save time, but those failures are too easy to ignore.

https://github.com/sillsdev/FieldWorks/actions/workflows/CI.yml completed in 14:09 for this PR, which is the third-shortest successful time in the last page runs (range 12:33-16:14), so we are saving next to no time by excluding "long-running" tests from CI builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1127)

PR #1129, opened 2026-09-09
Make VwPattern::ReplacementText fail explicitly with no current match

johnml1135:
`IVwPattern.ReplacementText` now fails with `E_UNEXPECTED` whenever the pattern has no current match, instead of sometimes returning the untouched source string. The test that caught this, `ReplaceAllIn_LeavesTerminalNoMatchState`, was order-dependent: it passed in some runs and failed in the CI run for #1127.

**What a reviewer will ask: does this change live find/replace?** No. `ReplaceAllWithAlgorithm` deliberately parks the pattern at a -1 found range with the text source still set, and `get_ReplacementText` had no rule for that state. It handed -1 to `VwSimpleTxtSrc::GetSubString`, whose `ReplaceTsString` calls are not error-checked, so the property succeeded or failed depending on what those calls did with a bad range. Both managed callers that read `ReplacementText` in production do so after a successful `FindIn`, so the guard cannot fire there: the bulk-replace fallback in `BulkEditBar.cs`, and the interactive Find/Replace flow in `FwFindReplaceDlg.cs`. The review question worth your time is whether `E_UNEXPECTED` is the right HRESULT for "no current match".

**Where to look**

- `Src/views/VwPattern.cpp`: the guard checks the text source and both ends of the found range before anything is derived.
- `Src/views/Views.idh`: the contract now names the three states that have no match.
- `VwPatternReplacementTests.cs`: two new tests pin the before-first-search and failed-FindIn states; the existing terminal-state test pins ReplaceAllIn.

**Deliberately not here**

- `VwSimpleTxtSrc::GetSubString` still discards the HRESULTs from `ReplaceTsString`. Wrapping them in `CheckHr` would change live find/replace behaviour and is out of scope.
- A `TrivialTextSrc::GetSubString` hardening that was in the working tree was dropped: the class has no callers in the repo.

**Verification**

Built with `build.ps1 -CommentHygiene -TokenHygiene`. `VwPatternReplacementTests` 36/36 standalone, the mode that failed in CI. Full `xWorksTests` assembly 1603 passed, 0 failed. Native `TestViews` [309-0-0], which exercises `get_ReplacementText` through `TestVwPattern.h`. Full suite once: 6067 tests, 6005 passed, 62 skipped, 0 failed, exit 0.

---

<details>

This record holds the diagnosis that led to a seven-line guard. Nothing was deleted from the tree for this PR; the reasoning simply never lived in a file. It is here because the guard looks like defensive noise until you know that the property's failure used to be an accident of an unchecked HRESULT three calls deeper.

</details>

<details>

- **Guard in `get_ReplacementText`, not in `GetSubString`.** The property is the contract surface; the text source is shared by live editing. Making the text source strict would change behaviour for callers that never touch replacement.
- **`E_UNEXPECTED`, not `E_FAIL`.** The state is reachable only by calling the property out of sequence, which is what `E_UNEXPECTED` means in the Views COM code elsewhere.
- **Test the two other entry states.** The terminal-state test existed but relied on incidental failure. Adding the before-first-search and failed-FindIn cases means any future change to the guard must say which of the three states it means to alter.

</details>

<details>

- **`CheckHr` around `ReplaceTsString` in `VwSimpleTxtSrc::GetSubString`.** Correct in isolation, but it turns a silent no-op into a thrown error on a path live find/replace uses. That deserves its own change with its own tests.
- **Hardening `TrivialTextSrc::GetSubString`.** It was in the working tree at one point. The class is only reachable through `CoCreateInstance` and nothing in the repo creates it, so the change would have been untested code.

</details>

<details>

- Terminal state: `Src/views/VwPattern.cpp` sets `m_ichMinFoundLog = m_ichLimFoundLog = -1` and keeps `m_qtsWhereFound` at the end of `ReplaceAllWithAlgorithm`.
- Unchecked HRESULTs: the two `ReplaceTsString` calls in `VwSimpleTxtSrc::GetSubString` in `Src/views/VwTxtSrc.cpp` are not wrapped in `CheckHr`.
- Production callers: `BulkEditBar.cs:5175` reads `ReplacementText` only after `FindIn` reports `ichMin >= 0`; `FwFindReplaceDlg.cs` (1263, 1823, 1861) reads it only on paths gated by `IsReplacePossible`/`m_fFound`, both of which require a successful find.
- Order dependence: the fixture passed inside the full `xWorksTests` run and failed standalone before the guard; both pass after it.

</details>

<details>

# Code Review Summary

**Branch**: fix/vwpattern-replaceall-state

**Base**: origin/main

**Date**: 2026-09-09

**Review model**: Claude Fable 5.1

**Files changed**: 3

## Overview

`IVwPattern.ReplacementText` had no defined behaviour when the pattern held no
current match. `ReplaceAllWithAlgorithm` leaves the pattern in exactly that
state on purpose, and the test asserting a failure there passed or failed
depending on run order because the failure came from an unchecked HRESULT in
`VwSimpleTxtSrc::GetSubString`, not from a contract. The branch adds an explicit
`E_UNEXPECTED` guard, documents the contract in Views.idh, and pins the other
two no-match states with tests.

Review ran as a two-axis pass (repository standards, and the branch against the
requirements stated in-session) with two Sonnet reviewers, followed by a
best-practice check with two more.

## Contract/API Changes

- `IVwPattern.ReplacementText` now returns `E_UNEXPECTED` before the first
  search, after a failed `FindIn`, and after `ReplaceAllIn`. Previously the
  result in those states was undefined.

None.

None.

### Minor - Consider

- [ ] ~~`VwSimpleTxtSrc::GetSubString` discards the `ReplaceTsString` HRESULTs~~
  _(out of scope: wrapping them in CheckHr changes live find/replace behaviour;
  the guard makes the bad range unreachable for this property)_

- `build.ps1 -CommentHygiene -TokenHygiene`: clean.
- `VwPatternReplacementTests` standalone (the mode that failed in CI): 36/36.
- Full `xWorksTests` assembly: 1603 passed, 0 failed.
- Native `TestViews`: [309-0-0].
- Full suite once: 6067 tests, 6005 passed, 62 skipped, 0 failed.

## Positive Observations

- Comments in Views.idh and VwPattern.cpp are accurate WHAT/WHY and within the
  200-character budget.
- The new tests assert the HRESULT value, not only that a COMException occurs.

## Interview Notes

- Author confirmed the fix and asked for it to ship separately from the
  test-infrastructure work so the product change is not held by infra debate.
- An unrelated `TrivialTextSrc::GetSubString` hardening was dropped from the
  working tree because the class has no callers in the repo.

## Suggested Review Focus

- [ ] Is `E_UNEXPECTED` the right HRESULT for "no current match"?
- [ ] Confirm no production caller reads `ReplacementText` before a successful
  `FindIn`.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1129)


Src/views/VwPattern.cpp
```diff
@@ -373,11 +373,18 @@ bool IsEmpty(ITsString * ptss)
...
+	@error E_UNEXPECTED if there is no current match. A pattern that has not searched yet,
+	one whose last FindIn failed, and one that has just completed a ReplaceAllIn are all in
+	that state.
 ----------------------------------------------------------------------------------------------*/
 STDMETHODIMP VwPattern::get_ReplacementText(ITsString ** pptssText)
 {
 	BEGIN_COM_METHOD;
 	ChkComOutPtr(pptssText);
+	// The answer is derived from the matched range of the text source the match was found in,
+	// so with no current match there is nothing to derive it from.
```
papeh, line 385: I don't think this comment is necessary. If it must be kept, a more concise version could be "If there is no current match, there can be no replacement text"

Src/xWorks/xWorksTests/Search/VwPatternReplacementTests.cs
```diff
@@ -472,6 +477,33 @@ public void ReplaceAllIn_LeavesTerminalNoMatchState()
...
+		[Test]
+		public void ReplacementText_AfterUnsuccessfulFind_Throws()
+		{
+			var pattern = MakePattern("missing");
+			pattern.ReplaceWith = TsStringUtils.MakeString("new", Ws);
+			var source = MakeTextSource("old old", out var length);
+			Find(pattern, source, 0, length, true, out var ichMin, out _);
+			AssertNotFound(ichMin, "pattern with no occurrence in the source");
+
 			Assert.Throws<COMException>(() => { var unused = pattern.ReplacementText; });
```
papeh: Should this also check for EUnexpected?

papeh (commented):
Looks good overall. Two questions.

Src/views/VwPattern.cpp
johnml1135: Taken verbatim:

```cpp
// If there is no current match, there can be no replacement text.
```

Yours is better and I have kept it as written. Mine described how the answer is produced, which the three conditions on the next line already show; yours states the contract, which is the part the code cannot say for itself. It also fits the repo's WHAT/WHY rule, where mine was narrating the HOW.

Src/xWorks/xWorksTests/Search/VwPatternReplacementTests.cs
johnml1135: Yes, it should have, and thank you for catching it — this was the one place the point of the change had not actually been applied.

The reason it matters is the reason the guard exists. Before it, reading `ReplacementText` in a no-match state threw a `COMException` *by accident*: a -1 range reached `VwSimpleTxtSrc::GetSubString`, whose `ReplaceTsString` results are not error-checked, so the failure was incidental rather than contractual. A bare `Assert.Throws<COMException>` therefore passes with the guard removed, which makes it a test of the symptom rather than the contract.

It now reads the same as its two siblings:

```csharp
var exception = Assert.Throws<COMException>(
    () => { var unused = pattern.ReplacementText; });

Assert.That(exception.HResult, Is.EqualTo(EUnexpected));
```

All three no-match states assert `E_UNEXPECTED` now, and all three pass: `ReplacementText_BeforeFirstSearch_Throws`, `ReplacementText_AfterUnsuccessfulFind_Throws` and `ReplaceAllIn_LeavesTerminalNoMatchState`. That third one is worth noting, because its passing with the HRESULT check confirms the guard answers `E_UNEXPECTED` after a failed `FindIn` too, rather than some other error arriving from deeper down.

The commit message had claimed all three asserted the HRESULT while only two did. Corrected along with the test, rather than left as a claim the code did not support.

papeh (approved):

PR #1133, opened 2026-09-10
LT-22691: Menus for reference-vector rows in the Avalonia detail view

mark-sil:
Reference-vector rows (Publication Settings: Subentries, Referenced Complex Forms, Semantic Domains, Publish In, ...) now have support for the label and edit field menus.

- Items are selectable (to support the move left/right commands).
- Ctrl+click runs the first jump.
- Backspace or Delete removes the focused item.
- Move Left / Move Right on the label menu and the item menu act on the current item.
- Restore the scroll offset, the vector row's current item, and focus after a re-show.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1133)

### DIFF
--- a/Src/Common/FwAvalonia/Detail/DataTree.cs
+++ b/Src/Common/FwAvalonia/Detail/DataTree.cs
@@ -54,6 +54,26 @@ namespace SIL.FieldWorks.Common.FwAvalonia.Detail
 		private readonly Action<string, bool> _expansionChanged;
 		private readonly Action<DetailMenuRequest> _menuRequested;
 		private readonly Action<DetailLinkRequest> _linkRequested;
+		// The vector rows of the shown fields, and the one whose item is current (at most one
+		// per view), so re-show continuity can read and restore that selection cheaply.
+		private readonly List<FwReferenceVectorField> _vectors = new List<FwReferenceVectorField>();
+
+		/// <summary>The reference-vector row that has a current item; null when none
+		/// does.</summary>
+		internal FwReferenceVectorField SelectedVector { get; private set; }
+
+		/// <summary>The reference-vector rows of the shown fields.</summary>
+		internal IReadOnlyList<FwReferenceVectorField> VectorRows => _vectors;
+
+		/// <summary>True when the rebuild that follows should give keyboard focus to the vector
+		/// row's current item.</summary>
+		internal bool VectorFocusRequested { get; private set; }
+
+		/// <summary>
+		/// Asks the rebuild that follows to give keyboard focus to the vector row's current item,
+		/// for a gesture (a menu-driven move) made while no editor had focus.
+		/// </summary>
+		public void RequestVectorFocusOnRebuild() => VectorFocusRequested = true;
 		private readonly IFwClipboard _clipboard;
 
 		// Computed once per view (not per field/row) from the widest WS abbreviation across the
@@ -299,6 +319,8 @@ namespace SIL.FieldWorks.Common.FwAvalonia.Detail
 		private void RebuildItems()
 		{
 			_form.Items.Clear();
+			_vectors.Clear();
+			SelectedVector = null;
 			var visible = DetailVisibility.ComputeVisibility(Model.Fields, GetRecordedExpansion);
 			for (var i = 0; i < Model.Fields.Count; i++)
 			{
@@ -488,12 +510,18 @@ namespace SIL.FieldWorks.Common.FwAvalonia.Detail
 			AutomationProperties.SetAutomationId(labelBlock, automationId + ".Label");
 			AutomationProperties.SetName(labelBlock, field.Label ?? field.Field ?? string.Empty);
 			ToolTip.SetTip(labelBlock, field.Label ?? field.Field); // 11.17: legacy label tooltips
-			// 13.3: the field's slice menu opens from a right-click on the label cell
-			// or from the hover "..." button in the left gutter.
-			var labelCell = WrapWithFieldMenu(labelBlock, field, automationId, out var labelKebab);
-
 			var editor = CreateEditor(field, automationId);
 			editor.Margin = new Thickness(0, 0, 0, FwAvaloniaDensity.FieldSpacing);
+			if (editor is FwReferenceVectorField vector)
+			{
+				_vectors.Add(vector);
+				vector.SelectionChanged += OnVectorSelectionChanged;
+			}
+
+			// 13.3: the field's slice menu opens from the label cell's right-click or the
+			// gutter "..." button; the editor's current item rides each request it raises.
+			var labelCell = WrapWithFieldMenu(labelBlock, field, automationId, out var labelKebab,
+				editor as IDetailItemSelection);
 
 			// Hover-reveal: the WHOLE row (label cell + editor) is the hover/focus
 			// surface for the field-options "..." and any editor affordance (chooser
@@ -507,6 +535,24 @@ namespace SIL.FieldWorks.Common.FwAvalonia.Detail
 			return new FieldContent { Content = editor, Label = labelCell };
 		}
 
+		// One current item per view: selecting in one vector row clears every other row's, so
+		// the item-relative commands always have a single target.
+		private void OnVectorSelectionChanged(object sender, EventArgs e)
+		{
+			if (!(sender is FwReferenceVectorField source))
+				return;
+			if (source.SelectedItemIndex < 0)
+			{
+				if (ReferenceEquals(SelectedVector, source))
+					SelectedVector = null;
+				return;
+			}
+			var previous = SelectedVector;
+			SelectedVector = source;
+			if (previous != null && !ReferenceEquals(previous, source))
+				previous.ClearSelection();
+		}
+
 		// The width of the left gutter holding the per-row field-options "..." button.
 		// Reserved on every row (when a host bridge is present) so labels align whether
 		// or not a row has a menu.
@@ -515,7 +561,7 @@ namespace SIL.FieldWorks.Common.FwAvalonia.Detail
 		// The label cell answers context-menu requests with the row's slice menu; the
 		// kebab opens its own menu or hotlinks. With no host bridge, content is unwrapped.
 		private Control WrapWithFieldMenu(Control inner, DetailField field, string automationId,
-			out Control kebab)
+			out Control kebab, IDetailItemSelection selection = null)
 		{
 			kebab = null;
 			if (_menuRequested == null)
@@ -546,7 +592,7 @@ namespace SIL.FieldWorks.Common.FwAvalonia.Detail
 				{
 					// No pointer position is available here, so the menu drops from the
 					// icon rather than from wherever the mouse sits.
-					_menuRequested(DetailMenuRequest.FromAnchor(button, field, kind));
+					_menuRequested(DetailMenuRequest.FromAnchor(button, field, kind, selection));
 				};
 				rail.Child = button;
 				kebab = button;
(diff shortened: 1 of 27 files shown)


Src/Common/FwAvalonia/Detail/FwFieldControls.cs
```diff
@@ -1262,8 +1273,61 @@ public FwReferenceVectorField(
 					// the right-click or the Remove flyout only opens over ink.
 					Background = FwAvaloniaDensity.TransparentBrush
 				};
-				AutomationProperties.SetAutomationId(text, automationId + ".Item." + item.Key);
+				AutomationProperties.SetAutomationId(text, ItemAutomationId(automationId, item.Key));
+				// Any button selects, so a right-click's menu acts on the item under the pointer;
+				// focus selects too. Items are focusable (a click focuses one) but not tab stops.
+				var itemIndex = index;
+				text.Focusable = true;
+				KeyboardNavigation.SetIsTabStop(text, false);
```
jasonleenaylor, line 1281: Items get `Focusable = true` but `IsTabStop` false, and I don't see arrow-key handling on the row or `Focusable` on the `StackPanel` — so is a mouse click the only way to make an item current? If it is, then Backspace/Delete, the context-menu key and Move Left/Right all need the mouse first, which makes the current keyboard work less valuable.

I could be missing something here, but if Claude and I evaluated it correctly addressing this in a follow up would be good.

Src/xWorks/Avalonia/Hosting/RecordEditView.ReferenceVectorMenus.cs
```diff
@@ -0,0 +1,239 @@
...
+		internal void MoveReferenceItem(DetailField field, string key, bool forward)
+		{
+			// Pending edits settle first, so the move is its own undo step and an invalid
+			// pending edit rolls back with its own warning instead of taking the move with it.
+			m_detailEditContext.Settle();
+			var context = m_detailEditContext.Current;
+			if (context == null || !context.TryMoveReferenceItem(field, key, forward))
+				return;
+			// The menu took keyboard focus; the re-show hands it to the moved item.
+			m_avaloniaEntryForm?.FocusVectorItemOnNextShow();
```
jasonleenaylor, line 240: `FocusVectorItemOnNextShow()` runs unconditionally, but the re-show only happens when the settle below returns clean, and `VectorFocusRequested` (`DataTree.cs:70`) has no reset. `Validate()` is a whole-record rule against current domain state, so on an entry with both the lexeme form and the citation form empty the settle cancels, the move rolls back, no re-show occurs, and the flag stays set on the live `DataTree` — the next re-show for any unrelated reason then pulls focus into this row. This is a bit of an edge case but the consequences would be really annoying. Moving the call inside the `Count == 0` branch should cover it — the re-show can't happen inside `Settle()` itself: the refresh runner is posted through `ScheduleOnUiThread`, and during `Commit()` the session is still open so `_isEditing()` holds it rather than delivering it.

Worth a test if you agree: `DetailObjectCommandExecutionTests` already drives `MoveReferenceItem` over a real cache in `SubentriesMoveRight_ThroughTheHost_ReordersTheRow`, so a sibling that clears both forms first, asserts the order is unchanged, then checks that an unrelated re-show does not pull focus into the row, is about a dozen lines. The existing focus tests cannot catch it — `NewVectorView` builds a fresh `DataTree` per re-show, so the stale-flag case never arises there.

Src/xWorks/Avalonia/Hosting/RecordEditView.ReferenceVectorMenus.cs
```diff
@@ -0,0 +1,239 @@
...
+		}
+
+		// A Move Left / Move Right item for the request's current item, enabled only when the
+		// row can be reordered and the item is not at that end.
+		private DetailMenuItem MoveCommandItem(DetailMenuRequest request, UIItemDisplayProperties display,
+			bool forward)
+		{
+			var field = request.Field;
+			var index = request.SelectedItemIndex;
+			var canMove = field.Kind == DetailFieldKind.ReferenceVector && field.CanReorderItems
```
jasonleenaylor: `IsEditable` is not consulted here, and not further down either — `RegisterReferenceMove` gates only on `canReorder`, so the handler exists and the move would write. Removal is gated on it: the Backspace/Delete handler in `FwFieldControls` sits under `if (editable)`, which is `editContext != null && field.IsEditable`.

Not reachable today — `CanReorderItems` defaults false and is set true only at the four `RegisterReferenceMove` sites, all composed with `isEditable: true` — so feel free to wave this off. An `&& field.IsEditable` would just keep the two gestures honest about the same flag.

jasonleenaylor (changes requested):
One thing to fix, one question and one nit inline. The fix is a line move.

Src/xWorks/Avalonia/Hosting/RecordEditView.ReferenceVectorMenus.cs
mark-sil: Good find. Fixed in 50ff014: the focus request now sits inside the `Count == 0` branch, so a refused move leaves nothing behind. Your reading of the ordering holds up. `AvaloniaDetailRefreshController` checks `_isEditing()` on both the notification and the delivery path and suspends rather than delivering while the session is open.

On the test, I wrote it and could not make it discriminate. A rolled-back move fires its own PropChanged, which re-shows the record and replaces the `DataTree`, so the flag is already gone when the assertion runs. I confirmed that by reverting the fix and re-running: it passes either way. So the symptom is narrower than we both assumed. On the old code focus is pulled into the row on the rollback's own re-show, immediately, rather than waiting for an unrelated one.

There is a second snag. Driving the invalid-record path headlessly pops the rolled-back-edit notification, which blocks an unattended run. I would rather ship no test than one that passes against the bug, but if you can see an assertion that separates the two paths I will add it.

Src/Common/FwAvalonia/Detail/FwFieldControls.cs
mark-sil: Your reading is correct. A mouse click, or a programmatic focus call in tests, is the only way to make an item current today. Items are focusable but not tab stops, the row is not focusable, and there is no arrow-key handling. So Backspace and Delete, the context-menu key, and Move Left and Right all need the mouse first.

Worth adding what testing against the legacy view showed: it does not fully work there either. Arrow keys do move between items, but the first item still needs a mouse click, and tabbing into the vector consumes a tab stop while selecting nothing.

So a follow-up should not just reach parity. Tabbing into the row should make an item current, arrows should move between items, and the current item should survive leaving and re-entering the row. There is also a host-level question, since the WinForms host may not forward Tab into the Avalonia pane at all. That is planned as its own PR.

Src/xWorks/Avalonia/Hosting/RecordEditView.ReferenceVectorMenus.cs
mark-sil: Confirmed not reachable today. All four `RegisterReferenceMove` call sites compose their row with `isEditable: true`, so `CanReorderItems` never turns true on a read-only row.

Taking it rather than waving it off. `MoveCommandItem` now gates on `field.IsEditable` too, which covers execution as well, since the item's action is only wired when it is enabled. Leaving removal gated while the move is not is the kind of asymmetry that becomes a real bug the first time a read-only reorderable row exists.

thejambi:
*[`Src/Common/FwAvalonia/Detail/DataTree.cs` line 518 at r2](https://reviewable.io/reviews/sillsdev/FieldWorks/1133#-P1FOAXI0xp3P9-EeOeq:-P1FOAXI0xp3P9-EeOer:bcbyw6j) ([raw file](https://github.com/sillsdev/FieldWorks/blob/50ff014923bc6990a2d5036128b3a5c5b5a3f9cf/Src/Common/FwAvalonia/Detail/DataTree.cs#L518)):*
> ```Smalltalk
> 			{
> 				_vectors.Add(vector);
> 				vector.SelectionChanged += OnVectorSelectionChanged;
> ```

Looks like there's no corresponding "-=" to unsubscribe from the OnVectorSelectionChanged action in the RebuildItems method.

Src/Common/FwAvalonia/Detail/DataTree.cs
```diff
@@ -488,12 +510,18 @@ private FieldContent AddField(int index, DetailField field)
...
-			// 13.3: the field's slice menu opens from a right-click on the label cell
-			// or from the hover "..." button in the left gutter.
-			var labelCell = WrapWithFieldMenu(labelBlock, field, automationId, out var labelKebab);
-
 			var editor = CreateEditor(field, automationId);
 			editor.Margin = new Thickness(0, 0, 0, FwAvaloniaDensity.FieldSpacing);
+			if (editor is FwReferenceVectorField vector)
+			{
+				_vectors.Add(vector);
+				vector.SelectionChanged += OnVectorSelectionChanged;
```
jasonleenaylor, line 518: @thejambi I dug into this one and I do not think it needs a change to land, though it is a reasonable thing to stop on.

Why it is safe as written: `RebuildItems` builds fresh editors on every pass — `BuildItem` calls `CreateEditor`, which is `SliceFactory.Build(...)`, with no caching or reuse — so each `FwReferenceVectorField` is subscribed exactly once, at construction. There is no path to a double-subscription, which is the failure a missing `-=` normally causes. The discarded vectors cannot fire either: they leave the visual tree so no pointer or focus event reaches them, and nothing retains a reference to call `SelectItem` or `ClearSelection` on them, since `_vectors` is cleared and `SelectedVector` nulled in the same method. And the delegate points from the dead vector at the live tree, not the other way round, so nothing is pinned — the classic leak is a long-lived publisher holding short-lived subscribers, which is the reverse of this.

Where it would belong if you want it anyway, purely so the pairing is visible rather than re-derived by the next reader: in `RebuildItems`, immediately above the existing `_vectors.Clear()`.

```csharp
foreach (var vector in _vectors)
    vector.SelectionChanged -= OnVectorSelectionChanged;
_vectors.Clear();
```

Two caveats if it is picked up. It would only be locally symmetric — it covers vectors dropped by a rebuild, but vectors still live when the whole `DataTree` goes away are not unsubscribed either way, and the class has no disposal hook at all to put that in. And it is not a convention this file already follows: `DataTree.cs` has five `+=` and no `-=`, the other four being lambdas on controls it builds and discards, the same lifetime story as this one. I suspect this line stood out because it is the only one bound to a named method.

jasonleenaylor (approved):
Approving. Both points from the last round are fixed, and I checked each against the code rather than the commit message: the focus request now sits behind the settle, and `MoveCommandItem` gates on `IsEditable` — which does cover execution, not just enablement, since the item's action is only wired when `canMove` holds.

The keyboard-reach question is answered and belongs in its own PR, as you said. Worth repeating your point here so it does not get lost: the follow-up should go past legacy parity rather than match it, since the legacy view also needs a click before the arrows do anything.

One follow-up worth filing, unrelated to anything this PR introduced: `_labelBlocks` gains an entry per row in `BuildItem` and is never cleared, so it grows on every expand/collapse and `ApplyLabelColumnWidth` writes `MaxWidth` to blocks that left the tree long ago. It only writes and never aggregates, so the label width stays correct — it is retained memory rather than a wrong result. Same missing clear is on `main`, so it predates this branch.

Src/xWorks/Avalonia/Hosting/RecordEditView.ReferenceVectorMenus.cs
jasonleenaylor: I went looking for an assertion that separates the two builds and couldn't find one worth having.

The only state that actually differs is `VectorFocusRequested` on the tree that was live at the gesture — and every route to it from `xWorksTests` is reflective. `m_avaloniaEntryForm` and `m_detailEditContext` are private fields on `RecordEditView`, `DetailHostControl.CurrentContent` is protected, and FwAvalonia declares no `InternalsVisibleTo` at all, so the internal flag isn't visible even to a friend assembly. Asserting the rolled-back order alone passes on both builds, which is the trap you already hit.

Doing it without reflection would mean widening those permanently to enable one corner case. I don't think that trade is worth it here, so: agreed, ship it without the test. Not adding one that passes against the bug was the right call.

jasonleenaylor (approved):
> @jasonleenaylor reviewed 27 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1133)* status: all files reviewed, 1 unresolved discussion (waiting on mark-sil).

PR #1134, opened 2026-09-10
Ban the word "legacy" in code comments and commit messages

mark-sil:
The code-comment standard allowed "legacy X" as a behavioral-parity WHY. Once the code it refers to is deleted, the word names nothing a reader can look up, and they cannot tell whether the behavior is still required.

Ban the term in comments and in commit messages, which are both read long after the referent is gone. PR descriptions and review discussion stay exempt: they are read alongside the diff while the context is live, so the team's own word is the clearest one there.

State the scope in AGENTS.md, where it is in context regardless of which skill fires, then repeat the relevant half where each artifact is authored -- the two skills for the ban, pr-pitch for the exemption, so a PR body is not written under a rule that does not apply to it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1134)

### DIFF
--- a/.claude/skills/fieldworks-code-commenting/SKILL.md
+++ b/.claude/skills/fieldworks-code-commenting/SKILL.md
@@ -73,15 +73,27 @@ an agent passes one of them on every run.
 
 ## Legacy references
 
-Only as a behavioral-parity WHY ("matches the legacy X"). No temporal
-framing -- not "the replacement for X", not "until we build Y". Prefer
-symbol names over line numbers, which rot.
+Do not use the word "legacy" in a comment -- not "matches the legacy X",
+not "as legacy does". Legacy is being deleted; once it is gone the word
+names nothing the reader can look up, and they cannot tell whether the
+behavior it describes is still required.
+
+State what the code does in its own terms instead -- "a keyboard gesture
+has no pointer position, so the menu anchors under the field" rather than
+"matches legacy SliceTreeNode, which anchors at the cursor". No temporal
+framing either -- not "the replacement for X", not "until we build Y".
+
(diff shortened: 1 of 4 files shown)


.claude/skills/fieldworks-code-commenting/SKILL.md
```diff
@@ -73,15 +73,27 @@ an agent passes one of them on every run.
...
+
+State what the code does in its own terms instead -- "a keyboard gesture
+has no pointer position, so the menu anchors under the field" rather than
+"matches legacy SliceTreeNode, which anchors at the cursor". No temporal
+framing either -- not "the replacement for X", not "until we build Y".
+
+This covers the assertion messages of new tests too. Commit messages
+follow the same rule (`.claude/skills/commit-messages/SKILL.md`). PR
+descriptions and review discussion are exempt, so do not coin a
+euphemism to avoid the word there.
```
papeh, line 89: I don't think we want to coin a euphemism anywhere. "PR Descriptions and review discussions are exempt from this rule" should suffice.

papeh (dismissed):
LG overall. One change request.

.claude/skills/fieldworks-code-commenting/SKILL.md
mark-sil: I can remove it, but my preference would be to leave it.  Because of the ban in other places, Claude thought it would be best to use the word "classic' instead of "legacy" in a PR I was working on.  Telling Claude not to "coin a
euphemism" was Claude's suggestion as a way to prevent it from doing that.

jasonleenaylor (approved):

PR #1135, opened 2026-09-10
LT-22672: Compose the Allomorphs slice, and the shared detail-view fixes it surfaced

thejambi:
The Allomorphs section now matches legacy's shape in the Avalonia detail view: no doubled
banner, no numbered header above rows that label themselves. **Is Abstract Form** is an editable
checkbox. The **Environments** row keeps its chooser and adds what a chooser cannot do:
typing an environment the project lacks, created and attached in one undoable step.

**Only 3 of the 36 changed files have "Allomorph" in the name.** Converting one slice kept
exposing defects every row in every tool already had. The review worth your time is the shared
changes, not the slice -- hence the order below.

### Where to look

- **Every row in every tool now passes a new check.** Before drawing a field, the composer asks
  the domain whether it applies to that object -- as legacy always did. Widest blast radius here.
- **A checkbox declaring `toggleValue` showed and stored the inverse of its property.** Six of the
  twelve shipped checkbox slices declare it, "Active" among them.
- **The view no longer rebuilds itself mid-click.** Any popup anchored in it holds the refresh
  off; a press released outside no longer strands one.
- **Chooser rows change everywhere.** A pick now commits, so Undo reverts it instead of being
  spent closing it; opening one highlights the value the row holds, not the first item; an empty
  chooser is a full-width target, not a six-pixel sliver.
- **Single-string rows lose a stray writing-system label**, and stop padding the gutter for
  rows that draw one.

### Deliberately not here

- **Per-command hotlinks.** Legacy shows one label per command; this shows a single
  "Field options" button. Sixteen sections across four areas, so it is its own ticket --
  **LT-22672 likely cannot close on this PR alone.**
- Legacy's inline typed environment editor, with its insert helpers and error command.
- Two detail-view-wide defects, written up separately: Ctrl+Z routing, and per-writing-system
  fonts on reference-vector items.
- Two rules have no test that fails without them: infix Positions (the layout hides that row
  first) and the cross-target `<if>` gate.

### Verification

Merged up to current `main`, including #1133, which edits four of the same files; build and
both hygiene gates clean. **xWorksTests `Avalonia`: 1637 passed. FwAvaloniaTests: 744
passed. 0 failed.** The manual pass was walked in the app,
including Delete / Move / Merge / Swap / Convert, which cannot be tested headlessly.

*Sections below: mechanisms, rejected designs, evidence -- for a close or automated pass.*

---

<details>
<summary><b>How to use these sections</b> -- and reading this a year from now</summary>

Everything below is detail behind the summary above: the mechanism for each change, the designs
that were rejected, and the evidence for every claim. Nothing here is needed to review the diff;
it is here for a close pass, for an automated reviewer with the whole tree in context, and for
whoever asks "why is it like this" later.

The analysis, design report, integration-test plan and manual-test checklist for this
conversion live under `Docs/migration/working/`, which is **gitignored**. They were never
merged and are not in the tree. What follows is their reasoning, synthesized.

Two things are worth knowing before changing this code. Much of the detail view's behaviour
here is invisible to the headless suites -- click sequencing, popup lifetime, rendering that
depends on a real template -- so a green run is not evidence on its own. And every behaviour
claim below was checked by falsification: suppress the behaviour, confirm the specific test
fails, restore. Tests in this area pass for the wrong reason easily; see Evidence.

</details>

<details>
<summary><b>Decisions, and why</b> -- Allomorphs slice behaviour</summary>

**The Environments row is a chooser AND create-capable.** Legacy's `PhoneEnvReferenceLauncher`
is a `ReferenceLauncher`: a `SimpleListChooser` behind the button *and* an editable
`PhoneEnvReferenceView` inline. Not one or the other. The row therefore composes as an ordinary
reference vector -- the chooser half, which already worked -- and its edit context additionally
offers `IReferenceItemCreation`. Create-on-type lives in the picker's existing filter box: type
something that matches nothing and a create row appears.

**The create row trails the matches.** Leading, it would take the default highlight, so Enter on
a partial query like `Sk` would mint a new `Sk` instead of picking `Sky`.

**Matching strips spaces**, as `ConnectToRealCache` does, so `/ # _` and `/#_` resolve to one
environment rather than creating a second.

**A malformed environment is created, not refused.** Legacy attaches the `PhEnvironment`
regardless of `CheckConstraints` and annotates it; refusing would discard what the user typed.
The annotation reads its message from `CheckConstraints` -- legacy's own oracle -- so the two
views cannot disagree about what is invalid. Composing must never write, and
`createAnnotation: false` is not on its own enough to promise that: of the three classes that
override `CheckConstraints`, only `PhEnvironment` honours the flag. The other two are adhoc
prohibitions, which the model gives owning collections only, so they never reach a reference
vector row -- that, not the flag, is why this stays a pure read.

**Created environments carry the vernacular writing system**, as `PhoneEnvReferenceView` does
(`m_wsVern`). An environment string is phonology in the vernacular script.

**Booleans stage through the existing `TrySetOption`** as "true"/"false" rather than widening
`IDetailEditContext`, and a toggle commits immediately, because legacy's `CheckBoxSlice` writes
it as it lands.

**Approved divergence -- the add picker is add-only.** Current members ride as unavailable keys,
so they list greyed rather than pre-checked, and removal is the per-item right-click. Legacy's
`SimpleListChooser` pre-checks members and treats the dialog as setting the whole membership.
Kept because "+" means add and the row owns removal. Recorded in the `FwReferenceVectorField`
class doc with its approver.

**Approved divergence -- synthesized chooser links.** The detail view synthesizes the
"Edit the ... list" jump for any possibility-list row the layout left linkless and puts it on
the row; legacy synthesizes only for `autoCustom` and puts it inside the chooser dialog.
Recorded on `CreateChooserLinks`.

</details>

<details>
<summary><b>What reaches past this slice</b> -- and why it had to</summary>

**Field relevance.** The composer never consulted `ICmObject.IsFieldRelevant`, so it showed rows
the domain says do not apply. Found because legacy drew Stem Allomorph Label on one allomorph of
four and the New UI on all four; the four differed only by morph type. The gate sits in `Walk`,
the single entry every node passes, so it covers the whole surface rather than the row that
exposed it. Relevance is **not** a hidden field -- Show Hidden Fields does not reveal an
irrelevant row, and the tests assert that with the flag on.

**Rebuild timing.** `LcmDetailEditSession.Commit` clears `IsOpen` *before* `EndUndoTask` raises
`PropChanged`. The refresh controller's "is the user editing?" guard therefore reads false, and
the commit's own notification recomposed the view while a click was still in flight -- the
release then reached a control that had been rebuilt away. A click in flight now counts as busy
for that guard, and so does a popup the view opened, because those flyouts anchor to a control
inside the view.

**Popups report themselves.** Rather than threading a callback to each flyout owner, a popup
resolves the view by walking up from its own target -- a threaded callback would reach only the
owners it is threaded to, and most of these live in classes that never receive one. The wiring
happens where a popup is BUILT, not at each owner, so a new site cannot forget it:
`FwOptionChooser.CreateOptionFlyout` is the single construction path for every option flyout in
the app, and `DetailMenuFlyout.Build` for every host-built menu. The remaining three are
hand-made in `FwFieldControls` and wired there. No teardown is needed: Avalonia closes a flyout
when its anchor leaves the visual tree, which a test pins.

**Choosers commit on the pick**, like a toggle and a vector add/remove. Staging alone left the
fenced session open, so the user's next Undo was spent by the session guard settling the pending
edit rather than reverting it. The chooser was the only interactive row kind not given
`context.Save`.

**Choosers.** The open list highlights the current value and scrolls it into view; typing still
highlights the first match, because typing is a search. `InsertEntryDlgViewModel` had hand-rolled
the same behaviour by poking `OptionsList.SelectedIndex`; it now calls the shared `SelectByKey`.

**The writing-system gutter** now draws only on multistring rows, matching legacy, which labels
each alternative of a `MultiStringSlice` and leaves a `StringSlice`'s single value unlabelled.
The one shared gutter width is measured from those rows alone; before, a plain row's
abbreviation padded a column it was not in (60px became 84.9px in the test that pins it).

**Checkbox slices honour `toggleValue`.** A slice declaring it shows and stores the LOGICAL
INVERSE of its property, on read and write, as legacy's `CheckBoxSlice` does -- "Active" is
stored as Disabled. The composer ignored the attribute. Six of the twelve shipped
checkbox-editor slices declare it: `Final`, `HideMinorEntry`, and the four `Disabled`/"Active"
rows. `IsAbstract` is not among them, so the Allomorphs section is unaffected.

**The relevance gate matches legacy's placement**: checked before the node kind is dispatched,
as `DataTree.ProcessSubpartNode` checks it before its own switch, so a conditional whose test
field is irrelevant drops its subtree in both. The one carve-out legacy has is now matched too:
no gate for an `<if>` aimed at another object, whose field belongs to that object's class.

</details>

<details>
<summary><b>Paths not taken</b> -- rejected designs, so they are not re-proposed</summary>

**A dedicated environments control replacing the chooser.** An owned control with typed lines,
per-row validation and find-or-create is a plausible reading of legacy, and it is wrong: legacy
is a `ReferenceLauncher`, so replacing the chooser removes half of what it does. The route here
extends the row that already shipped. A working implementation of the typed half is on the local
branch `wip/FwEnvironmentField` if the inline editor is picked up later; its validation and
space-stripped matching are already reused here.

**Holding only the `EditCompleted` re-show** does not work: `PropChanged` schedules a rebuild
independently, and `Commit` clears `IsOpen` before raising it, so the notification arrives with
the guard already false.

**Releasing the hold at the pointer release** does not work either: a click that opens a picker
is not finished when the button goes up, and rebuilding then dismisses the picker.

**Docking the gear to the right edge** is defensible parity -- legacy puts its `...` there -- but
`CreateConfigureGear` returns null unless the row resolves an editor target, so most chooser rows
have no gear and would stretch into empty space regardless.

**Filling every chooser's cell**, not just an empty one: a chooser with a value then lights the
whole row on hover, past both the value and the gear.

**`UndoLimit = 0`** to stop the TextBox eating Ctrl+Z assumes Avalonia then leaves the gesture
unhandled. If it does not, Ctrl+Z does nothing at all, which is worse than the current
behaviour.

</details>

<details>

**RootSite-embedded views are still unconverted.** GAP-ROOTSITE-VIEWS stands. This branch did
not replace `PhoneEnvReferenceView`; it added create-on-type to a chooser row. Legacy's inline
typed editor -- with `OnInsertSlash`, `InsertEnvironmentBar`, `InsertNaturalClass`,
`InsertOptionalItem`, `InsertHashMark` and `OnShowEnvironmentError` -- has no equivalent here.
Do not cite this branch as precedent for "RootSite views are convertible".

**`MoAffixProcess`** is deferred, not dropped: it degrades to a labelled Unsupported row and
needs the `AffixRuleFormulaSlice` plugin. **`MsEnvFeatures`** remains an Unsupported row.

</details>

<details>

- **Per-command hotlinks (P15/P16).** `CreateHotlinkStrip` renders one "Field options" button;
  legacy paints one label per command with per-command enabled state. Blocked on nothing
  technical, but it is shared UI across 16 declarations in four areas, and per-command
  enablement comes from xCore display handlers that only evaluate against a laid-out, visible
  DataTree -- so it needs a manual pass per area, not a headless test.
- **Semantic snapshot (P4)** for the Allomorphs subtree -- independent, unstarted.
- **Ctrl+Z routing** and **per-writing-system fonts on reference-vector items** -- both have
  written notes; both are detail-view-wide, not Allomorphs.
- **`MoAffixAllomorph.Position` relevance** has no test that bites. Its row is excluded by the
  layout's `choice`/`where guidequals` filter before `IsFieldRelevant` is consulted, so the
  domain rule is unexercised. The test kept for the negative case says so in its own doc.

</details>

<details>

**Suites.** `xWorksTests -TestFilter Avalonia`: 1637 passed, 2 skipped. `FwAvaloniaTests`: 744
passed, 1 skipped. `build.ps1 -CommentHygiene -TokenHygiene` clean. Run after merging current
`main` in.

One pre-existing teardown error shows in the xWorks run and is **not from this branch**:
`LexiconFirstSliceEditContextEdgeCaseTests` throws `SynchronizationLockException` disposing the
cache, because `Validate_WhitespaceOnlyLexeme_IsAnError` stages an edit and never commits or
cancels it, leaving the unit-of-work write lock held. `git blame` puts it in LT-22625 (#964);
no commit here or in #1133 touches that file. No test fails.

**Relevance rule coverage.** Five classes override `IsFieldRelevant`.
`MoStemAllomorph.StemName`, `MoAffixForm.InflectionClasses` and `MoStemMsa.InflectionClass` are
pinned by tests verified to fail with the gate removed; the latter two compose the same fixture
twice with one precondition changed, so the row is shown to appear as well as to vanish. `MoAffixAllomorph.Position` is
double-gated and its domain rule unexercised (above). `VirtualOrdering` is not shown in a detail
view, so there is nothing to gate.

**Falsification.** Each behaviour is pinned by a test confirmed to fail when the behaviour is
suppressed. The measurements: an unfilled chooser cell is **6.0px** wide; filling every cell
gives **300px in a 300px cell**; without the relevance gate, 4 environment tests fail with
"Expected: ReferenceVector, But was: Text".

**Traps in this test environment**, each of which produced a test that passed either way until
it was checked. `FakeDetailEditContext` reports `IsOpen` only until its first commit, so a
second autosave in one test never fires. `PointerCaptureLost` is a Direct routed event, so a
Bubble registration for it never runs and does not route up from a descendant. An untemplated
Avalonia control still satisfies `Bounds > 0` and has visual children, so a render assertion
must name something only a real template produces -- the box `Border`, for instance.

**Not run:** `-Native` (no native changes) and installer validation (no WiX changes).

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1135)


Src/Common/FwAvalonia/Detail/FwOptionChooser.cs
```diff
@@ -622,6 +657,94 @@ private void RememberSeen(IReadOnlyList<DetailChoiceOption> options)
...
+		{
+			_createOption = null;
+			_createText = string.Empty;
+			if (!_allowCreate)
+				return results;
+
+			var typed = (query ?? string.Empty).Trim();
+			if (typed.Length == 0)
+				return results;
+			if (results.Any(o => string.Equals(o.Name, typed, StringComparison.CurrentCultureIgnoreCase)))
```
mark-sil: The picker's matcher and the domain's matcher disagree about spaces, so the create row misfires whenever the user's spacing differs from the stored spacing.

`ApplyFilter` (L773-776) matches the literal query as a substring, so typing `/ _ #` does not match a stored `/_#` and the existing environment is filtered out of the list. This guard then finds nothing to suppress against, and the only row offered reads `Create "/ _ #"`. `FindOrCreateEnvironment` (`DetailComposer.cs:1713`) compares through `StripSpaces`, so it resolves what the picker could not -- the two ends are using different rules on the same text.

Two outcomes:

- **Environment exists in the project, not on this field.** It attaches the existing one, which is right, but the row promised to create it.
- **Environment already on this field.** `ReferenceCreate` returns false at `DetailComposer.cs:1681`, and `FwFieldControls.cs:1552-1556` has no else branch -- the flyout closes and nothing is said.

Repro: an allomorph already carrying `/_#`. Open "+", type `/ _ #`, Enter. Only `Create "/ _ #"` is offered; taking it closes the picker and changes nothing.

Typed with matching spacing, everything behaves correctly -- the member lists greyed and no create row appears. So the design is right and only the matcher is out of step. `StripSpaces` is private to `DetailComposer`; the cheapest fix is probably to have the create row ask the context whether the text already resolves, rather than teaching the picker about environment normalization.

thejambi: Confirmed, and thank you — this is a better analysis than mine. I'd looked at the silent no-op during pre-PR review and rated it "consistent with how the add path rejects duplicates"; I missed that the same mismatch hides the existing environment from the list, which is the real defect.

Fixed in `12015bd46`. I went a slightly different route than asking the context: the row now publishes the comparison its domain uses, and the picker runs both the filter and the create-row guard through it. Asking the context would suppress the create row but leave the existing option still filtered out, so the user would get an empty picker where today they at least get a row that works. Normalizing both ends fixes the visibility and the guard together.

Typing `/ _ #` now lists `/_#` — greyed when it's already on the field, since members ride as unavailable keys — and offers no create row. Rows that mint nothing from typed text publish no comparison and keep literal matching, which has its own test. Verified by making the comparison the identity: the positive tests fail with `Expected: some item equal to "/_#" But was: < "Create "/ _ #"" >`.

mark-sil (approved):
> @mark-sil reviewed 36 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1135)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on thejambi).

Src/xWorks/Avalonia/Composer/DetailComposer.cs
```diff
@@ -529,9 +535,46 @@ private ViewDefinitionModel CompileForObjectWithOverrides(ICmObject obj, string
 
 			private bool HideWhenEmpty(ViewNode node) => node.Visibility == ViewVisibility.IfData && !_showHidden;
 
+			/// <summary>
+			/// Whether the DOMAIN says this field does not apply to this object, which legacy
+			/// asks before building a slice (SliceFilter -> ICmObject.IsFieldRelevant). StemName
```
jasonleenaylor, line 540: `SliceFilter.IncludeSlice` has two gates, not one: it looks the slice's `id` up in the tool's filter list and hides the row if it is listed, and only then asks `IsFieldRelevant`. This adds the second. Worth saying so here, since the comment currently reads as the whole contract.

Not for this PR. The `id` never reaches the composer -- it is not in `XmlLayoutImporter.HandledSliceAttributes` and `ViewNode` has no `Id` -- so the filter half needs an importer change first. Filed as LT-22802 with the one live case I found: Grammar > Production Restrictions sets `filterPath=...\basicPlusFilter.xml`, whose five `CmPossibility*` entries are real slice ids in `CmPossibilityParts.xml`, so the new UI shows Status, Discussion, Confidence, Researchers and Restrictions where legacy hides them. Categories and Compound Rules carry the same `filterPath` but are unaffected: different layouts.

Src/xWorks/xWorksTests/Avalonia/Composer/DetailFieldRelevanceTests.cs
```diff
@@ -0,0 +1,181 @@
...
+using SIL.LCModel.Core.Text;
+using SIL.LCModel.Infrastructure;
+
+namespace SIL.FieldWorks.XWorks
+{
+	/// <summary>
+	/// The composer asks the DOMAIN whether a field applies to an object before emitting a row,
+	/// which legacy does through SliceFilter -> ICmObject.IsFieldRelevant.
+	///
+	/// Five classes override it in liblcm. MoStemAllomorph.StemName is covered by
```
jasonleenaylor, line 17: `MoStemMsa` overrides this for a second field, and that one has no test here. From the IL in `SIL.LCModel.dll`: flid 5001007 (`FromPartsOfSpeech`, labelled "Attaches to Categories") returns false unless `OwningLexEntryHasProCliticOrEnclitic`. `Morphology.fwlayout:39` declares that part `visibility="always"`, so before this change the row composed on every stem MSA; now it disappears for every entry without a clitic. That is a larger visible change than the StemName case, and nothing pins it.

(The same override's `InflectionClass` limb also returns false when `PartOfSpeechRA` is null, which the compound-rule test does not reach either.)

It is the shape of the test above it, with one precondition:

```csharp
[Test]
public void Compose_FromPartsOfSpeech_OnlyForAnEntryWithAClitic()
{
    ILexEntry entry = null;
    IMoStemMsa msa = null;
    NonUndoableUnitOfWorkHelper.Do(Cache.ActionHandlerAccessor, () =>
    {
        entry = Cache.ServiceLocator.GetInstance<ILexEntryFactory>().Create();
        msa = Cache.ServiceLocator.GetInstance<IMoStemMsaFactory>().Create();
        entry.MorphoSyntaxAnalysesOC.Add(msa);
    });

    var fields = DetailComposer.Compose(entry, Cache, showHiddenFields: true).Model.Fields;
    Assert.That(HasRow(fields, "FromPartsOfSpeech", msa.Hvo), Is.False,
        "no clitic on the entry, so the domain says the row does not apply");

    // Positive control: add a clitic allomorph, change nothing else.
    NonUndoableUnitOfWorkHelper.Do(Cache.ActionHandlerAccessor, () =>
    {
        var clitic = Cache.ServiceLocator.GetInstance<IMoStemAllomorphFactory>().Create();
        entry.AlternateFormsOS.Add(clitic);
        clitic.Form.set_String(Cache.DefaultVernWs,
            TsStringUtils.MakeString("clitico", Cache.DefaultVernWs));
        clitic.MorphTypeRA = MorphTypes.GetObject(MoMorphTypeTags.kguidMorphProclitic);
    });

    fields = DetailComposer.Compose(entry, Cache, showHiddenFields: true).Model.Fields;
    Assert.That(HasRow(fields, "FromPartsOfSpeech", msa.Hvo), Is.True,
        "the same row on the same object composes once the entry has a clitic");
}
```

I have not run it.

PR #1143, opened 2026-09-16
LT-22691: Answer the reorder-vector menu natively in the Avalonia detail view

mark-sil:
**Draft.** Opened so the mechanism can be looked at alongside the proposal to retire the hidden DataTree command adapter one menu id at a time. This is the first id.

**What changes**

The Avalonia detail view still builds a hidden WinForms DataTree so xCore menu commands can be answered by its slices and DTMenuHandler through the mediator. This PR gives one menu id, mnuReorderVector (Move Left, Move Right, Alphabetical Order on reference-vector rows), a native authority that owns display and execution, so the mediator and the hidden tree play no part in it.

- `IDetailMenuAuthority`: answers every leaf under the menu ids it owns. A hidden leaf is null; a leaf it does not know throws, so an owned id can never leak an item as visible-but-disabled.
- `XCoreMenuBridge` consults the authority before the mediator display query, so owned leaves cost no Display round trip. `OwnsAll` lets the host skip building the adapter when every id in a request is owned. Today the label menu always merges mnuDataTree-Object, so that skip is inert until the shared menu is native.
- `ReorderVectorMenuAuthority` answers from the row: Move Left and Right from `CanReorderItems` and the request's current item, Alphabetical Order from the new `DetailField.CanResetItemOrder` (the layout's `reorder="true"`). Alphabetical Order executes natively through the new `IDetailEditContext.TryResetReferenceOrder`, backed by `VirtualOrderingServices.ResetVO`.
- The label-menu interceptor entries for the Move commands are gone; the item menu keeps them, sharing one item builder.

**Tests** (xWorksTests, 34 pass)

- Equivalence: the label menu rendered through the adapter and through the authority is textually identical, for the Subentries row with no current item and for the read-only Complex Forms row.
- Without the adapter, the mediator renders the three items greyed out because no handler answers them; the authority renders them correctly.
- A real reference sequence offers the moves but no Alphabetical Order.
- Alphabetical Order through the host discards the virtual ordering as one undo step.
- Contract: every leaf under mnuReorderVector in the shipped configuration is one the authority answers and carries a configuration node.

**Known and deferred** (recorded in the retirement notes, to be decided with the next menu ids)

- List-populated leaves have no configuration node, so ownership by XML ancestry cannot claim them. Bites at mnuDataTree-MultiStringSlice.
- Nested submenu leaves are displayed by xCore before the bridge sees them. Bites at mnuDataTree-Object.
- The Undo label after Alphabetical Order is the edit session's per-field label, not "Undo Alphabetical Order"; the Move path has had the same divergence since #1133.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1143)


jasonleenaylor (approved):
Approving. The adapter-versus-native equivalence tests are the right way to prove a swap like this. One divergence in the Move visibility rule, not for this PR:

`CanReorderItems` is `(real ReferenceSequence && !virtual) || node.Reorder` (`DetailComposer.cs:2369-2374`). `VectorReferenceView.cs:348-350` hides the moves for three more cases: `m_rootFieldName == "Complex Forms"`, `== "Compare"`, and the `LexReferenceCollectionView` / `LexReferenceTreeBranchesView` views. `m_rootFieldName` is the localized label (`ReferenceSlice.cs:55`), and `LexEntryParts.xml:824` is `label="Complex Forms" menu="mnuReorderVector" reorder="true"` — so WinForms hides Move Left and Move Right on that row and the authority's rule does not.

Not reachable today — that row composes as read-only Text, `CanReorderRow` is false, and both paths hide the moves, which is what `ReorderVectorLabelMenu_ReadOnlyComplexFormsRow_RendersTheSameTreeAsTheAdapter` pins. It bites only if `ComplexFormEntries` ever composes editable.

Not a straight port either: that condition matches on a localized label, so it already misses under a non-English UI. Noting it as something to either say at the rule or carry into a Jira issue when the next menu id moves across.

mark-sil (commented):
If I understand this correctly, the divergence is theoretical today, and from what I know there are no plans for ComplexFormEntries to compose editable.

If that is correct, then I think this falls in the category of 'busy work'. The Complex Forms equivalence test already asserts that the row composes read-only and that the moves are hidden. If that row ever composes editable, the test fails and points at exactly this.

Claude confirmed a note at the rule would add nothing the test does not already do, and logging a Jira issue for a theoretical problem doesn't seem like a good idea.

My preference would be to do nothing with issues like this. If you see a reason this one is different I'm glad to hear it, but otherwise I'd like to treat divergences that are unreachable today as needing no follow-up.

> @mark-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1143#-:-P2DL8YA2RQQ1ZlRsHJW:bpea40)* status: 0 of 15 files reviewed, all discussions resolved.

PR #1144, opened 2026-09-17
LT-22691: Reorder the root list when Move Field moves a top-level row

mark-sil:
**Start here:** `Src/xWorks/Avalonia/Hosting/RecordEditView.Avalonia.cs`, `ApplyMoveField` — the entire production change is four lines in that one method.

Move Up / Move Down on an Avalonia detail row at the **top level of its own layout** enabled the menu item and then wrote nothing. `ApplyMoveField` bailed out whenever `ViewNodeLocation.ParentStableId` was null, which is exactly what `ViewDefinitionOverrideEditor.LocateTarget` reports for a root-level row. Enablement reads only the sibling order, so the item stayed live the whole time: a Move Up that looked available and silently dropped every click. It now passes an empty parent id instead of bailing.

The empty `id` is not a new format. It is the key the override subsystem already uses for the root list: `ViewDefinitionOverrideDiffer` emits root reorders under `RootParentKey` (`""`), `ViewDefinitionOverrideApplier` rebuilds the root list under the same key, and its `ReportUnresolved` already exempts it, with the comment "root inserts and root-level reorder". All of that predates this branch, so a build without this change still reads a file written by a build with it.

**Where to look**

- `ApplyMoveField` — the `?? string.Empty`. Pinned by the new test, and by the negative control below.
- `MoveField_OnATopLevelRow_ReordersTheRootList` — picks its subject by shape, not by name. The reason is in "Decisions" below, and it is not stylistic.
- The reworded comment on `MutateOverrideAndRefresh` — a claim that had gone stale, not a behaviour change.

**Deliberately not here**

- **Move Field defect 1**: a move swaps with the adjacent *template* sibling whether or not that sibling is rendered, so a swap with a hidden neighbour still shows nothing. Untouched.
- **Enablement parity**: Avalonia enables Move Up where the WinForms detail view grays it. Open question, detailed below.

**Verification.** Build clean; the new test passes and fails without the fix; manually confirmed on a restored Sena 3. Details below. No full `xWorksTests` run.

---

<details>

This branch carried no working documents, so nothing was deleted to produce this record — the reasoning below was written for the PR directly.

The short version: `.viewoverride.json` has always keyed the root list on an empty string. Every part of the override subsystem knew that except the one method that writes a reorder from the Move Field menu, which treated "no parent" as "no move" and returned. The fix is to stop treating those as the same thing.

</details>

<details>

**Pass the empty key rather than teach the applier about null.** The alternative was to let `ViewOverrideOperation` carry a null `StableId` for root-level ops and have the applier map null to the root list. Rejected: `ViewOverrideOperation` deliberately rejects a null `StableId` in its constructor, the JSON serializer writes and reads `id` verbatim, and the differ already produces the empty-string form. A second spelling of "the root" would mean two readers, two writers, and a migration for files already on disk. Passing an empty id makes the menu writer agree with the differ that was already writing that value.

**The test selects its subject by shape, not by name.** An earlier version of this test named the Lexeme Form row and asserted it could move down. That row is the last root of its layout, so the `Assume` failed and NUnit **silently skipped the test inside an otherwise green run** — the worst possible outcome for a regression test. The current version searches the composed entry for the first row whose own layout places it at the top level (`LocateTarget` reports no parent) and that can move, then hard-`Assert`s that such a row exists. A shape search with a hard assert cannot skip. The consequence to accept is that a layout change can move the test onto a different row without anyone noticing; a `TestContext.WriteLine` prints the row, class, layout and direction on every run, so the run record always says which row was exercised.

**The comment on `MutateOverrideAndRefresh` was reworded, not just re-wrapped.** It claimed the hidden DataTree and the part-ref inventory were untouched. The inventory half is misleading: the writing-system copy path exists precisely because the hidden adapter writes the inventory itself. The comment now states what that method actually guarantees — it writes the override file, not the shipped XML and not the part-ref inventory. Both edited comment blocks were re-wrapped in the process, because the originals already exceeded the 98-column line limit and the 200-character block budget the hygiene gate enforces on added lines.

</details>

<details>
<summary><b>Surprising finding — the WinForms enablement rule is not index-based</b></summary>

Left open deliberately.

Avalonia enables Move Up on the Citation Form row; the WinForms detail view grays it. Avalonia asks only `Index > 0` / `Index < SiblingOrder.Count - 1` over **template** siblings. `Slice.CheckValidMove` requires three separate things, any one of which grays the item:

1. `MoveableFieldReferenceForSlice()` must be non-null — and `IsMoveableNode` returns false for a part node that represents a sequence *as a whole* (the "Variant Type" under `ref="EntryRefs"` case). Such a row is never moveable at that level, in either direction.
2. `GetSibling(dir)` must find a **rendered** neighbour: it walks `ContainingDataTree.Slices`, skipping deeper children, stopping at shallower depth, and skipping nested headers that share a fieldRef.
3. `Prev/NextSliceSiblingPart` must then find that rendered neighbour's part by walking template siblings from this one.

So Avalonia over-enables in three distinct ways, of which the hidden-neighbour case (defect 1) is only one.

The Avalonia move does work — Citation Form and Lexeme Form visibly swap. That is **not** being claimed as an improvement: someone may have intended Lexeme Form to always be first, which would make the WinForms refusal deliberate rather than a bug. Worth settling before anyone changes Avalonia enablement in either direction, and out of scope here.

</details>

<details>

- **Defect 1 — swaps with an unrendered template sibling.** Deterministic, not flaky: when the neighbour is `visibility="never"` or `ifdata` with no data, the move saves and refreshes correctly and nothing visible changes. Unblocking it means deciding whether Avalonia should match `GetSibling`'s rendered-neighbour walk, which is the same decision as the enablement question above.

</details>

<details>

**Automated**

```
.\test.ps1 -CommentHygiene -TokenHygiene -TestProject xWorksTests -TestFilter "MoveField_OnATopLevelRow_ReordersTheRootList"
```

Build succeeded, 0 warnings, 0 errors. Comment hygiene and token hygiene both clean. `Total tests: 1, Passed: 1`. The run printed its subject: `class=LexEntry layout=Normal template=LexEntry/Normal/#1 index=1 up=True`.

**Negative control.** The same filter with the production fix reverted: `Total tests: 1, Failed: 1`, failing on *"an enabled Move must write a project override, not silently do nothing"*. The test pins the defect rather than passing vacuously.

**Manual**, on a restored Sena 3 in the Avalonia detail view:

1. Open any lexical entry in Lexicon Edit.
2. Right-click the **Citation Form** label, then Move Field, then Move Up.
3. Citation Form and Lexeme Form swap places.
4. `<project>\ConfigurationSettings\LexEntry.Normal.viewoverride.json` is written, with an empty `"id"` on the `reorderChildren` op.

Note for anyone repeating this: do **not** use the WinForms detail view for step 2 — it grays Move Up on that row, which is the parity gap described above and not this defect.

**Not run.** No full `xWorksTests` pass. The fixture bootstraps a real window per test, and the change is confined to one method plus one new test.

</details>

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/FieldWorks/1144)


jasonleenaylor (approved):
LGTM

